PR #22, opened 2021-04-01
Createsession forreal

smialko:
CREATING FIXED SESSION 

**SUMMARY** 
Creating a fixed wifi session results in adding a new session to the dashboard. Measurements refresh every minute. If there are no measurements, there's a label displayed on the session view.  Session name, start time, end time, type, and device type are displayed - the data isn't mocked anymore. 
The app communicates with the web app to create an empty fixed session, configures the AB3, and gets measurements from the web. 
The created session is saved to the phone's database. 

**CODE**
- views are presented as full-screen covers (except ChooseSessionView)
 
**ChooseSessionTypeView**
- there's an environment object `sessionContext` used to temporarily save sessions information. 
- session type and UUID is saved to sessionContext, 
- after choosing the session type user is redirected to the proper view ( FIXED: Bluetooth unabled/turned off -> TurnOnBluetooth, FIXED&Bluetooth enabled&turned on -> PowerABView, MOBILE -> SelectDeviceView) 
- when user taps on fixed session  there should be a Bluetooth auth popup,

**SelectPeripheralView**
- selected Bluetooth device is saved to `sessionContext`

**AddNameAndTagsView**
- if a session is fixed there are pickers for indoor/outdoor, wifi/cellular
- if the user picks wifi, there should be a sheet displayed with a text field for wifi credentials - only for a password if the app was able to get wifi's SSID, for password and SSID otherwise, 
- if the user selects indoor, the app is saving a fake location to context (location can't be nil - it's needed to configure AB3), otherwise it's using LocationProvider to obtain the user's current location, 
- name, tags, indoor/outdoor, wifi credentials are saved to the `sessionContext`

**ConfirmCreatingSessionView & CreateSessionContext**
- when the user taps on "record session" the setupAB function is called, 
- it's calling `newOrExisting` function that creates a new session in the database, 
- after assigning values to session attributes context is saved, 
- FIXED WIFI : 
 * `CreateSessionApi` is a. call to API to create an empty session in the web app, 
 * if creating empty session is successful `configureFixedWifiSession` is triggered to configure AB3, 
 
**SessionHeaderView**
- session is fetched from the database (in Dashboard and it's passed to SessionHeaderView), 
- all the data displayed on session view is pulled from database, 

**MainTabBarView & DownloadMeasurementsService**
- on appear app starts downloading measurements with DownloadMeasurementsService, 
- DownloadMeasurementsService is making a call to API every 60s to get fixed measurements,
- if the call was successful, the app checks if this session is already in a database and either saves measurements to existing session or creates a new session, 
- then the context is saved.


AirCasting/ABConnector/SessionContext.swift
```diff
@@ -17,9 +17,10 @@ class CreateSessionContext: ObservableObject {
     var peripheral: CBPeripheral?
     var wifiSSID: String?
     var wifiPassword: String?
+    var isIndoor: Bool?
     var startingLocation: CLLocationCoordinate2D?
-    var deviceType: DeviceType?
-    
+    var deviceType: DeviceType? = DeviceType.AIRBEAM3
```
anna1901, line 22: does it have to be optional then?

smialko: True, thanks

smialko:
I'm merging this PR, but I'd be grateful for a CR ;)

AirCasting/CreateSessionViews/ChooseSessionTypeView.swift
```diff
@@ -119,18 +119,18 @@ struct ChooseSessionTypeView: View {
...
-                    })
+                EmptyView()
+                    .fullScreenCover(isPresented: $isPowerABLinkActive) {
+                        CreatingSessionFlowRootView {
+                            PowerABView()
+                        }
+                    }
+                EmptyView()
+                    .fullScreenCover(isPresented: $isTurnBluetoothOnLinkActive) {
+                        CreatingSessionFlowRootView {
```
anna1901, line 130: Could you explain what is happening here?

AirCasting/CreateSessionViews/ConfirmCreatingSessionView.swift
```diff
@@ -32,6 +33,7 @@ struct ConfirmCreatingSessionView: View {
                         
             Button(action: {
                 sessionContext.setupAB()
+                didStartRecordingSession = true
```
anna1901, line 36: Do we use this anywhere?

AirCasting/CreateSessionViews/WifiPopupView.swift
```diff
@@ -12,15 +12,14 @@ import SystemConfiguration.CaptiveNetwork
...
     @Binding var wifiPassword: String
     @Binding var wifiSSID: String
 
     var body: some View {
         VStack(alignment: .leading, spacing: 30){
-            if isWifiNameDisplayed {
+            if isSSIDTextfieldDisplayed || wifiSSID.isEmpty {
                 providePasswordTitle
                 createTextfield(placeholder: "Wi-Fi name", binding: $wifiSSID)
                 createTextfield(placeholder: "Password", binding: $wifiPassword)
```
anna1901, line 25: Should we hide user's password when he is entering it? We could use secure field https://developer.apple.com/documentation/swiftui/securefield
Actually same could be done at log in

AirCasting/Services/UpdateSessionParams.swift
```diff
@@ -0,0 +1,70 @@
...
+        session.name = output.title
+        session.tags  = output.tag_list
+        session.startTime  = dateFormatter.date(from: output.start_time)
+        session.endTime  = dateFormatter.date(from: output.end_time)
+        session.version = Int16(output.version)
+    
+        for (_, streamOutput) in output.streams {
+            let stream: MeasurementStream = self.context.newOrExisting(id: streamOutput.id)
+            
+            stream.sensorName = streamOutput.sensor_name
```
anna1901, line 33: Have you considered checking if the stream exists in the db and if it does then just saving the measurement and not update or of the other information? If that would work then it would help with the issue of resetting thresholds set by the user

AirCasting/CreateSessionViews/ConfirmCreatingSessionView.swift
smialko: We don't, nice catch, thanks. I'm removing it.

AirCasting/CreateSessionViews/WifiPopupView.swift
smialko: Both signup and login views have SecureFied.
I can change it here but I have some doubts about it. It's not that easy to uncover input in a secure field. That's why I chose to use the normal text field. We don't check if the password is correct and if it's incorrect the AB3 won't be configured correctly and the user won't get any measurements. We won't be able to inform users why the session is empty.

AirCasting/Services/UpdateSessionParams.swift
```diff
@@ -0,0 +1,70 @@
...
+class UpdateSessionParamsService {
+    
+    var context: NSManagedObjectContext {
+        PersistenceController.shared.container.viewContext
+    }
+    
+    func updateSessionsParams(session: Session, output: FixedSession.FixedMeasurementOutput) {
+        let dateFormatter = ISO8601DateFormatter.defaultLong
+    
+        session.id = Int64(output.id)
```
anna1901, line 20: Do we need to save this id from web to our db? I think session id is set automatically and we shouldn't change it. It doesn't have to be the same in our db as it is in web.

AirCasting/Extensions/AirCasting 2021-04-01 15-37-29/DistributionSummary.plist
```diff
@@ -0,0 +1,270 @@
...
+			<dict>
+				<key>id</key>
+				<string>LXAGXH6CF6</string>
+				<key>name</key>
+				<string>HabitatMap, Inc.</string>
+			</dict>
+		</dict>
+	</array>
+</dict>
+</plist>
```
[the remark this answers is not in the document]
smialko: Umm, it must have been generated by XCode. I was sending a build to Micheal and I guess that's when it was created.

AirCasting/Extensions/AirCasting 2021-04-01 15-37-29/ExportOptions.plist
```diff
@@ -0,0 +1,27 @@
...
+	<key>signingStyle</key>
+	<string>manual</string>
+	<key>stripSwiftSymbols</key>
+	<true/>
+	<key>teamID</key>
+	<string>LXAGXH6CF6</string>
+	<key>thinning</key>
+	<string>&lt;none&gt;</string>
+</dict>
+</plist>
```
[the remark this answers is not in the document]
smialko: Same here, it's some distribution info that was created by XCode when I was sending a build.

AirCasting/Models/SessionExtension.swift
```diff
@@ -0,0 +1,48 @@
...
+//
+//  Created by Lunar on 30/03/2021.
+//
+
+import Foundation
+
+extension Session {
+    
+    var pm1Stream: MeasurementStream? {
+        let matach = measurementStreams?.first(where: { (stream) -> Bool in
```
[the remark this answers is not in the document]
smialko: Ugh, I hate this keyboard. Thanks.

AirCasting/Services/DownloadMeasurmentsService.swift
```diff
@@ -0,0 +1,59 @@
...
+        let syncDate = Date().addingTimeInterval(-8000)
+        sink = FixedSession
+            .getFixedMeasurement(uuid: uuid,
+                                 lastSync: syncDate)
+            .sink { (completion) in
+                switch completion {
+                case .finished:
+                    print("sucess")
+                case .failure(let error):
+                    print("ERROR: \(error)")
```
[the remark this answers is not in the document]
smialko: True, I'm adding ticket to trello

AirCasting/Services/UpdateSessionParams.swift
```diff
@@ -0,0 +1,70 @@
...
+        session.uuid = output.uuid
+        session.type = SessionType.from(string: output.type)?.rawValue ?? -1
```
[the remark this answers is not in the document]
smialko: There's no such type. There has to be a value for session type and it's optional.  So if it's -1 there's something wrong in the app. I didn't want to make it mobile/fixed here, I think it might be more difficult to debug.

AirCasting/Services/UpdateSessionParams.swift
smialko: Hmm, I guess not.

AirCasting/CreateSessionViews/ChooseSessionTypeView.swift
smialko: Sure. It's complicated to explain as it was created because of some bugs I encountered.
The main bug was - when I pushed view as fullScreenCover I couldn't dismiss it. So I decided to add the cancel button. But, there's PowerAB view that can be the first fullScreenCover view (so it needs cancel button) or the second view, after TurnBluetoothOn view, and then it had Back button (it's in navigation view so it's added automatically). Adding the cancel button on this view resulted in having two buttons. There were other problems with that but these are the main two. 
So I created CreatingSessionFlowRootView that has a content of type View.  `content` is a closure that returns `Content` of type `View`. It has no initializers so when I initialize `CreatingSessionFlowRootView` I'm providing one of our views there. 
I added the cancel button in `CreatingSessionFlowRootView`. So now, instead of creating cancel buttons on TurnBluetoothOn and PowerAB and trying to find out when should I display cancel on PowerAB view, I'm creating it on CreatingSessionFlowRootView that displays the content (depending on the situation it's one of our views) and adds cancel button to the navigation bar.

anna1901: I get it. Thanks for the explanation :)

AirCasting/CreateSessionViews/WifiPopupView.swift
anna1901: My bad, I got confused and thought that the password was visible when I was signing in, but I see that it's not. When I was signing in I noticed that we didn't use keyboard for emails there, and it could be a better experience for user if we did. But that's an irrelevant issue here. :) This might be good idea to show the password to the user in this case.

AirCasting/SessionViews/SessionHeaderView.swift
```diff
@@ -78,9 +106,52 @@ struct SessionHeaderView: View {
...
+        let h: Double
+    }
+    func extractLatestMeasurements() -> LatestMeasurements? {
+        let pm1Value = session.pm1Stream?.latestValue ?? 0
+        let pm25Value = session.pm2Stream?.latestValue ?? 0
+        let pm10Value = session.pm10Stream?.latestValue ?? 0
+        let fValue = session.FStream?.latestValue ?? 0
+        let hValue = session.HStream?.latestValue ?? 0
+        
+        //TODO: change logic here (session status)
```
anna1901, line 124: Just in case I'm letting you know it's still here :)

AirCasting/CreateSessionViews/WifiPopupView.swift
smialko: I already changed the keyboard for email address, thanks :)

PR #23, opened 2021-04-08
Create microphone session

anna1901:
## Summary
This PR introduces feature for creating a mobile microphone session.
There is an environment object called `microphoneManager` which gets created when the app in launched. Then, when user presses "Start recording" button in `ConfirmCreatingSession` view the microphoneManager starts recording noise level every 60 seconds and it saves it to the database as a measurement.

## Features
* microphone session gets created and saved to the database
* decibels measurement stream is created and assigned to the session
* every 60 seconds microphoneManager gathers data about the background noise from microphone (in dBFS. Values range from -160 (silence) to 0 (a maximum volume a microphone can detect))
* the noise information is saved to database as measurement value, together with information about user's location and time of measurement

This PR doesn't include implementation of calculation dB values based on dBFS returned by function that gathers noise level information from microphone (`recorder.averagePower(forChannel: 0)`)
In those links there is an explanation why it is a complex thing to do:
https://stackoverflow.com/questions/2445756/how-can-i-calculate-audio-db-level
https://forum.juce.com/t/how-to-convert-decibels-dbfs-to-real-db-spl/36856/26


AirCasting/MicrophoneSession/MicrophoneManager.swift
```diff
@@ -0,0 +1,125 @@
...
+        
+        let measurement = Measurement(context: self.context)
+        
+        measurement.latitude = startingLocation.latitude
+        measurement.longitude = startingLocation.longitude
+        measurement.value = Double(level)
+        measurement.time = Date()
+        
+        session?.endTime = measurement.time
+        session?.status = 0
```
smialko, line 80: It would be a bit more readable if you use SessionStatus raw value here. 

`session?.status = Int16(SessionStatus.RECORDING.rawValue)`

AirCasting/MicrophoneSession/MicrophoneManager.swift
```diff
@@ -0,0 +1,125 @@
...
+            try audioSession.setCategory(AVAudioSession.Category.playAndRecord)
+            try audioSession.setActive(true)
+            try recorder = AVAudioRecorder(url:url, settings: recordSettings)
+            
+        } catch {
+            return
+        }
+        
+        recorder.isMeteringEnabled = true
+        recorder.record()
```
smialko, line 61: I think it might not work properly when it's the first time creating a mic session. 
This code will be executed even if the user is on permission popup, and didn't give us permission yet. Then, the session will be created, `recorder.record()` will be called, etc - all of this before knowing if we have access to the mic. 
I think it would be safer to move this code to the if, like this :
```
        if audioSession.recordPermission != .granted {
 ....... } else {
 self.session = session
        createDBStream()
        locationProvider.requestLocation()
        
        let url = URL(fileURLWithPath: "/dev/null", isDirectory: true)
        
        let recordSettings: [String: Any] = [
            AVFormatIDKey: NSNumber(value: kAudioFormatAppleLossless),
            AVSampleRateKey: 44100.0,
            AVNumberOfChannelsKey: 1,
            AVEncoderAudioQualityKey: AVAudioQuality.min.rawValue
        ]
        
        do {
            try audioSession.setCategory(AVAudioSession.Category.playAndRecord)
            try recorder = AVAudioRecorder(url:url, settings: recordSettings)
            
        } catch {
            return
        }
        
        recorder.record()
        
        levelTimer = Timer.scheduledTimer(timeInterval: 1, target: self, selector: #selector(getMeasurement), userInfo: nil, repeats: true)
}
```

Maybe, you could extra this part of code to a different function, too?

AirCasting/MicrophoneSession/MicrophoneManager.swift
```diff
@@ -0,0 +1,125 @@
...
+    }
+    
+    func stopRecording() {
+        levelTimer.invalidate()
+    }
+    
+    private
+    
+    func createDBStream() {
+        if let session = session {
```
smialko, line 100: I'm not sure if that's the case in our app, depends if the mobile mic session should refresh recording automatically after f.e. force quitting the app. 
If yes, you do not check if this stream already exists in the database, you create a new one every time.

smialko (commented):
I'm really impressed how you handled this feature. And your code is super clean & easy to read 👏

AirCasting/MicrophoneSession/MicrophoneManager.swift
anna1901: This doesn't work because code in the else statement doesn't execute at all if the permission wasn't granted prior to entering this view.

AirCasting/MicrophoneSession/MicrophoneManager.swift
```diff
@@ -0,0 +1,125 @@
...
+        print(measurement)
+        
+        do {
+            try context.save()
+        } catch {
+            print(error)
+        }
+    }
+    
+    func stopRecording() {
```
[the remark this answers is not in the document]
anna1901: Yes, but it is going to be used for stopping recording

AirCasting/MicrophoneSession/MicrophoneManager.swift
anna1901: We would need some kind of resumeRecording function for that and if that's the behavior we want we should implement it in a separate PR.

AirCasting/CreateSessionViews/ConfirmCreatingSessionView.swift
```diff
@@ -32,24 +43,33 @@ struct ConfirmCreatingSessionView: View {
...
+                    sessionContext.startMicrophoneSession(microphoneManager: microphoneManager)
+                } else {
+                    sessionContext.setupAB()
+                }
+                self.isActive = true
             }, label: {
                 Text("Start recording")
                     .bold()
             })
+            .background(
```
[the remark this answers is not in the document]
anna1901: Yes, I know, that's why I left a todo note and created trello ticket for that. It is related to all of the sessions, not only microphone, so I thought it could be done separately.

AirCasting/MicrophoneSession/MicrophoneManager.swift
```diff
@@ -0,0 +1,125 @@
...
+
+class MicrophoneManager: ObservableObject {
+    var context: NSManagedObjectContext {
+        PersistenceController.shared.container.viewContext
+    }
+    var session: Session?
+    var measurementStream: MeasurementStream?
+    
+    private var recorder: AVAudioRecorder!
+    private var levelTimer = Timer()
```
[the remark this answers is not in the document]
anna1901: Because it's a timer for noise level measurement. :D Is it confusing?

PR #24, opened 2021-04-09
Display microphone session

anna1901:
# Summary

Data from microphone session gets displayed in dashboard in session cell. It is possible now to stop recording session and it is impossible to start recording another microphone session when the other one is still recording.

## Features
* Displaying measurements in sessionHeader
* Add button for stopping microphone session recording
* Prevent creating more than one microphone session

### DIFF
--- a/AirCasting/MicrophoneSession/MicrophoneManager.swift
+++ b/AirCasting/MicrophoneSession/MicrophoneManager.swift
@@ -87,8 +87,6 @@ class MicrophoneManager: ObservableObject {
         
         measurementStream?.addToMeasurements(measurement)
         
-        print(measurement)
-        
         do {
             try context.save()
         } catch {
--- a/AirCasting/MicrophoneSession/MicrophoneManager.swift
+++ b/AirCasting/MicrophoneSession/MicrophoneManager.swift
@@ -17,19 +17,24 @@ class MicrophoneManager: ObservableObject {
     var session: Session?
     var measurementStream: MeasurementStream?
     
+    //This variable is used to block recording more than one microphone session at a time
+    var isRecording = false
+    
     private var recorder: AVAudioRecorder!
     private var levelTimer = Timer()
     
     private var locationProvider = LocationProvider()
     
     func startRecording(session: Session) {
+        if isRecording { return }
+        
         let audioSession = AVAudioSession.sharedInstance()
         
         //Check microphone permisson
         if audioSession.recordPermission != .granted {
             audioSession.requestRecordPermission { (isGranted) in
                 if !isGranted {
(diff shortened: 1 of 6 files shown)


AirCasting/MicrophoneSession/MicrophoneManager.swift
```diff
@@ -17,19 +17,24 @@ class MicrophoneManager: ObservableObject {
     var session: Session?
     var measurementStream: MeasurementStream?
     
+    //This variable is used to block recording more than one microphone session at a time
+    var isRecording = false
```
mateuszmackowiak, line 21: Seams like this value should not be editable from outside of the class. Please add `private(set)` or `private`

AirCasting/Models/SessionContext.swift
```diff
@@ -20,7 +20,7 @@ class CreateSessionContext: ObservableObject {
     var wifiPassword: String?
     var isIndoor: Bool?
     var startingLocation: CLLocationCoordinate2D?
-    var deviceType: DeviceType?
+    var deviceType: DeviceType = DeviceType.AIRBEAM3 // It is set here temporarily to fix bug with fixed sessions
```
mateuszmackowiak, line 23: If temporary maybe add a `TODO` note?

AirCasting/SessionViews/SessionHeaderView.swift
```diff
@@ -13,12 +13,23 @@ struct SessionHeaderView: View {
...
     @ObservedObject var session: Session
-        
+    @EnvironmentObject private var microphoneManager: MicrophoneManager
+    
     var body: some View {
         VStack(alignment: .leading, spacing: 13){
             dateAndTime
             nameLabelAndExpandButton
-            measurements
+            if (session.deviceType == DeviceType.MIC.rawValue) {
```
mateuszmackowiak, line 22: should we reference rawValue in comparison? Maybe it would be better to use an enum case in `session`

AirCasting/SessionViews/SessionHeaderView.swift
```diff
@@ -132,6 +159,11 @@ struct SessionHeaderView: View {
             return nil
         }
     }
+    
+    func lastMicMeasurement() -> Int {
+        return Int(session.dbStream?.latestValue ?? 0)
```
mateuszmackowiak, line 164: Can the value be negative? Wouldn't a UInt be a better return type?

anna1901: For now it's only negative

AirCasting/MicrophoneSession/MicrophoneManager.swift
anna1901: Good idea, thanks

PR #25, opened 2021-04-13
Logger

mateuszmackowiak:
basic implementation of Logger in the app. 
usage:
`Log.info("some log")`

Log levels:
error - really used begins with ❗️. should only be used in a critical places where almost an assert or fatal error should be used.
warning - for all errors in the app that we assume may happen but It is ok. Like invalid response from server. In that case we have a fallback but it would be helpful to know. Addionally warn logs begin with a ⚠️ to be noticed in the output console.
info - basic logs
debug - only for logs in debug mode
verbose - additional log level for more verbose log types. Maybe not needed


anna1901 (approved):
Looks great 👍

PR #26, opened 2021-04-13
Measurements api uni tests and a little refactor of api model

mateuszmackowiak:


AirCasting/Services/DownloadMeasurmentsService.swift
```diff
@@ -38,12 +38,13 @@ class DownloadMeasurementsService: ObservableObject {
                 }
             } receiveValue: { [weak self] (fixedMeasurementOutput) in
                 guard let self = self else { return }
-                
-                // Fetch session by id from Core Data
-                let session: Session = self.context.newOrExisting(uuid: fixedMeasurementOutput.uuid)
-                UpdateSessionParamsService().updateSessionsParams(session: session, output: fixedMeasurementOutput)
-                
-                try! self.context.save()
+                DispatchQueue.main.async {
```
anna1901, line 41: Tu nie będzie tego samego problemu z kontekstem?

anna1901: I'm not sure if it's a similar case to that previous issue but I just want to make sure

AirCasting/APICommunicator/FixedSession.swift
```diff
@@ -7,52 +7,78 @@
...
-        var tag_list: String
-        var start_time: String
-        var end_time: String
-        var deleted: Bool?
-        var version: Int
-        var streams: [String : StreamOutput]
+    struct FixedMeasurementOutput: Decodable, Hashable {
+        typealias ID = Int64
+        let id: ID
+        let type: SessionType
```
anna1901, line 16: Are we going to have to ensure type is of the same data type as is expected in the database (int or string)?

anna1901 (approved):
All looks great! 👍 There are just a couple of comments left but they are not blockers.

AirCasting/APICommunicator/FixedSession.swift
mateuszmackowiak: SessionType has a method `init(from decoder: Decoder) throws {` that is responsible for decoding from json. So the value will decode from a string value. then an additional computed value `var rawValue: Int64` is used to map to the database (hopefully temporary solution - the final one should store string value).

AirCasting/Services/DownloadMeasurmentsService.swift
mateuszmackowiak: This thread change to main was done before inside the `getFixedMeasurement` so this is the code that moves back to the main thread so the `UpdateSessionParamsService().updateSessionsParams` that currently is using the core data `viewContext` will work. Final version of the code should use a child context of the `viewContext` into which the database update should be made.

PR #29, opened 2021-04-16
Fix microphone session flow

anna1901:
### Summary
* Change flow for checking users permission for microphone. User gets ask to give permission to access microphone when they choose microphone in the `ChooseSessionTypeView`. If they give permission then we go further with the views for creating microphone session. When they don't give permission then they get sent to setting app every time they try to choose microphone
* Add functionality for resuming recording after audio session gets interrupted eg. by incoming phonecall or by another app

### DIFF
--- a/AirCasting/CreateSessionViews/SelectDeviceView.swift
+++ b/AirCasting/CreateSessionViews/SelectDeviceView.swift
@@ -93,7 +93,19 @@ struct SelectDeviceView: View {
                     isTurnOnBluetoothLinkActive = true
                 }
             } else if selected == 2 {
-                isMicLinkActive = true
+                if microphoneManager.recordPermissionGranted() {
+                    isMicLinkActive = true
+                } else {
+                    microphoneManager.requestRecordPermission { (isGranted) in
+                        DispatchQueue.main.async {
+                            if isGranted {
+                                isMicLinkActive = true
+                            } else {
+                                SettingsManager.goToAuthSettings()
+                            }
+                        }
+                    }
+                }
             }
         }, label: {
             Text("Choose")
@@ -113,7 +125,6 @@ struct SelectDeviceView: View {
                     EmptyView()
                 })
             NavigationLink(
-                // TO DO: change destination
                 destination: AddNameAndTagsView(),
                 isActive: $isMicLinkActive,
                 label: {
--- a/AirCasting/MicrophoneSession/MicrophoneManager.swift
+++ b/AirCasting/MicrophoneSession/MicrophoneManager.swift
@@ -1,8 +1,5 @@
 //
-//  microphoneManager.swift
-//  AirCasting
-//
-//  Created by Anna Olak on 11/03/2021.
+//  Created by Lunar on 11/03/2021.
 //
 
 import Foundation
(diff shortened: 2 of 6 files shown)


AirCasting/MicrophoneSession/MicrophoneManager.swift
```diff
@@ -21,86 +18,110 @@ class MicrophoneManager: ObservableObject {
...
-        if audioSession.recordPermission != .granted {
-            audioSession.requestRecordPermission { (isGranted) in
-                if !isGranted {
-                    //TODO: pop-up that informs user that we need access to mic with ability do go back to homescreen
-                    fatalError("You must allow audio recording for this demo to work")
-                }
+        if audioSession.recordPermission == .granted {
+            addInterruptionsObserver()
+            
+            self.session = session
```
mateuszmackowiak, line 32: I know that at this point it might be not easy to do but this needs refactoring. Session is a NSManagedObject. Keeping references to this objects are a bad idea. They are strictly related to the NSManageObjectContext. If ever this session would be removed or edited in any other way in the background the reference may fail/ crash. Also may not be a good idea to base recording logic on a database entity. A session could be created at the begging of a recording if really needed but until recording is finished or has at least reasonable amount of data to store this should be kept in some kind of in memory only object. For future reference lets call this kind of objects PODO's (Plane old data object).

AirCasting/MicrophoneSession/MicrophoneManager.swift
```diff
@@ -21,86 +18,110 @@ class MicrophoneManager: ObservableObject {
     private(set) var isRecording = false
     
     private var recorder: AVAudioRecorder!
+    private(set) var audioSession = AVAudioSession.sharedInstance()
```
mateuszmackowiak, line 21: It makes no sense to initialize this object whit this reference until it is needed. Pleas use `lazy var`

AirCasting/MicrophoneSession/MicrophoneManager.swift
```diff
@@ -21,86 +18,110 @@ class MicrophoneManager: ObservableObject {
...
         }
-        
-        recorder.isMeteringEnabled = true
-        recorder.record()
-        isRecording = true
-        
-        levelTimer = Timer.scheduledTimer(timeInterval: 1, target: self, selector: #selector(getMeasurement), userInfo: nil, repeats: true)
     }
     
     @objc func getMeasurement() {
```
mateuszmackowiak, line 62: please make it private. `@objc private func`

AirCasting/MicrophoneSession/MicrophoneManager.swift
```diff
@@ -21,86 +18,110 @@ class MicrophoneManager: ObservableObject {
...
+    func stopRecording() {
+        levelTimer.invalidate()
+        isRecording = false
+        recorder.stop()
+        recorder = nil
+        session?.status = Int16(SessionStatus.FINISHED.rawValue)
+        do {
+            try context.save()
+        } catch {
+            print("Error for saving session when stopping mic recording: \(error)")
```
mateuszmackowiak, line 80: It would be better to propagate this error further. Its not a good idea to have silent failure. Print may be a solution for some cases where the failure can be ignored. in this case if the context will not be saved then the whole process would make no sense,  the whole manangeobjectcontex also can be polluted with object that also in the future can fail because ignored save here will cause all the next save to fail. It should be clear form the top that a process can fail. Please mark the method `throws`.  If `stopRecording` is called directly in the ui then calll `try! stopRecording()` with a force unwrap try. [Fail-fast](https://dzone.com/articles/fail-fast-principle-in-software-development) also applies here but the call methods like this that could fail should propagate the error because this buiness logic may be used in other places where in theory it would be ok to let it silently fail.

AirCasting/MicrophoneSession/MicrophoneManager.swift
```diff
@@ -21,86 +18,110 @@ class MicrophoneManager: ObservableObject {
...
+        }
+    }
+    
+    @objc func handleInterruption(notification: Notification) {
```
mateuszmackowiak, line 84: please make it private. except the Notification Center here there should be no direct calls to this method form outside of the class.

AirCasting/MicrophoneSession/MicrophoneManager.swift
```diff
@@ -21,86 +18,110 @@ class MicrophoneManager: ObservableObject {
...
+        guard let userInfo = notification.userInfo,
+            let typeValue = userInfo[AVAudioSessionInterruptionTypeKey] as? UInt,
+            let type = AVAudioSession.InterruptionType(rawValue: typeValue) else {
+                return
+        }
+
+        switch type {
+        case .began:
+            levelTimer.invalidate()
+            session?.status = Int16(SessionStatus.DISCONNETCED.rawValue)
```
mateuszmackowiak, line 94: Seams like ` Int16(SessionStatus.DISCONNETCED.rawValue)` is something that repeats itself. 
if we would hide the raw value type that is kept in the database private in the session object and expose our enum setter and getter it would definitely be safer, more readable etc. 

Please change `@NSManaged public var status: Int16` to : 
```
public var status: SessionStatus? {
        get { (value(forKey: "status") as? Int).flatMap(SessionStatus.init(rawValue:)) }
        set { setValue(newValue?.rawValue, forKey: "status") }
    }
```
But also keep in mind the other comment about not storing unfinished session in the database but storing it after successful recoding session.

AirCasting/MicrophoneSession/MicrophoneManager.swift
```diff
@@ -21,86 +18,110 @@ class MicrophoneManager: ObservableObject {
...
-        session?.status = 0
+        session?.status = Int16(SessionStatus.RECORDING.rawValue)
         
         measurementStream?.addToMeasurements(measurement)
         
         do {
             try context.save()
         } catch {
-            print(error)
+            print("Error saving microphone measurement: \(error)")
```
mateuszmackowiak, line 121: Pleas propagate the error.

AirCasting/MicrophoneSession/MicrophoneManager.swift
```diff
@@ -21,86 +18,110 @@ class MicrophoneManager: ObservableObject {
...
         
         measurement.latitude = startingLocation.latitude
         measurement.longitude = startingLocation.longitude
-        measurement.value = Double(level)
+        measurement.value = Double(value)
         measurement.time = Date()
         
         session?.endTime = measurement.time
-        session?.status = 0
+        session?.status = Int16(SessionStatus.RECORDING.rawValue)
```
mateuszmackowiak, line 114: Use `session?.status = .RECORDING`. Also capitalized enum cases are not keeping the `Swift` standards.

AirCasting/SessionViews/SessionHeaderView.swift
```diff
@@ -33,7 +34,9 @@ struct SessionHeaderView: View {
         }
         .font(Font.moderate(size: 13, weight: .regular))
         .foregroundColor(.aircastingGray)
-        
+        .onAppear() {
+            print("\(session.name) status is: \(session.status)")
```
mateuszmackowiak, line 38: Please use `Log.info` and make the log more verbose. Give more context to it. You know what is this log but You'll probably forget what was it doing very fast. Be kind to other developers and to Your future self and let us know what is happening. if the session.name would be `foo` the log would be `foo status is recoding`. maybe `microfon session "foo" status is recording` or similar depending on what exactly is happening.

AirCasting/MicrophoneSession/MicrophoneManager.swift
```diff
@@ -122,9 +143,15 @@ class MicrophoneManager: ObservableObject {
             session.addToMeasurementStreams(measurementStream!)
         }
     }
-
+    
     func obtainCurrentLocation() -> CLLocationCoordinate2D {
         return locationProvider.currentLocation?.coordinate ?? CLLocationCoordinate2D(latitude: 200.0, longitude: 200.0)
     }
     
+    func addInterruptionsObserver() {
```
mateuszmackowiak, line 151: there is no removal of the notification. if the microton manage would be only one or would exist only for the time being of the record it maybe enough. But if there would ever be a different object that is using `AVAudioSession` then  its interruption would influence this object. also maybe pass in the `audioSession` as the `object:` argument and not nil then the Notification Center will not call this observer when the object is not the seam as the one calling it.

AirCasting/MicrophoneSession/MicrophoneManager.swift
```diff
@@ -21,86 +18,110 @@ class MicrophoneManager: ObservableObject {
...
+                return
+        }
+
+        switch type {
+        case .began:
+            levelTimer.invalidate()
+            session?.status = Int16(SessionStatus.DISCONNETCED.rawValue)
+        case .ended:
+            levelTimer = Timer.scheduledTimer(timeInterval: 1, target: self, selector: #selector(getMeasurement), userInfo: nil, repeats: true)
+        default: ()
```
mateuszmackowiak, line 97: Please use all the cases rather then using default. Let the compilation help us and if new case would be added a failure here would require to rethink this situation, how does the new enum influence the flow. Without it may be forgotten and ignored and that could lead to potential bugs that would be very hard to find.

AirCasting/CreateSessionViews/SelectDeviceView.swift
```diff
@@ -93,7 +93,19 @@ struct SelectDeviceView: View {
                     isTurnOnBluetoothLinkActive = true
                 }
             } else if selected == 2 {
-                isMicLinkActive = true
+                if microphoneManager.audioSession.recordPermission == .granted {
+                    isMicLinkActive = true
+                } else {
+                    microphoneManager.audioSession.requestRecordPermission { (isGranted) in
```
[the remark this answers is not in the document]
anna1901: I'm not sure how I can move `requestRecordPermission` function out of this view if I need to put the view logic inside the closure in this function. Otherwise I'm not sure how the app can wait for the user's choice before moving to another screen/going to settings.

AirCasting/MicrophoneSession/MicrophoneManager.swift
anna1901: You already corrected it in your last PR so I can leave it as it is?

AirCasting/MicrophoneSession/MicrophoneManager.swift
anna1901: Where should the removal be implemented?

AirCasting/MicrophoneSession/MicrophoneManager.swift
anna1901: `session?.status = .RECORDING` doesn't work and I get an error "Type 'Int16' has no member 'RECORDING'"

AirCasting/CreateSessionViews/SelectDeviceView.swift
```diff
@@ -93,7 +93,19 @@ struct SelectDeviceView: View {
                     isTurnOnBluetoothLinkActive = true
                 }
             } else if selected == 2 {
-                isMicLinkActive = true
+                if microphoneManager.recordPermissionGranted() {
```
anna1901, line 96: I created this function instead of checking here if audiosession.recordPermission == .granted. I thought maybe it would be better if microphoneManager was the one to know what the recordPermission should be equal to if permission is granted. But let me know if you prefer your solution with var recordPermission

AirCasting/MicrophoneSession/MicrophoneManager.swift
anna1901: done

AirCasting/MicrophoneSession/MicrophoneManager.swift
```diff
@@ -14,117 +11,156 @@ class MicrophoneManager: ObservableObject {
...
+        if audioSession.recordPermission == .granted {
+            addInterruptionsObserver()
+
+            self.session = session
+            createDBStream(for: session)
+            locationProvider.requestLocation()
+
+            let url = URL(fileURLWithPath: "/dev/null", isDirectory: true)
+
+            let recordSettings: [String: Any] = [
```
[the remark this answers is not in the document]
anna1901: Do you mean that this should be declared outside of function as the class variable?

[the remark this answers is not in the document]
mateuszmackowiak: not a variable because it does not change. A static let - constant 😄

mateuszmackowiak (approved):

PR #34, opened 2021-04-26
Session type changed to string and enforced uuid

mateuszmackowiak:


AirCasting/CoreData/Session.swift
```diff
@@ -60,46 +60,25 @@ public class Session: NSManagedObject, Identifiable {
...
-    public var uuid: SessionUUID? {
-        get { (value(forKey: "uuid") as? String).flatMap(SessionUUID.init(rawValue:)) }
-        set { setValue(newValue?.rawValue, forKey: "uuid") }
+    public var uuid: SessionUUID! {
+        get { SessionUUID(rawValue: value(forKey: "uuid") as! String) }
+        set { setValue(newValue.rawValue, forKey: "uuid") }
     }
 
-    public var deviceType: DeviceType? {
+    public var deviceType: DeviceType! {
```
smialko, line 68: Why it's forced unwrapped here? I know how it works, but I don't know why you did it this way - won't it crash when we create a session not knowing what device type is? 🤔

smialko (approved):
I have one question about force unwrapping DeviceType, but it all looks good to me 👌

AirCasting/CoreData/Session.swift
mateuszmackowiak: You are right. Will revoke

PR #35, opened 2021-04-26
Return to dashboard after creating session

anna1901:
Now after confirming creating session user goes back to dashboard view.


smialko (approved):
Nice!

PR #41, opened 2021-05-06
Thresholds slider

smialko:
**Note:** I use only one threshold - `ThresholdsSliderView(threshold: thresholds[0])`. 
In the future, we'll be using thresholds depending on the selected measurement stream. We do not have the functionality of selecting the stream yet, so I decided to simplify it for the sake of this chore.  I'm adding a ticket to Trello with the reminder about the thresholds. 

**Summary**
- I fixed the edit button view on session view, 
- I created a mocked threshold for previews, 
- When the user changes the threshold on slider or HeatMapView, the values are updated everywhere, 
- I fixed creating multi slider - it wasn't working when thresholds were negative.


AirCasting/Extensions/SensorThreshold.swift
```diff
@@ -0,0 +1,48 @@
...
+            thresholdMedium = Int32(newThresholds[2])
+            thresholdHigh = Int32(newThresholds[3])
+            thresholdVeryHigh = Int32(newThresholds[4])
+        }
+    }
+}
+
+extension SensorThreshold {
+    // mocked SensorThreshold is used for previews (& temporarily some views)
+    static var mock: SensorThreshold {
```
mateuszmackowiak, line 31: This should be avoided. If needed for preview should be wrapped in `#if DEBUG` and if needed for development should not end up in merged develop code. just by referencing without context save possible to create garbage value that could cause random failures in random places that could lead to dead end investigation. Also if needed that means a protocol should lld be created and mocks unreleased to core data could be made. All in all please avoid at all costs. Even if needed for at start for feature branch development please start with `@available(*, deprecated, renamed: "DEVELOPMENT ONLY")`  annotation

AirCasting/Graph/Graph.swift
```diff
@@ -73,13 +73,13 @@ class UI_PollutionGraph: UIView {
         lineChartView.leftYAxisRenderer = renderer
     }
     
-    func updateWith(values: [Float]) {
+    func updateWith(thresholdValues: [Float]) {
         guard let renderer = renderer else { return }
```
mateuszmackowiak, line 77: method does not describe that it is fail-able. Please consider [Fail-fast](https://www.google.com/search?client=safari&rls=en&q=fail+fast&ie=UTF-8&oe=UTF-8) rule. Make it failable and by walking upwards handle the optionally.

AirCasting/Graph/Graph.swift
```diff
@@ -73,13 +73,13 @@ class UI_PollutionGraph: UIView {
...
-        renderer.thresholds = values
+        renderer.thresholds = thresholdValues
         
         //min & max yaxis values
-        lineChartView.leftAxis.axisMinimum = Double(values.first ?? 0)
-        lineChartView.leftAxis.axisMaximum = Double(values.last ?? 200)
+        lineChartView.leftAxis.axisMinimum = Double(thresholdValues.first ?? 0)
+        lineChartView.leftAxis.axisMaximum = Double(thresholdValues.last ?? 200)
         print("\(lineChartView.leftAxis.axisMinimum) - \(lineChartView.leftAxis.axisMaximum)")
```
mateuszmackowiak, line 83: Please consider using log, and also make the logs clear. This print may make sense for You now but is not understandable out of context where You are not debuging and not analilzing this part of code.

AirCasting/Graph/Graph.swift
```diff
@@ -73,13 +73,13 @@ class UI_PollutionGraph: UIView {
...
         
         // Needed to redraw the chart
```
mateuszmackowiak, line 85: Is the code under not understandable? are the comments needed? the code seams pretty self explanatory. Seams more like for a tutorial land not production code

AirCasting/Graph/Graph.swift
```diff
@@ -125,19 +125,19 @@ class MultiColorGridRenderer: YAxisRenderer {
...
         UI_PollutionGraph()
     }
     
     func updateUIView(_ uiView: UI_PollutionGraph, context: Context) {
-        uiView.updateWith(values: thresholds)
+        uiView.updateWith(thresholdValues: thresholds.rawThresholdsBinding.wrappedValue)
     }
 }
 
 struct PollutionGraph_Previews: PreviewProvider {
```
mateuszmackowiak, line 139: Please wrap in `#if DEBUG`

AirCasting/Graph/GraphView.swift
```diff
@@ -40,6 +44,17 @@ struct GraphView: View {
 
 struct GraphView_Previews: PreviewProvider {
```
mateuszmackowiak, line 45: Please wrap in `#if DEBUG`

AirCasting/Graph/GraphView.swift
```diff
@@ -40,6 +44,17 @@ struct GraphView: View {
...
     static var previews: some View {
-        GraphView(thresholds: .constant([0,1,2,3]))
+        GraphView(thresholds: [.mock])
     }
 }
+
+struct ObservedObjectWrapper<T: ObservableObject, Content: View>: View {
```
mateuszmackowiak, line 51: Is this related to Graf? Seams like `Lazy` is the word that would describe this better

AirCasting/Map/AirMapView.swift
```diff
@@ -38,7 +39,7 @@ struct AirMapView: View {
 
 struct Map_Previews: PreviewProvider {
```
mateuszmackowiak, line 40: Please wrap in `#if DEBUG`

AirCasting/Map/GoogleMapView.swift
```diff
@@ -15,7 +15,7 @@ struct GoogleMapView: UIViewRepresentable {
     typealias UIViewType = GMSMapView
 
     let pathPoints: [PathPoint]
-    let thresholds: [Float]
+    var thresholds: [Float] = []
```
mateuszmackowiak, line 18: private(set) / private ?

AirCasting/Map/GoogleMapView.swift
```diff
@@ -87,15 +87,22 @@ struct GoogleMapView: UIViewRepresentable {
 
     func colorPolyline(point: PathPoint) -> UIColor {
         let measurement = point.measurement
+        guard thresholds.count >= 5 else { return UIColor.white }
+        
+        let veryLow = Float(thresholds[0])
+        let low = Float(thresholds[1])
+        let medium = Float(thresholds[2])
+        let high = Float(thresholds[3])
+        let veryHigh = Float(thresholds[4])
```
mateuszmackowiak, line 96: Seams like array is not a good model for the `thresholds` if it requires 4 elements always. Maybe can be changed to a light understandable struct?

AirCasting/Map/GoogleMapView.swift
```diff
@@ -124,7 +131,7 @@ struct GoogleMapView_Previews: PreviewProvider {
                                    PathPoint(location: CLLocationCoordinate2D(latitude: 40.93,
                                                                               longitude: -73.83),
                                              measurement: 30)],
-                      thresholds: [0, 25, 50, 75, 100])
+                      thresholds: SensorThreshold.mock.rawThresholdsBinding.wrappedValue)
```
mateuszmackowiak, line 134: please wrap the preview in `#if DEBUG`

AirCasting/Models/Session+Mock.swift
```diff
@@ -18,3 +18,6 @@ extension Session {
     }
     
 }
+
+
+
```
mateuszmackowiak, line 23: Please avoid committing only white signs only.

AirCasting/SessionViews/MultiSliderView.swift
```diff
@@ -26,26 +35,24 @@ struct MultiSliderView: View {
...
         GeometryReader { geometry in
-            let frameWidth = geometry.frame(in: .local).size.width
-            
             ZStack {
                 colors.last
                 ForEach(thresholdButtonValues.indices.reversed(), id: \.self) { index in
                     colors[index]
-                        .frame(width: CGFloat(thresholdButtonValues[index]) * frameWidth / CGFloat(thresholdVeryHigh))
-                        .position(x: CGFloat(thresholdButtonValues[index]) * frameWidth / CGFloat(thresholdVeryHigh) / 2,
+                    // TODO: handle situation when thresholdVeryHigh = 0
```
mateuszmackowiak, line 47: Please use `#warning`

AirCasting/SessionViews/MultiSliderView.swift
```diff
@@ -56,6 +63,11 @@ struct MultiSliderView: View {
         .frame(height: 5)
     }
     
+    func xFor(value: Float, geometry: GeometryProxy) -> CGFloat {
```
mateuszmackowiak, line 66: not sure if a good name. what is value? what exactly is `x`

AirCasting/SessionViews/MultiSliderView.swift
```diff
@@ -96,6 +108,8 @@ struct MultiSliderView: View {
 
 struct MultiSlider_Previews: PreviewProvider {
```
mateuszmackowiak, line 109: Please wrap in `#if DEBUG`

AirCasting/SessionViews/SessionCellView.swift
```diff
@@ -46,7 +49,7 @@ struct SessionCellView: View {
         let allLocationPoints = provider.allLocations
         let points = allLocationPoints.map { (location) in
             PathPoint(location: location.coordinate,
-                      measurement: Float(arc4random() % 200))
+                      measurement: Float(-10))
```
mateuszmackowiak, line 52: what does `-10` mean? Seams a little magical. please make a variable and "define" this magic. 
tenor

AirCasting/SessionViews/SessionHeaderView.swift
```diff
@@ -173,6 +172,7 @@ struct SessionHeader_Previews: PreviewProvider {
     static var previews: some View {
```
mateuszmackowiak, line 172: Please wrap in `#if DEBUG`

AirCasting/MicrophoneSession/MicrophoneManager.swift
```diff
@@ -158,7 +158,7 @@ private extension MicrophoneManager {
             thresholds.thresholdLow = -40
             thresholds.thresholdMedium = -30
             thresholds.thresholdHigh = -20
-            thresholds.thresholdVeryHigh = 0
+            thresholds.thresholdVeryHigh = 10
```
mateuszmackowiak, line 161: Maybe define a struct with default values so the code would be more understandable.

mateuszmackowiak (changes requested):

AirCasting/Graph/GraphView.swift
smialko: Well, I'm a little confused. I agree with you, that wrapping preview in `#if DEBUG` is a good idea. But we have lots of previews not wrapped in DEBUG, shouldn't we be consistent? 
I can wrap this one, but in my opinion either we should add a ticket to Trello to wrap them all, or I need an explanation why should I wrap this one and others can be unwrapped.

AirCasting/SessionViews/SessionCellView.swift
smialko: It's a magical way to avoid crashing the app because random measurements that were used for map view were bigger than thresholds. It was necessary here because I needed to check if the slider  & HeatMap work properly on both the graph and map view. 
I know, it's not a real fix - why this shortcut? Because it'll be fixed when we start providing a map with real data, and in this case, there's not much difference between generating random integers and setting a fixed value. 
But it's my bad, I should have added a comment there.

AirCasting/Extensions/SensorThreshold.swift
smialko: Thanks for the tip, I'll be doing that in the future. 
I added `#if DEBUG ` here.

AirCasting/Graph/Graph.swift
smialko: Those prints are leftovers from when I was creating the graph, I didn't add them now. 
I guess it's not necessary anymore so I'm removing it.

AirCasting/Graph/Graph.swift
smialko: Well, tbh I was adding those comments for myself. Creating the graph was a tough task for me, so I was describing lots of code to make the task easier. 
I'm removing this comment.

AirCasting/Graph/GraphView.swift
smialko: I added  `#if DEBUG` here, but I think that we should refactor the code so that all previews will be wrapped i  `#if DEBUG` :)

AirCasting/Graph/GraphView.swift
smialko: I think I don't understand your comment here.
What I use here is a convenient view wrapper that lets me observe changes to a single object from thresholds array.
It's impossible to mark array as an `@ObservedObject`, and I have to observe changes there so I needed a view for each of `thresholds` elements.
I should have added a comment there. I'm adding one now. :)

smialko: After our call I removed it to see why I implemented it in the first place...and everything is working just fine. It seems like there was a problem somewhere else.

AirCasting/Models/Session+Mock.swift
smialko: Sorry 😬

AirCasting/SessionViews/MultiSliderView.swift
smialko: It's super weird.
When I was adding a `#warning` there, I was getting an error that there's too much operation and the compiler cannot compile in a reasonable time. I wanted to paste a screenshot to prove it but it doesn't appear anymore. 
So I changed comment to warning.

AirCasting/SessionViews/MultiSliderView.swift
smialko: I chaged it to `    func calculateXAxisSize(thresholdValue: Float, geometry: GeometryProxy) -> CGFloat {...}`
I hope it's good enough :)

AirCasting/Graph/Graph.swift
smialko: I added it and learnt new thing, thanks ;)

mateuszmackowiak (approved):

PR #44, opened 2021-05-17
Picker tabs

smialko:
**SUMMARY**
- I created a custom picker view for selecting session type, 
- I fetch sessions only for selected section, 
- I created CoreDataHook that triggers fetch request when something changes in database and refresh the session list on view.


AirCasting.xcodeproj/project.pbxproj
```diff
@@ -1162,7 +1171,7 @@
 				);
 				PRODUCT_BUNDLE_IDENTIFIER = org.habitatmap.AirCasting;
 				PRODUCT_NAME = "$(TARGET_NAME)";
-				PROVISIONING_PROFILE_SPECIFIER = "AirCasting Ad hoc distribution";
+				PROVISIONING_PROFILE_SPECIFIER = "Michael Heimbinder";
```
mateuszmackowiak, line 1174: Are this changes intended? this will break the CI

AirCasting.xcworkspace/contents.xcworkspacedata
```diff
@@ -4,6 +4,19 @@
    <FileRef
       location = "group:AirCasting.xcodeproj">
    </FileRef>
+   <Group
+      location = "group:AirCasting/../Dashboard"
```
mateuszmackowiak, line 8: Is there a reason for Dashboard folder to be outside of the project?

AirCasting/CoreData/CoreDataHook.swift
```diff
@@ -0,0 +1,40 @@
+// Created by Lunar on 17/05/2021.
+//
+
+import Foundation
+import CoreData
+
+class CoreDataHook: NSObject, ObservableObject {
```
mateuszmackowiak, line 7: please make it final

AirCasting/CoreData/CoreDataHook.swift
```diff
@@ -0,0 +1,40 @@
...
+    
+    @Published var sessions: [Session] = []
+    var context: NSManagedObjectContext!
```
mateuszmackowiak, line 10: will it be edited? please make it a let

mateuszmackowiak: no sense to force unwrap it

Dashboard/AirSectionPickerView.swift
```diff
@@ -0,0 +1,55 @@
...
+            .frame(width: 300)
+            .previewLayout(.sizeThatFits)
+    }
+}
+
+enum SelectedSection: String, CaseIterable {
+    case following = "Following"
+    case mobileActive = "Mobile active"
+    case mobileDormant = "Mobile dormant"
+    case fixed = "Fixed"
```
mateuszmackowiak, line 39: are the raw string values to be displayed to the user? Maybe a good idea to use LocalilzedString(

Dashboard/DashboardView.swift
```diff
@@ -0,0 +1,72 @@
...
+                            SessionCellView(session: session)
+                        }
+                    }
+                    .padding()
+                }
+                .frame(maxWidth: .infinity)
+                .background(Color.aircastingGray.opacity(0.05))
+            }
+        }
+        .navigationBarTitle("Dashboard")
```
mateuszmackowiak, line 41: LocalizedString?

Dashboard/DashboardView.swift
```diff
@@ -0,0 +1,72 @@
...
+        .onChange(of: selectedView) { selectedSection in
+            
+            let request = NSFetchRequest<Session>(entityName: "Session")
```
mateuszmackowiak, line 44: If we have a very nice abstraction that separates the core Data from the view layer maybe also move the predicate logic to `CoreDataHook` with

Dashboard/DashboardView.swift
```diff
@@ -0,0 +1,72 @@
...
+//
+//  Created by Lunar on 01/02/2021.
+//
+
+import SwiftUI
+import CoreData
+
+struct DashboardView: View {
+    
+    @State private var selectedView = SelectedSection.mobileActive
```
mateuszmackowiak, line 13: the variable is not a view so maybe `selectedSection` would be a better name

AirCasting/CoreData/CoreDataHook.swift
```diff
@@ -0,0 +1,40 @@
...
+    
+    var fetchRequest: NSFetchRequest<Session>! {
```
mateuszmackowiak, line 12: seams like You are just editing the predicate. So maybe make it private lazy and expose the setup method that accepts the `SelectedSection` ? Also the NSFetchedResultsController can be a lazy var. just the performFetch() needs to be called after the predicate change.

AirCasting/CoreData/CoreDataHook.swift
```diff
@@ -0,0 +1,53 @@
...
+        }
+        try fetchedResultsController.performFetch()
+        fetchedResultsController.delegate = self
+        sessions = fetchedResultsController.fetchedObjects ?? []
+    }
+}
+
+extension CoreDataHook: NSFetchedResultsControllerDelegate {
+    
+    private func controllerDidChangeContent(_ controller: NSFetchedResultsController<NSFetchRequestResult>) throws {
```
mateuszmackowiak, line 43: You have changed the declaration of the function. this will not be called as a delegate method because of throws. 

It would make more sense to force unwrap fetchedResultsController in this case. IMO NSManagedObjectContext should be passed in the constructor

AirCasting/CoreData/CoreDataHook.swift
```diff
@@ -0,0 +1,54 @@
...
+        sessions = fetchedResultsController.fetchedObjects ?? []
+    }
+}
+
+enum FetchError: Error {
```
mateuszmackowiak, line 52: seams no longer used

AirCasting/CoreData/CoreDataHook.swift
```diff
@@ -0,0 +1,54 @@
...
+            request.predicate = NSPredicate(format: "type == %@", SessionType.fixed.rawValue)
+        case .mobileActive:
+            request.predicate = NSPredicate(format: "type == %@ AND status == %li", SessionType.mobile.rawValue, SessionStatus.RECORDING.rawValue)
+        case .mobileDormant:
+            request.predicate = NSPredicate(format: "type == %@ AND status == %li", SessionType.mobile.rawValue, SessionStatus.FINISHED.rawValue)
+        case .following:
+            request.predicate = NSPredicate(format: "followedAt != NULL")
+        }
+        
+        request.sortDescriptors = [NSSortDescriptor(key: "startTime", ascending: true)]
```
mateuszmackowiak, line 28: this can be set only once just before NSFetchedResultsController init so You can just capture `let predicate`

mateuszmackowiak (approved):

PR #50, opened 2021-05-24
Graph part.1 

smialko:
**What's done:** 
- presenting real data on the graph and "drawing" it in real-time,
- zooming in/out the graph,
- time labels change values when zooming/scrolling the graph.

**Note**
Please, remember it work in progress - just part 1 of the graph task is done.


AirCasting/Graph/Graph.swift
```diff
@@ -124,21 +168,36 @@ class MultiColorGridRenderer: YAxisRenderer {
...
     func makeUIView(context: Context) -> UI_PollutionGraph {
         UI_PollutionGraph()
     }
     
     func updateUIView(_ uiView: UI_PollutionGraph, context: Context) {
-       try? uiView.updateWith(thresholdValues: thresholds.rawThresholdsBinding.wrappedValue)
+        
+        try? uiView.updateWith(thresholdValues: thresholds.rawThresholdsBinding.wrappedValue)
+        
+        let entries = stream.measurements?.compactMap({ item -> ChartDataEntry? in
```
mateuszmackowiak, line 182: You can use `stream.allMeasurements` without `guard let measurement = item as? MeasurementEntity else {                return nil  }`

AirCasting/Graph/Graph.swift
```diff
@@ -59,24 +43,29 @@ class UI_PollutionGraph: UIView {
...
-        dataSet.setColor(UIColor(.white))
-        dataSet.mode = .linear
-        dataSet.lineWidth = 4
         
         renderer = MultiColorGridRenderer(viewPortHandler: lineChartView.viewPortHandler,
-                                            yAxis: lineChartView.leftAxis,
-                                            transformer: lineChartView.getTransformer(forAxis: .left))
+                                          yAxis: lineChartView.leftAxis,
+                                          transformer: lineChartView.getTransformer(forAxis: .left))
         guard let renderer = renderer else { return }
```
anna1901, line 60: Shouldn't we throw renderer error here?

smialko: True, thanks!

AirCasting/Models/Session+Mock.swift
```diff
@@ -15,8 +15,25 @@ extension SessionEntity {
         let session: SessionEntity = try! context.newOrExisting(uuid: SessionUUID(rawValue: "mock")!)
         session.type = .mobile
         // ...
+                
         return session
     }
 }
 
+extension MeasurementStreamEntity {
```
anna1901, line 23: Is there a reason why it's not in a separate file?

mateuszmackowiak (approved):

AirCasting/Models/Session+Mock.swift
smialko: Well, I guess I wanted to have mocked session and mocked stream in one file. 
Do you think I should move it elsewhere?

anna1901: It's not very intuitive to look for a MeasurementStreamEntity extension in SessionMock file

PR #51, opened 2021-05-27
Logout clearing sessions and canceling pending api requests

mateuszmackowiak:


AirCasting/AuthViews/UserAuthenticationSession.swift
```diff
@@ -29,8 +29,11 @@ final class UserAuthenticationSession: ObservableObject {
     }
 
     func deauthorize() throws {
-        self.token = nil
+        URLSession.shared.getAllTasks { tasks in
+            tasks.forEach { $0.cancel() }
+        }
```
smialko, line 34: You've already done that in the logout function, is there a reason to repeat canceling all tasks here?

smialko (commented):
It's still possible to sign in with incorrect data. Is it something we can fix or it's up to the backend part?

AirCasting/AuthViews/UserAuthenticationSession.swift
mateuszmackowiak: If there would be a pending call to api that synchronises the sessions there is no check if the user is logged in or not. so if we wouldn't cancel it the returned sessions would again be added into the database.

AirCasting/CoreData/SessionStorage.swift
```diff
@@ -0,0 +1,19 @@
...
+final class SessionStorage: ObservableObject {
+    let persistenceController: PersistenceController
+    init(persistenceController: PersistenceController) {
+        self.persistenceController = persistenceController
+    }
+
+    func clearAllSession() throws {
+        let deleteRequest = NSBatchDeleteRequest(fetchRequest: SessionEntity.fetchRequest())
+        deleteRequest.resultType = .resultTypeObjectIDs
+        let res = try persistenceController.editContext().execute(deleteRequest)
```
pawelgil, line 16: This code will not inform any `NSManagedObjectContext` instance in use about the change made to the store. Take a look at the [Apple guide](https://developer.apple.com/library/archive/featuredarticles/CoreData_Batch_Guide/BatchDeletes/BatchDeletes.html) to batch delete requests. IFK if this is the expected behavior here, just pointing this CoreData "gotcha" out.

mateuszmackowiak: good catch. in current usage it was intended but maybe not a good idea in global scope.

smialko (approved):

PR #52, opened 2021-05-27
Graph part 2.

smialko:
@anna1901 I merged graph part 1 but you asked there why `extension MeasurementStreamEntity {` isn't in a separate file. The answer is: I wanted to have mocked session and mocked stream in one file. If you think I should move it to a separate file, please write a comment in this PR :-) 

**Summary**
- I limit the number of measurements for the fixed session to 1440 (24h of measurements). I create a background context to fetch session, check if any of the session streams has more than 1440 measurements. If yes, I remove the surplus from the beginning of the array of measurements. 
- I add a zoom mobile session graph to show the last 30 minutes by setting 1800 as a maximum visible range for the X axis.

### DIFF
--- a/AirCasting/Graph/Graph.swift
+++ b/AirCasting/Graph/Graph.swift
@@ -8,7 +8,7 @@
 import SwiftUI
 import Charts
 
-class UI_PollutionGraph: UIView {
+class AirCastingGraph: UIView {
     
     let lineChartView = LineChartView()
     var renderer: MultiColorGridRenderer?
@@ -21,7 +21,7 @@ class UI_PollutionGraph: UIView {
         try? setupGraph()
     }
     
-    func setupGraph() throws {
+    private func setupGraph() throws {
         
         lineChartView.translatesAutoresizingMaskIntoConstraints = false
         NSLayoutConstraint.activate([
@@ -65,6 +65,14 @@ class UI_PollutionGraph: UIView {
         lineChartView.leftYAxisRenderer = renderer
     }
     
+    private func zoomoutToThirtyMinutes(dataSet: LineChartDataSet) {
+        let thirtyMinutesMeasurementCount = 60 * 30
+        lineChartView.setVisibleXRangeMaximum(Double(thirtyMinutesMeasurementCount))
+        lineChartView.moveViewToX(dataSet.xMax)
+        //enable zoom out
+        lineChartView.setVisibleXRangeMaximum(dataSet.xMax)
+    }
+    
     func updateWithThreshold(thresholdValues: [Float]) throws {
         guard let renderer = renderer else {
             throw GraphError.rendererError
@@ -78,7 +86,7 @@ class UI_PollutionGraph: UIView {
         lineChartView.setNeedsDisplay()
     }
     
-    func updateWithEntries(entries: [ChartDataEntry], isSessionMobile: Bool) {
+    func updateWithEntries(entries: [ChartDataEntry], isAutozoomEnabled: Bool) {
         let dataSet = LineChartDataSet(entries: entries)
         let data = LineChartData(dataSet: dataSet)
         lineChartView.data = data
@@ -92,17 +100,9 @@ class UI_PollutionGraph: UIView {
         dataSet.mode = .linear
         dataSet.lineWidth = 4
         
-        if !didMoveOrScaleGraph && isSessionMobile {
+        if !didMoveOrScaleGraph && isAutozoomEnabled {
             zoomoutToThirtyMinutes(dataSet: dataSet)
         }
-    }
-    
-    func zoomoutToThirtyMinutes(dataSet: LineChartDataSet) {
-        let thirtyMinutesMeasurementCount = 60 * 30
-        lineChartView.setVisibleXRangeMaximum(Double(thirtyMinutesMeasurementCount))
-        lineChartView.moveViewToX(dataSet.xMax)
-        //enable zoom out
-        lineChartView.setVisibleXRangeMaximum(dataSet.xMax)
     }
     
     required init?(coder: NSCoder) {
@@ -110,7 +110,7 @@ class UI_PollutionGraph: UIView {
     }
 }
 
-extension UI_PollutionGraph: ChartViewDelegate {
+extension AirCastingGraph: ChartViewDelegate {
     
     // Callbacks when the chart is scaled / zoomed via pinch zoom gesture.
     @objc func chartScaled(_ chartView: ChartViewBase, scaleX: CGFloat, scaleY: CGFloat) {
@@ -194,17 +194,17 @@ class MultiColorGridRenderer: YAxisRenderer {
 }
 
 struct Graph: UIViewRepresentable {
-    typealias UIViewType = UI_PollutionGraph
+    typealias UIViewType = AirCastingGraph
     
     @ObservedObject var stream: MeasurementStreamEntity
     @ObservedObject var thresholds: SensorThreshold
-    var isSessionMobile: Bool
+    var isAutozoomEnabled: Bool
     
-    func makeUIView(context: Context) -> UI_PollutionGraph {
-        UI_PollutionGraph()
+    func makeUIView(context: Context) -> AirCastingGraph {
+        AirCastingGraph()
     }
 
-    func updateUIView(_ uiView: UI_PollutionGraph, context: Context) {
+    func updateUIView(_ uiView: AirCastingGraph, context: Context) {
 
         try? uiView.updateWithThreshold(thresholdValues: thresholds.rawThresholdsBinding.wrappedValue)
         
@@ -213,14 +213,14 @@ struct Graph: UIViewRepresentable {
             let chartDataEntry = ChartDataEntry(x: timeInterval, y: measurement.value)
             return chartDataEntry
         }) ?? []
-        uiView.updateWithEntries(entries: entries, isSessionMobile: isSessionMobile)
+        uiView.updateWithEntries(entries: entries, isAutozoomEnabled: isAutozoomEnabled)
     }
 }
 
 #if DEBUG
-struct PollutionGraph_Previews: PreviewProvider {
+struct AirCastingGraph_Previews: PreviewProvider {
     static var previews: some View {
-        Graph(stream: .mock, thresholds: .mock, isSessionMobile: true)
+        Graph(stream: .mock, thresholds: .mock, isAutozoomEnabled: true)
     }
 }
 #endif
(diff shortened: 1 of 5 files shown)


AirCasting/Graph/Graph.swift
```diff
@@ -48,7 +50,7 @@ class UI_PollutionGraph: UIView {
         lineChartView.xAxis.drawLabelsEnabled = true
         lineChartView.xAxis.labelCount = 2
         lineChartView.extraBottomOffset = 25
-        
+
```
mateuszmackowiak, line 53: try always to keep the changes as clean as possible. this space can lead to unnecessary conflicts for others. Before pushing usually look thru Your changed and revoke committing white space in places where the cade was not touched.

AirCasting/Graph/Graph.swift
```diff
@@ -90,13 +91,37 @@ class UI_PollutionGraph: UIView {
         dataSet.setColor(UIColor(.white))
         dataSet.mode = .linear
         dataSet.lineWidth = 4
+        
+        if !didMoveOrScaleGraph && isSessionMobile {
+            zoomoutToThirtyMinutes(dataSet: dataSet)
+        }
+    }
+    
+    func zoomoutToThirtyMinutes(dataSet: LineChartDataSet) {
```
mateuszmackowiak, line 100: sepperate the private methods from public to avoid confusion of API readability.

AirCasting/Graph/Graph.swift
```diff
@@ -173,6 +198,7 @@ struct Graph: UIViewRepresentable {
     
     @ObservedObject var stream: MeasurementStreamEntity
     @ObservedObject var thresholds: SensorThreshold
+    var isSessionMobile: Bool
```
mateuszmackowiak, line 201: Until now the `UI_PollutionGraph` had no reference to any of the business models which created a stable standalone component that could be used for other functionality. and the `Graph` was the one that glue it together. Imagine using `UI_PollutionGraph` in different place for displaying anything else. Despite the name of the class which already is a little wrong because it not only represents Pollution but also microphone session data it was  a nice component. Maybe the `UI_PollutionGraph` can expose a more generic method that could be used in the graph?

mateuszmackowiak: Also maybe a name that represents what it does and not what it is used for `isSessionMobile` and what it relay doe is `zoomoutToThirtyMinutes`.

AirCasting/Services/DownloadMeasurmentsService.swift
```diff
@@ -67,6 +67,10 @@ final class DownloadMeasurementsService: MeasurementUpdatingService {
                     } catch {
                         assertionFailure("Failed to save context \(error)")
                     }
+                    PersistenceController.shared.performBackgroundTask { bgContext in
```
mateuszmackowiak, line 70: There is already a reference to `persistenceController` in this class. Why use a shared instance?

mateuszmackowiak: `DownloadMeasurementsService` could be defiantly easier to setup if ever need if `RemoveOldMeasurementsService ` would be constructed as a lazy var.

AirCasting/Services/RemoveOldMeasurementsService.swift
```diff
@@ -0,0 +1,53 @@
...
+
+
+final class RemoveOldMeasurementsService {
+    
+    enum Error: Swift.Error {
+        case sessionFetchFailed
+        case removingMeasurementsSurplusFailed
+        
+    }
+    private let twentyFourHoursMeasurementCount = 24 * 60
```
mateuszmackowiak, line 15: static ?

AirCasting/Services/RemoveOldMeasurementsService.swift
```diff
@@ -0,0 +1,53 @@
...
+    
+    func removeOldestMeasurements(context: NSManagedObjectContext, uuid: SessionUUID) {
```
mateuszmackowiak, line 17: It would be nicer with the label `in:` `removeOldestMeasurements(context: context, uuid: id)` vs `removeOldestMeasurements(in: context, uuid: id)`

AirCasting/Services/RemoveOldMeasurementsService.swift
```diff
@@ -0,0 +1,53 @@
...
+        /// In fixed sessions we need to remove measurements older than 24h, but we treat 24 not like a date, but as a sum of measurements.
+        /// We know that we have 60 measurements per hour, so we take 1440 last measurements for each stream, and we remove older than the first of them.
```
mateuszmackowiak, line 19: This comment could be nicer to have as a comment over `removeOldestMeasurements` then code completion will display it to be read by developers trying to use the method without entering the code.

AirCasting/Services/RemoveOldMeasurementsService.swift
```diff
@@ -0,0 +1,53 @@
...
+        do {
+            let session = try fetchSession(context: context, uuid: uuid)
+            removeMeasurementsSurplus(context: context, session: session)
+        } catch {
+            Log.error("Error: \(Error.removingMeasurementsSurplusFailed)")
```
mateuszmackowiak, line 24: Seams like if the error would be propagated forward, the users of `removeOldestMeasurements` whould be responsible to handle it. In some case it could end with a log but in case of a log in the upper level the log could contain more reelable context. And some users would like to know that it fail and do something more with it. For example display an error to a user, react with a fallback logic or even crash if it failure could lead to larger problems.

AirCasting/Services/RemoveOldMeasurementsService.swift
```diff
@@ -0,0 +1,53 @@
...
+        }
+    }
+    
+    private func fetchSession(context: NSManagedObjectContext, uuid: SessionUUID) throws -> SessionEntity {
```
mateuszmackowiak, line 28: Maybe use already existing
```
extension NSManagedObjectContext {
    func existingSession(uuid: SessionUUID) throws -> SessionEntity  {
```

AirCasting/Services/RemoveOldMeasurementsService.swift
```diff
@@ -0,0 +1,53 @@
...
+        let fetchedSessions = try context.fetch(request)
+        
+        guard let session = fetchedSessions.first else {
+            throw Error.sessionFetchFailed
+        }
+        return session
+    }
+    
+    private func removeMeasurementsSurplus(context: NSManagedObjectContext, session: SessionEntity) {
+        for stream in session.measurementStreams! {
```
mateuszmackowiak, line 41: And what if there is no `session.measurementStreams` ? this will crash. Please forward the error or ignore if it is ok if the session head no streams. In case of fixed sessions we currently create sessions without measurement streams so this probably would be a common case.

AirCasting/Services/RemoveOldMeasurementsService.swift
```diff
@@ -0,0 +1,53 @@
...
+            let stream: MeasurementStreamEntity = stream as! MeasurementStreamEntity
+            let measurementsCount = stream.measurements!.count
```
mateuszmackowiak, line 43: until set `stream.measurements` is nil and this is a common case.

AirCasting/Services/RemoveOldMeasurementsService.swift
```diff
@@ -0,0 +1,53 @@
...
+        
+        guard let session = fetchedSessions.first else {
+            throw Error.sessionFetchFailed
+        }
+        return session
+    }
+    
+    private func removeMeasurementsSurplus(context: NSManagedObjectContext, session: SessionEntity) {
+        for stream in session.measurementStreams! {
+            let stream: MeasurementStreamEntity = stream as! MeasurementStreamEntity
```
mateuszmackowiak, line 42: ` for stream in session.measurementStreams as? [MeasurementStreamEntity]{`  would probably be a nicer solution

AirCasting/Graph/Graph.swift
smialko: I renamed this class to `AirCastingGraph`.

AirCasting/Services/RemoveOldMeasurementsService.swift
smialko: I think, it won't crash because core data would just return an empty array. But I'm replacing this force unwrapp with `guard let`.

AirCasting/Graph/Graph.swift
mateuszmackowiak: still maybe not ideal with auto zoom enabled. In few weeks time You will not remember what it is autozooming to. It is a graph so what autozoom means exactly. But lets go with this name. Defiantly much better then isSessionMobile👍

AirCasting/Services/DownloadMeasurmentsService.swift
```diff
@@ -67,6 +70,14 @@ final class DownloadMeasurementsService: MeasurementUpdatingService {
                     } catch {
                         assertionFailure("Failed to save context \(error)")
                     }
+                    persistenceController.performBackgroundTask { [weak self] bgContext in
+                        do {
+                            try self?.removeOldService.removeOldestMeasurements(in: bgContext,
+                                                                                uuid: uuid)
```
mateuszmackowiak, line 76: please use capture list to capture strongly removeOldService and not weakly self. The logic should always happened even if self would delicate. 
```
 persistenceController.performBackgroundTask { [removeOldService] bgContext in
    do { 
      try removeOldService.removeOldestMeasurements(in: bgContext,                                                                                uuid: uuid)
}
```

AirCasting/Services/DownloadMeasurmentsService.swift
```diff
@@ -19,7 +19,10 @@ final class DownloadMeasurementsService: MeasurementUpdatingService {
     private lazy var fixedSessionService = FixedSessionAPIService(authorisationService: authorisationService)
     private var timerSink: Cancellable?
     private var lastFetchCancellableTask: Cancellable?
-
+    private lazy var removeOldService: RemoveOldMeasurementsService = {
+        RemoveOldMeasurementsService()
+    }()
```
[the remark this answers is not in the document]
smialko: I think I had a reason to do this as computed property but I can't remember and can't find it now. Anyways, changing it to "normal" lazy var.  :D

AirCasting/Services/DownloadMeasurmentsService.swift
smialko: Uh, I couldn't do that. Errors everywhere. 
But I found this solution, is that better than `[weak self]`?
 
```
    private func updateForSession(uuid: SessionUUID) {
        let syncDate = Date().addingTimeInterval(-100)
        lastFetchCancellableTask = fixedSessionService.getFixedMeasurement(uuid: uuid, lastSync: syncDate, completion: { [removeOldService, persistenceController] result in
```

mateuszmackowiak (approved):

PR #54, opened 2021-06-01
Midnight lines

smialko:
I added midnight lines to the graph. I don't know why limit lines don't reach the top, couldn't find the reason for that. If you have any ideas, please share :) 

IMG_3831

### DIFF
--- a/AirCasting/Graph/Graph.swift
+++ b/AirCasting/Graph/Graph.swift
@@ -8,191 +8,6 @@
 import SwiftUI
 import Charts
 
-class AirCastingGraph: UIView {
-    
-    let lineChartView = LineChartView()
-    var renderer: MultiColorGridRenderer?
-    var didMoveOrScaleGraph = false
-    
-    init() {
-        super.init(frame: .zero)
-        self.addSubview(lineChartView)
-        lineChartView.delegate = self
-        try? setupGraph()
-    }
-    
-    private func setupGraph() throws {
-        
-        lineChartView.translatesAutoresizingMaskIntoConstraints = false
-        NSLayoutConstraint.activate([
-            lineChartView.leadingAnchor.constraint(equalTo: self.leadingAnchor),
-            lineChartView.trailingAnchor.constraint(equalTo: self.trailingAnchor),
-            lineChartView.topAnchor.constraint(equalTo: self.topAnchor),
-            lineChartView.bottomAnchor.constraint(equalTo: self.bottomAnchor)
-        ])
-        //set edges
-        lineChartView.minOffset = 0.0
-        
-        //remove border lines and legend
-        lineChartView.xAxis.drawLabelsEnabled = false
-        
-        lineChartView.leftAxis.drawLabelsEnabled = false
-        lineChartView.leftAxis.drawAxisLineEnabled = false
-        lineChartView.leftAxis.drawGridLinesEnabled = true
-        
-        lineChartView.rightAxis.enabled = false
-        lineChartView.rightAxis.drawLabelsEnabled = false
-        lineChartView.rightAxis.drawAxisLineEnabled = false
-        
-        lineChartView.legend.enabled = false
-        lineChartView.scaleYEnabled = false
-        
-        lineChartView.xAxis.drawLabelsEnabled = true
-        lineChartView.xAxis.labelCount = 2
-        lineChartView.extraBottomOffset = 25
-                
-        lineChartView.xAxisRenderer = TimeAxisRenderer(viewPortHandler: lineChartView.viewPortHandler,
-                                                       xAxis: lineChartView.xAxis,
-                                                       transformer: lineChartView.getTransformer(forAxis: .left))
-        
-        
-        renderer = MultiColorGridRenderer(viewPortHandler: lineChartView.viewPortHandler,
-                                          yAxis: lineChartView.leftAxis,
-                                          transformer: lineChartView.getTransformer(forAxis: .left))
-        guard let renderer = renderer else {
-            throw GraphError.rendererError
-        }
-        lineChartView.leftYAxisRenderer = renderer
-    }
-    
-    private func zoomoutToThirtyMinutes(dataSet: LineChartDataSet) {
-        let thirtyMinutesMeasurementCount = 60 * 30
-        lineChartView.setVisibleXRangeMaximum(Double(thirtyMinutesMeasurementCount))
-        lineChartView.moveViewToX(dataSet.xMax)
-        //enable zoom out
-        lineChartView.setVisibleXRangeMaximum(dataSet.xMax)
-    }
-    
-    func updateWithThreshold(thresholdValues: [Float]) throws {
-        guard let renderer = renderer else {
-            throw GraphError.rendererError
-        }
-        renderer.thresholds = thresholdValues
-        
-        lineChartView.leftAxis.axisMinimum = Double(thresholdValues.first ?? 0)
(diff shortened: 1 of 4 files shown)


anna1901:
I know it's a change from different pull request but could you change force unwrap of dbstream in graph? Maybe we could redirect to that page only for this specific stream or chose always the first stream of a given session? We don't know when we are going to merge the functionality for choosing a stream and until then the app is going to crash when the user enters graph view from airbeam session

AirCasting/Graph/Graph.swift
```diff
@@ -214,6 +224,50 @@ struct Graph: UIViewRepresentable {
...
+        while (midnight > firstMeasurementDate) {
+            let previus = Calendar.current.nextDate(after: midnight,
+                                                    matching: components,
+                                                    matchingPolicy: Calendar.MatchingPolicy.previousTimePreservingSmallerComponents,
+                                                    repeatedTimePolicy: Calendar.RepeatedTimePolicy.first,
+                                                    direction: .backward)
+            
+            let formatter = DateFormatter()
+            formatter.dateStyle = .short
+            formatter.timeStyle = .short
```
mateuszmackowiak, line 248: formatter is not used

AirCasting/Graph/Graph.swift
```diff
@@ -105,6 +108,13 @@ class AirCastingGraph: UIView {
         }
     }
     
+    func updateMidnightLines(with limitLines: [ChartLimitLine]) {
+        lineChartView.xAxis.removeAllLimitLines()
```
mateuszmackowiak, line 112: Maybe store the last displayed and only update on view if the values are different?

AirCasting/Graph/Graph.swift
```diff
@@ -214,6 +224,50 @@ struct Graph: UIViewRepresentable {
...
+            if let midnightPoint = previus?.timeIntervalSince1970 {
+                midnightPoints.append(midnightPoint)
+            }
+            midnight = previus!
+        }
+        return midnightPoints
+    }
+    
+    func getLimitLines() -> [ChartLimitLine] {
+        let points = getMidnightsPoints(startingDate: stream.allMeasurements?.first?.time ?? Date())
```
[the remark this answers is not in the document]
smialko: @mateuszmackowiak Hmm, any tips on how to do that? I tried to find a way to observe midnightPoints but didn't come up with anything. Do you have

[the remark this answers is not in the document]
smialko: I feel stupid, I was trying to find a complex solution and didn't think of the simple one. :D

AirCasting/Graph/Graph.swift
```diff
@@ -105,6 +108,13 @@ class AirCastingGraph: UIView {
         }
     }
     
+    func updateMidnightLines(with limitLines: [ChartLimitLine]) {
+        lineChartView.xAxis.removeAllLimitLines()
+            for line in limitLines {
+                lineChartView.xAxis.addLimitLine(line)
```
pawelgil, line 114: lines 113 and 114 are indented too far (also sorry for being nit-picky 🤣)

AirCasting/Graph/Graph.swift
```diff
@@ -214,6 +225,48 @@ struct Graph: UIViewRepresentable {
...
+    }
+    
+    func getMidnightsPoints(startingDate: Date) -> [Double] {
+        let now = Date()
+        var midnight = Calendar.current.date(bySettingHour: 0, minute: 0, second: 0, of: now)!
+        let components = Calendar.current.dateComponents([.hour, .minute, .second], from: midnight)
+        let firstMeasurementDate = startingDate
+        var midnightPoints: [Double] = []
+        
+        while (midnight > firstMeasurementDate) {
```
pawelgil, line 243: I'm quite sure this `while` and associated helper vars can be replaced by some map/filter combo (tho it's too complex to come up with solution on the spot)

AirCasting/Graph/Graph.swift
```diff
@@ -214,6 +225,48 @@ struct Graph: UIViewRepresentable {
             return chartDataEntry
         }) ?? []
         uiView.updateWithEntries(entries: entries, isAutozoomEnabled: isAutozoomEnabled)
+        
+        let limitLines = getLimitLines()
+        if limitLines != midnightLines {
+            midnightLines = limitLines
+            uiView.updateMidnightLines(with: limitLines)
```
pawelgil, line 232: I'm wondering why the property `var midnightLines` is here in `Graph` instead of `AirCastingGraph`. I think it would be nicer to do so in `AirCastingGraph`, utilizing a `didSet` call. Note the `update` function can be made `private` in this scenario.

Inside `AirCastingGraph`:
``` Swift
var limitLines: [ChartLimitLine] = [] {
   didSet {
      guard oldValue != limitLines { return }
      updateLimitLines(with: limitLines)
   }
}

private updateLimitLines(with limitLines: [ChartLimitLine]) {
   ...
}
```

and here simply do:

```Swift
uiView.limitLines = getLimitLines()
```

As a side note: I also think anything named `midnight` should not be present in this view and be changed to `limit`. The reason is that it doesn't really need to be midnight, it can be any limit. Currently this midnights calculation logic is a stray - it has no clear and obvious home (it's not **really** UI, is it?). This is why layering would be beneficial. We'd have this nicely separated and unit tested inside some `Controller` object which would take care of fetching stuff from DB/Service and translating it into limits (midnights).

AirCasting/Graph/Graph.swift
```diff
@@ -214,6 +225,48 @@ struct Graph: UIViewRepresentable {
...
+            midnight = previus!
+        }
+        return midnightPoints
+    }
+    
+    func getLimitLines() -> [ChartLimitLine] {
+        let points = getMidnightsPoints(startingDate: stream.allMeasurements?.first?.time ?? Date())
+        var limitLines: [ChartLimitLine] = []
+        
+        for point in points {
```
pawelgil, line 261: It's generally considered a good practice in Swift to use [higher order functions](https://engineering.nodesagency.com/categories/ios/2019/07/03/Higher-Order-Functions-In-Swift) instead of classic `for/while/do`. In this example I recommend using the [map(:_)](https://developer.apple.com/documentation/swift/array/3017522-map):
``` Swift
return points.map {
   let line = ChartLimitLine(limit: $0)
   line.lineColor = UIColor.aircastingDarkGrey
   ...
}
```

AirCasting/Graph/GraphView.swift
```diff
@@ -21,7 +21,7 @@ struct GraphView: View {
             
             ZStack(alignment: .topLeading) {
                 #warning("Replace dbStream with currently selected")
-                Graph(stream: session.dbStream!,
+                Graph(stream: session.measurementStreams?[0] as! MeasurementStreamEntity,
```
pawelgil, line 24: I'm new, is it guaranteed at this point that `measurementStreams` is not empty?

pawelgil (commented):
looking good, some stylistic stuff found

AirCasting/Graph/GraphView.swift
smialko: At this moment it should be guaranteed. We shouldn't be able to expand the session cart where the button triggering graph view is.
I'm attaching the design of the cart without measurements:
https://projects.invisionapp.com/d/main#/console/19745972/414476839/preview

AirCasting/Graph/Graph.swift
smialko: Info for others: I talked with Paweł and decided to leave this code as it is for now :)

pawelgil (approved):

PR #55, opened 2021-06-04
Averaging

smialko:
Please, check only graph averaging with SwiftSimplify. 
The rest is work in progress, and it might be removed if everyone agrees to use SwiftSimplify :-) 

Documentation: https://github.com/malcommac/SwiftSimplify

### DIFF
--- a/AirCasting/Graph/Graph.swift
+++ b/AirCasting/Graph/Graph.swift
@@ -7,6 +7,20 @@
 
 import SwiftUI
 import Charts
+import SwiftSimplify
+
+extension ChartDataEntry: Point2DRepresentable {
+    public var xValue: Float {
+        Float(self.x)
+    }
+    
+    public var yValue: Float {
+        Float(self.y)
+    }
+    public var cgPoint: CGPoint {
+        .init(x: CGFloat(xValue), y: CGFloat(yValue))
+    }
+}
 
 struct Graph: UIViewRepresentable {
     typealias UIViewType = AirCastingGraph
@@ -14,13 +28,14 @@ struct Graph: UIViewRepresentable {
     @ObservedObject var stream: MeasurementStreamEntity
     @ObservedObject var thresholds: SensorThreshold
     var isAutozoomEnabled: Bool
+    let simplifiedGraphEntryThreshold = 1000
     
     func makeUIView(context: Context) -> AirCastingGraph {
         AirCastingGraph()
     }
-
+    
     func updateUIView(_ uiView: AirCastingGraph, context: Context) {
-
+        
         try? uiView.updateWithThreshold(thresholdValues: thresholds.rawThresholdsBinding.wrappedValue)
         
         let entries = stream.allMeasurements?.compactMap({ measurement -> ChartDataEntry? in
@@ -28,10 +43,21 @@ struct Graph: UIViewRepresentable {
             let chartDataEntry = ChartDataEntry(x: timeInterval, y: measurement.value)
             return chartDataEntry
         }) ?? []
-        uiView.updateWithEntries(entries: entries, isAutozoomEnabled: isAutozoomEnabled)
-        
         let allLimitLines = getLimitLines()
-        uiView.limitLines = allLimitLines
+        uiView.limitLines = allLimitLines        
+        simplifyGraphline(entries: entries, uiView: uiView)
+    }
+    
+    private func simplifyGraphline(entries: [ChartDataEntry], uiView: AirCastingGraph) {
+        if entries.count > simplifiedGraphEntryThreshold {
+            let simplifiedPoints = SwiftSimplify.simplify(entries,
+                                                          tolerance: 0.000000001,
+                                                          highestQuality: true)
+            uiView.updateWithEntries(entries: simplifiedPoints, isAutozoomEnabled: isAutozoomEnabled)
+            print("Simplified \(entries.count) to \(simplifiedPoints.count)")
+        } else {
+            uiView.updateWithEntries(entries: entries, isAutozoomEnabled: isAutozoomEnabled)
+        }
     }
     
     func getMidnightsPoints(startingDate: Date) -> [Double] {
--- a/AirCasting.xcodeproj/project.pbxproj
+++ b/AirCasting.xcodeproj/project.pbxproj
@@ -125,6 +125,7 @@
 		ACD01F952600A54400B8F65F /* UserDefaults+Preferences.swift in Sources */ = {isa = PBXBuildFile; fileRef = ACD01F942600A54400B8F65F /* UserDefaults+Preferences.swift */; };
 		ACD32A4F2600A8DD00818946 /* SettingsView.swift in Sources */ = {isa = PBXBuildFile; fileRef = ACD32A4E2600A8DD00818946 /* SettingsView.swift */; };
 		ACDF82FC265BBDE200A4D04A /* UIFont.swift in Sources */ = {isa = PBXBuildFile; fileRef = ACDF82FB265BBDE200A4D04A /* UIFont.swift */; };
+		ACF600BE2678D7F4005DD97C /* SwiftSimplify in Frameworks */ = {isa = PBXBuildFile; productRef = ACF600BD2678D7F4005DD97C /* SwiftSimplify */; };
 		ACF77DFB25A8888700BB85B7 /* SessionCellView.swift in Sources */ = {isa = PBXBuildFile; fileRef = ACF77DFA25A8888700BB85B7 /* SessionCellView.swift */; };
 		ACF7C05F260CC9FF007C24AE /* DownloadMeasurmentsService.swift in Sources */ = {isa = PBXBuildFile; fileRef = ACF7C05E260CC9FF007C24AE /* DownloadMeasurmentsService.swift */; };
 		ACFC708925C1A36A00D609D9 /* UIColor.swift in Sources */ = {isa = PBXBuildFile; fileRef = ACFC708825C1A36A00D609D9 /* UIColor.swift */; };
@@ -290,6 +291,7 @@
 			buildActionMask = 2147483647;
 			files = (
 				AC6D567C25AC540D00D221BA /* Charts in Frameworks */,
+				ACF600BE2678D7F4005DD97C /* SwiftSimplify in Frameworks */,
 				F76B6F660F0BD398629EDD83 /* Pods_AirCasting.framework in Frameworks */,
 				AC787D75267B41990034B222 /* AirCastingStyling in Frameworks */,
 			);
@@ -709,6 +711,7 @@
 			packageProductDependencies = (
 				AC6D567B25AC540D00D221BA /* Charts */,
 				AC787D74267B41990034B222 /* AirCastingStyling */,
+				ACF600BD2678D7F4005DD97C /* SwiftSimplify */,
 			);
 			productName = AirCasting;
 			productReference = AC7EB37225A6FB3A00A7F2AF /* AirCasting.app */;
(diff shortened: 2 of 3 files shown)


AirCasting/Graph/GraphDataGenerator.swift
```diff
@@ -0,0 +1,43 @@
...
+        for measurement in allMeasurements {
+            fill += fillFactor
+            
+            if fill > 1 {
+                fill = 0.0
+            }
+        }
+        
+        
+
```
mateuszmackowiak, line 31: Is the spacing really needed?

AirCasting/Graph/GraphDataGenerator.swift
```diff
@@ -0,0 +1,43 @@
+// Created by Lunar on 01/06/2021.
+//
+
+import Foundation
+
+class GraphDataGeneraator {
```
mateuszmackowiak, line 6: final, double aa

AirCasting/SessionViews/ChartEntriesCreator.swift
```diff
@@ -0,0 +1,47 @@
+// Created by Lunar on 05/05/2021.
+//
+
+import Foundation
+import Charts
+
+final class ChartEntriesCreator {
+    var entries: [ChartDataEntry] = []
```
mateuszmackowiak, line 8: private(set)

AirCasting/SessionViews/ChartEntriesCreator.swift
```diff
@@ -0,0 +1,47 @@
...
+    var stream: MeasurementStreamEntity
```
mateuszmackowiak, line 9: private(set)

AirCasting/SessionViews/ChartEntriesCreator.swift
```diff
@@ -0,0 +1,47 @@
...
+    init(stream: MeasurementStreamEntity) {
+        self.stream = stream
+    }
+    
+    func generateEntries() -> [ChartDataEntry] {
+        guard let lastMeasurementTime = lastMeasurementTime() else {
+            return []
+        }
+        
+        let sessionStartTime = stream.session.startTime!
```
mateuszmackowiak, line 24: I think that we don't set startTime in all cases. if we do then maybe better to force unwrap the value in declaration and not in all usages.

AirCasting/SessionViews/ChartView.swift
```diff
@@ -9,10 +9,11 @@ import SwiftUI
 import Charts
 
 class UI_PollutionChart: UIView {
+    let lineChartView = LineChartView()
```
mateuszmackowiak, line 12: private ?

AirCasting/SessionViews/ChartView.swift
```diff
@@ -70,19 +58,79 @@ class UI_PollutionChart: UIView {
...
     
     func updateUIView(_ uiView: UI_PollutionChart, context: Context) {
+        let chartCreator = ChartEntriesCreator(stream: stream)
+        var entries = chartCreator.generateEntries()
+        
+        if entries.isEmpty {
+            return()
+        } else {
+            entries.sort { (e1, e2) -> Bool in
+                e1.x < e2.x
```
mateuszmackowiak, line 78: why exactly we sort this by x? this seams strange

AirCasting/SessionViews/ChartView.swift
```diff
@@ -70,19 +58,79 @@ class UI_PollutionChart: UIView {
...
+        
+        //format data labels
+        formatData(data: data)
+        
+        //format line and dots
+        formatDataSet(dataSet: dataSet)
+    }
+    
+    private func formatData(data: LineChartData) {
+        let formatter = NumberFormatter()
```
mateuszmackowiak, line 94: Please move the DefaultValueFormatter to be lazy var.

AirCasting/SessionViews/ChartView.swift
```diff
@@ -70,19 +58,79 @@ class UI_PollutionChart: UIView {
...
+        dataSet.circleRadius = 6
+        
+        //line color
+        dataSet.setColor(UIColor.aircastingGray.withAlphaComponent(0.7))
+    }
+    
+    private func generateColorsSet(for entries: [ChartDataEntry]) -> [UIColor] {
+        var colors: [UIColor] = []
+        for entry in entries {
+            switch Int32(entry.y) {
```
mateuszmackowiak, line 114: there seams to be a common problem to have this values of different types and we seams to change this very freely but the int conversion can lead to strange bugs in strange places. 
On a 32-bit platform, Int is the same size as Int32.
On a 64-bit platform, Int is the same size as Int64.

AirCasting/SessionViews/ChartView.swift
```diff
@@ -70,19 +58,79 @@ class UI_PollutionChart: UIView {
...
+            case thresholds.thresholdVeryLow..<thresholds.thresholdLow:
+                colors.append(UIColor.aircastingGreen.withAlphaComponent(0.5))
+            case thresholds.thresholdLow..<thresholds.thresholdMedium:
+                colors.append(UIColor.aircastingYellow.withAlphaComponent(0.5))
+            case thresholds.thresholdMedium..<thresholds.thresholdHigh:
+                colors.append(UIColor.aircastingOrange.withAlphaComponent(0.5))
+            default:
+                colors.append(UIColor.aircastingRed.withAlphaComponent(0.5))
```
mateuszmackowiak, line 122: please define lazy vars out of `UIColor.aircastingRed.withAlphaComponent(0.5)` to avoid multiple allocations of same color.

AirCasting/SessionViews/SessionCellView.swift
```diff
@@ -27,7 +28,10 @@ struct SessionCellView: View {
             thresholds: Array(thresholds))
             if !isCollapsed {
                 VStack(alignment: .trailing, spacing: 40) {
-                    pollutionChart
+                    #warning("The stream should be chosen based on users selection")
+                    if let streams = session.measurementStreams {
```
mateuszmackowiak, line 32: Maybe we can have a helper method like in measurement:
```
 public var allMeasurements: [MeasurementEntity]? {
        measurements?.array as? [MeasurementEntity]
    }
```
to get the list of allMeasurementStreams where the `as!` would not be needed.

AirCasting/SessionViews/SessionCellView.swift
```diff
@@ -27,7 +28,10 @@ struct SessionCellView: View {
...
+                        pollutionChart(stream: streams[0] as! MeasurementStreamEntity)
```
mateuszmackowiak, line 33: There are cases where the stream is not created so assuming the array has elements here is not safe

AirCasting/SessionViews/SessionCellView.swift
```diff
@@ -58,6 +58,12 @@ private extension SessionCellView {
             Text("map")
         }
     }
+    
+    func pollutionChart(stream: MeasurementStreamEntity) -> some View {
```
mateuszmackowiak, line 62: this is not exactly pollution in all cases. is this proper name?

AirCasting/Graph/Graph.swift
```diff
@@ -213,7 +238,61 @@ struct Graph: UIViewRepresentable {
...
+        while (midnight > firstMeasurementDate) {
+            let previus = Calendar.current.nextDate(after: midnight,
+                                                    matching: components,
+                                                    matchingPolicy: Calendar.MatchingPolicy.previousTimePreservingSmallerComponents,
+                                                    repeatedTimePolicy: Calendar.RepeatedTimePolicy.first,
+                                                    direction: .backward)
+            
+            let formatter = DateFormatter()
+            formatter.dateStyle = .short
+            formatter.timeStyle = .short
```
mateuszmackowiak, line 273: Still the date formatter is unused and initialised every second

smialko: Fixed it in https://github.com/HabitatMap/AirCastingiOS/pull/54

AirCasting/Graph/Graph.swift
```diff
@@ -213,7 +238,61 @@ struct Graph: UIViewRepresentable {
             let chartDataEntry = ChartDataEntry(x: timeInterval, y: measurement.value)
             return chartDataEntry
         }) ?? []
-        uiView.updateWithEntries(entries: entries, isAutozoomEnabled: isAutozoomEnabled)
+        
+        if entries.count > 1000 {
```
pawelgil, line 242: it would be nice to put this 1000 into a `let` with a descriptive name (what it actually stands for?). This way it will be easier to understand what this line means without reading further down.

pawelgil: I'd also extract this whole `if/else` into a separate function and inside I'd invert the condition using guard:
``` Swift
func cleverNameHere(entries: [Entry], ...) {
   guard entries <= 1000 else {
      return
   }
   let simplifiedPoints = SwiftSimplify.simplify(entries, tolerance: 0.000000001, highestQuality: true)
   ...
}
```

pawelgil (commented):
I only reviewed the last commit. Only stylistic suggestions

AirCasting/Graph/GraphDataGenerator.swift
smialko: There's no GraphDataGenerator anymore. I don't know why github didn't mark this comment as outdated.

AirCasting/Graph/GraphDataGenerator.swift
smialko: There's no GraphDataGenerator anymore. I don't know why github didn't mark this comment as outdated.

AirCasting/Graph/Graph.swift
```diff
@@ -203,17 +227,75 @@ struct Graph: UIViewRepresentable {
...
-        uiView.updateWithEntries(entries: entries, isAutozoomEnabled: isAutozoomEnabled)
+        
+        simplifyGraphline(entries: entries, uiView: uiView)
+        
+        let limitLines = getLimitLines()
+        uiView.updateMidnightLines(with: limitLines)
+    }
+    
+    private func simplifyGraphline(entries: [ChartDataEntry], uiView: AirCastingGraph) {
+        let DEFAULT_LIMIT = 1000
```
pawelgil, line 248: interesting that you decided to make it snake upper-cased. This is some oldschool `♥️Objective-C♥️` stuff over here ;) But usually this kind of `REALLY_CONSTANT_STUFF` would go to the class scope:
``` Swift
class MegaTurboSuperToolbar: UIToolbar {
   let CONSTANT_STUFF = 999

   ...
}
```

I'm not sure I dislike this tho. @mateuszmackowiak what do you think?

smialko: Well tbh, I was trying to come up with something but all names were looooooong. So I took a look at this name in Android app and copied the idea. ;D

smialko: I can move it to class scope :)

pawelgil: I think long names are not bad in such cases, we're way passed 80-char wide terminals so we can get a bit more verbose than `default limit` which frankly is not much better than putting a bare `1000` here :) I'd say let's make it `snakeCase`, class scoped and named verbosely like `let simplifiedGraphEntryThreshold = 1000` ?

smialko: Thanks, I'll do that! :D

AirCasting/Graph/GraphDataGenerator.swift
mateuszmackowiak: probably just removed form Xcode reference and not from file system.

mateuszmackowiak (approved):

PR #56, opened 2021-06-09
wifi_popup

jankrzempek:
https://trello.com/c/uEXckYam/72-fixed-wifi-session-if-wifi-ssid-and-password-is-empty-user-cant-go-to-the-confirmationview-there-should-be-alert-informing-why

### DIFF
--- a/AirCasting/CreateSessionViews/CreateSessionDetailsView.swift
+++ b/AirCasting/CreateSessionViews/CreateSessionDetailsView.swift
@@ -5,8 +5,8 @@
 //  Created by Anna Olak on 24/02/2021.
 //
 
-import SwiftUI
 import CoreLocation
+import SwiftUI
 
 struct CreateSessionDetailsView: View {
     let sessionCreator: SessionCreator
@@ -18,12 +18,13 @@ struct CreateSessionDetailsView: View {
     @State var wifiPassword: String = ""
     @State var wifiSSID: String = ""
     @State private var isConfirmCreatingSessionActive: Bool = false
+    @State private var showingAlert = false
     @EnvironmentObject private var sessionContext: CreateSessionContext
     // Location tracker is needed to get wifi SSID (more info CNCopyCurrentNetworkInfo documentation.
     @StateObject private var locationTracker = LocationTracker()
-    
-    @Binding var creatingSessionFlowContinues : Bool
-    
+
+    @Binding var creatingSessionFlowContinues: Bool
+
     var body: some View {
         GeometryReader { geometry in
             ScrollView(.vertical) {
@@ -47,12 +48,11 @@ struct CreateSessionDetailsView: View {
                 .frame(maxWidth: .infinity, minHeight: geometry.size.height, alignment: .top)
             }
         }
-                .simultaneousGesture(
-        
+        .simultaneousGesture(
             DragGesture(minimumDistance: 2, coordinateSpace: .global)
-                .onChanged({ (_) in
+                .onChanged { _ in
                     UIApplication.shared.sendAction(#selector(UIResponder.resignFirstResponder), to: nil, from: nil, for: nil)
-                })
+                }
         )
     }
 }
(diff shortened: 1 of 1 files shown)


AirCasting/CreateSessionViews/CreateSessionDetailsView.swift
```diff
@@ -47,54 +48,68 @@ struct CreateSessionDetailsView: View {
...
             }
             getAndSaveStartingLocation()
             isConfirmCreatingSessionActive = true
-            #warning("TODO: handle wifi empty password")
-            if wifiSSID != "" && wifiPassword != "" {
+            if wifiSSID != "", wifiPassword != "" {
                 sessionContext.wifiSSID = wifiSSID
                 sessionContext.wifiPassword = wifiPassword
+                showingAlert = false
+            } else if isWiFi, wifiSSID == "", wifiPassword == "" {
```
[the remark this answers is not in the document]
mateuszmackowiak: Seams like the check for empty is made few times. Imo it also is ok to have an empty wifiPassword. Some users don't have a wifipassword

AirCasting/CreateSessionViews/CreateSessionDetailsView.swift
```diff
@@ -47,54 +48,68 @@ struct CreateSessionDetailsView: View {
...
-                })
+                }
         )
     }
 }
 
 private extension CreateSessionDetailsView {
     var continueButton: some View {
         Button(action: {
+            showingAlert = false
```
smialko, line 63: I think you don't have to do `showingAlert = false` here. You create it with an initial value so at this point it's still false. if we change it to true and then dismiss the alert it'll change the value back to false.

AirCasting/CreateSessionViews/CreateSessionDetailsView.swift
```diff
@@ -47,54 +48,68 @@ struct CreateSessionDetailsView: View {
...
                 sessionContext.isIndoor = isIndoor
             }
             getAndSaveStartingLocation()
             isConfirmCreatingSessionActive = true
-            #warning("TODO: handle wifi empty password")
-            if wifiSSID != "" && wifiPassword != "" {
+            if wifiSSID != "", wifiPassword != "" {
                 sessionContext.wifiSSID = wifiSSID
                 sessionContext.wifiPassword = wifiPassword
+                showingAlert = false
```
smialko, line 74: Same as above :)

AirCasting/CreateSessionViews/WifiPopupView.swift
```diff
@@ -29,9 +29,13 @@ struct WifiPopupView: View {
...
             VStack(spacing: 10) {
-                Button("Connect") {
-                    presentationMode.wrappedValue.dismiss()
-                } .buttonStyle(BlueButtonStyle())
+                if wifiSSID != "" && wifiPassword != "" {
+                    confirmationButton
+                } else {
+                    confirmationButton
+                    .accentColor(.gray)
+                    .disabled(true)
```
smialko, line 37: Considering what @mateuszmackowiak wrote, there's a possibility not to have a wifi password. So I guess we shouldn't disable connect button when the password is empty. :<

AirCasting/CreateSessionViews/WifiPopupView.swift
```diff
@@ -66,6 +70,12 @@ struct WifiPopupView: View {
             isSSIDTextfieldDisplayed = true
         }
     }
+    
+    var confirmationButton: some View {
```
smialko, line 74: What do you think about renaming it to `connectButton`? 
I'm not good with the names, it's just an idea, please ignore it if you prefer it this way :D

AirCasting/CreateSessionViews/WifiPopupView.swift
smialko: But you can leave some of the logic, and check `if wifiSSID.isEmpty` 🤔

AirCasting/CreateSessionViews/CreateSessionDetailsView.swift
```diff
@@ -120,15 +135,15 @@ private extension CreateSessionDetailsView {
                 Text("Wi-Fi").tag(true)
             }
             .pickerStyle(SegmentedPickerStyle())
-            .onChange(of: isWiFi) { (_) in
+            .onChange(of: isWiFi) { _ in
```
smialko, line 138: nice catch 🙃

AirCasting/CreateSessionViews/CreateSessionDetailsView.swift
```diff
@@ -47,54 +48,68 @@ struct CreateSessionDetailsView: View {
...
+            .alert(isPresented: $showingAlert, content: {
+                Alert(title: Text("Wi-Fi credentials are empty "),
+                      message: Text("Do you want to pop up Wi-Fi screen?"),
+                      primaryButton: .default(Text("Show Wi-fi screen")) {
+                          isWifiPopupPresented = true
+                      },
+                      secondaryButton: .default(Text("Cancel")))
+            })
+            .background(Group {
+                if (!isWiFi && wifiSSID == "" && wifiPassword == "") || (!isWiFi && wifiSSID != "" && wifiPassword != "") || (isWiFi && wifiSSID != "" && wifiPassword != "") {
```
smialko, line 93: I think we don't need it here. 
NavigationLink is inactive until `isConfirmCreatingSessionActive = true` and we check the wifi credentials in a button action. 🙂

smialko (changes requested):
😎

pawelgil:
Let's maybe squash commits (will require push -f) to leave one meaningful batch of changes for the sake of a clean git history?

jankrzempek:
> Let's maybe squash commits (will require push -f) to leave one meaningful batch of changes for the sake of a clean git history?



> Let's maybe squash commits (will require push -f) to leave one meaningful batch of changes for the sake of a clean git history?

I did it, but for the first time and not sure if it's like it should be

smialko (approved):
👌🤠

PR #57, opened 2021-06-11
Select stream

smialko:
_Note: I don't have the AB3 to test this functionality._ 

**SUMMARY**
- You can select a stream on the session cart,
- Selected stream should have borderline around measurement value, and it should has the same colour as the dot,
- Chart, Graph, and Map will display data for selected stream, 
- Thresholds used for Chart, Graph, and Map depends on selected stream.

### DIFF
--- a/AirCasting/CoreData/SessionEntity.swift
+++ b/AirCasting/CoreData/SessionEntity.swift
@@ -76,6 +76,17 @@ public class SessionEntity: NSManagedObject, Identifiable {
         get { SessionType(rawValue:(value(forKey: "type") as! String)) }
         set { setValue(newValue.rawValue, forKey: "type") }
     }
+    
+    public var allStreams: [MeasurementStreamEntity]? {
+        measurementStreams?.array as? [MeasurementStreamEntity]
+    }
+    
+    func streamWith(sensorName: String) -> MeasurementStreamEntity? {
+        return self.allStreams?.first { stream in
+            stream.sensorName == sensorName
+        }
+    }
+    
 }
 
 extension NSFetchRequest where ResultType == SessionEntity {
--- a/AirCasting/CoreData/SessionEntity.swift
+++ b/AirCasting/CoreData/SessionEntity.swift
@@ -82,7 +82,7 @@ public class SessionEntity: NSManagedObject, Identifiable {
     }
     
     func streamWith(sensorName: String) -> MeasurementStreamEntity? {
-        return self.allStreams?.first { stream in
+       allStreams?.first { stream in
             stream.sensorName == sensorName
         }
     }
--- a/AirCasting/Graph/GraphView.swift
+++ b/AirCasting/Graph/GraphView.swift
@@ -10,33 +10,33 @@ import SwiftUI
 struct GraphView: View {
     
     let session: SessionEntity
-    let thresholds: [SensorThreshold]
-    @Binding var selectedStream: String
+    let threshold: SensorThreshold
+    @Binding var selectedStream: MeasurementStreamEntity?
     
     var body: some View {
         VStack(alignment: .trailing) {
             SessionHeaderView(action: {},
                               isExpandButtonNeeded: false,
                               session: session,
-                              thresholds: thresholds,
+                              threshold: threshold,
                               selectedStream: $selectedStream).padding()
             
             ZStack(alignment: .topLeading) {
-                if let selectedStream = session.streamWith(sensorName: selectedStream) {
+                if let selectedStream = selectedStream {
                     Graph(stream: selectedStream,
-                          thresholds: thresholds[0],
+                          thresholds: threshold,
                           isAutozoomEnabled: session.type == .mobile)
                 }
                 StatisticsContainerView()
             }
             
-            NavigationLink(destination: HeatmapSettingsView(changedThresholdValues: thresholds[0].rawThresholdsBinding)) {
+            NavigationLink(destination: HeatmapSettingsView(changedThresholdValues: threshold.rawThresholdsBinding)) {
                 EditButtonView()
             }
             .padding(.horizontal)
             .padding(.top)
             
-            ThresholdsSliderView(threshold: thresholds[0])
+            ThresholdsSliderView(threshold: threshold)
                 .padding()
                 // Fixes labels covered by tabbar
                 .padding(.bottom)
@@ -50,7 +50,7 @@ struct GraphView: View {
 struct GraphView_Previews: PreviewProvider {
     static var previews: some View {
         #warning("Change selected stream")
-        GraphView(session: .mock, thresholds: [.mock], selectedStream: .constant("db"))
+        GraphView(session: .mock, threshold: .mock, selectedStream: .constant(nil))
     }
 }
 #endif
--- a/AirCasting/Graph/GraphView.swift
+++ b/AirCasting/Graph/GraphView.swift
@@ -22,10 +22,11 @@ struct GraphView: View {
                               selectedStream: $selectedStream).padding()
             
             ZStack(alignment: .topLeading) {
-                #warning("Replace dbStream with currently selected")
-                Graph(stream: session.dbStream!,
-                      thresholds: thresholds[0],
-                      isAutozoomEnabled: session.type == .mobile)
+                if let selectedStream = session.streamWith(sensorName: selectedStream) {
+                    Graph(stream: selectedStream,
+                          thresholds: thresholds[0],
+                          isAutozoomEnabled: session.type == .mobile)
+                }
                 StatisticsContainerView()
             }
             
--- a/AirCasting/Graph/GraphView.swift
+++ b/AirCasting/Graph/GraphView.swift
@@ -11,13 +11,15 @@ struct GraphView: View {
     
     let session: SessionEntity
     let thresholds: [SensorThreshold]
+    @Binding var selectedStream: String
     
     var body: some View {
         VStack(alignment: .trailing) {
             SessionHeaderView(action: {},
                               isExpandButtonNeeded: false,
                               session: session,
-                              thresholds: thresholds).padding()
+                              thresholds: thresholds,
+                              selectedStream: $selectedStream).padding()
             
             ZStack(alignment: .topLeading) {
                 #warning("Replace dbStream with currently selected")
@@ -46,7 +48,8 @@ struct GraphView: View {
 #if DEBUG
 struct GraphView_Previews: PreviewProvider {
     static var previews: some View {
-        GraphView(session: .mock, thresholds: [.mock])
+        #warning("Change selected stream")
+        GraphView(session: .mock, thresholds: [.mock], selectedStream: .constant("db"))
     }
 }
 #endif
--- a/AirCasting/Map/AirMapView.swift
+++ b/AirCasting/Map/AirMapView.swift
@@ -11,12 +11,12 @@ import Foundation
 import CoreData
 
 struct AirMapView: View {
-    var thresholds: [SensorThreshold]
+    var threshold: SensorThreshold
     @ObservedObject var session: SessionEntity
-    @Binding var selectedStream: String
+    @Binding var selectedStream: MeasurementStreamEntity?
     
     private var pathPoints: [PathPoint] {
-        if let measurementStream = session.streamWith(sensorName: selectedStream) {
+        if let measurementStream = selectedStream {
             return measurementStream.allMeasurements?.compactMap {
                 if let location = $0.location {
                     return PathPoint(location: location, measurement: $0.value)
@@ -34,17 +34,17 @@ struct AirMapView: View {
             SessionHeaderView(action: {},
                               isExpandButtonNeeded: false,
                               session: session,
-                              thresholds: thresholds,
+                              threshold: threshold,
                               selectedStream: $selectedStream)
             ZStack(alignment: .topLeading) {
                 GoogleMapView(pathPoints: pathPoints,
-                              thresholds: thresholds[0])
+                              thresholds: threshold)
                 StatisticsContainerView()
             }
-            NavigationLink(destination: HeatmapSettingsView(changedThresholdValues: thresholds[0].rawThresholdsBinding)) {
+            NavigationLink(destination: HeatmapSettingsView(changedThresholdValues: threshold.rawThresholdsBinding)) {
                 EditButtonView()
             }
-            ThresholdsSliderView(threshold: thresholds[0])
+            ThresholdsSliderView(threshold: threshold)
                 // Fixes labels covered by tabbar
                 .padding(.bottom)
         }
@@ -56,8 +56,7 @@ struct AirMapView: View {
 #if DEBUG
 struct Map_Previews: PreviewProvider {
     static var previews: some View {
-        #warning("Change selected stream")
-        AirMapView(thresholds: [SensorThreshold.mock], session: .mock, selectedStream: .constant("db"))
+        AirMapView(threshold: .mock, session: .mock, selectedStream: .constant(nil))
     }
 }
 #endif
--- a/AirCasting/Map/AirMapView.swift
+++ b/AirCasting/Map/AirMapView.swift
@@ -16,17 +16,11 @@ struct AirMapView: View {
     @Binding var selectedStream: MeasurementStreamEntity?
     
     private var pathPoints: [PathPoint] {
-        if let measurementStream = selectedStream {
-            return measurementStream.allMeasurements?.compactMap {
-                if let location = $0.location {
-                    return PathPoint(location: location, measurement: $0.value)
-                } else {
-                    #warning("TODO: Do something with no location points")
-                    return nil
-                }
-            } ?? []
-        }
-        return []
+        return selectedStream?.allMeasurements?.compactMap {
+            #warning("TODO: Do something with no location points")
+            guard let location = $0.location else { return nil }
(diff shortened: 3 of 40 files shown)


AirCasting/SessionViews/MeasurementsView.swift
```diff
@@ -0,0 +1,104 @@
...
+        
+        let toShow = allStreams.compactMap { $0 }
+        return toShow
+    }
+    private var hasAnyMeasurements: Bool {
+        streamsToShow.filter { $0.latestValue != nil }.count > 0
+    }
+    
+    var body: some View {
+        if hasAnyMeasurements {
```
mateuszmackowiak, line 27: please capture streamsToShow as let and move the logic from hasAnyMeasurements so the `streamsToShow` would not be called mupllitplle times.

AirCasting/SessionViews/MeasurementsView.swift
```diff
@@ -0,0 +1,104 @@
...
+    
+    func showStreamName() -> String {
+        guard let streamName = stream.sensorName else {return ""}
+        return streamName
+            .drop { $0 != "-" }
+            .replacingOccurrences(of: "-", with: "")
+    }
+    
+    func colorBorder(stream: MeasurementStreamEntity) -> Color {
+        switch Int32(value) {
```
mateuszmackowiak, line 91: there are multiple changes to compare trashholds and values into Int32 etc. Can we unify the types so no potential insecure projections would be made?

AirCasting/SessionViews/SessionCellView.swift
```diff
@@ -47,21 +55,39 @@ struct SessionCellView: View {
...
+                }
+            }
         }
     }
     
     func pollutionChart(stream: MeasurementStreamEntity) -> some View {
-        ChartView(stream: stream, thresholds: thresholds[0])
-            .frame(height: 200)
+        Group {
+            if let threshold = thresholdFor(selectedStream: selectedStream) {
```
mateuszmackowiak, line 85: seams like evaluating `thresholdFor(selectedStream` is done many times. Please evaluate it once.

AirCasting/SessionViews/MeasurementsView.swift
```diff
@@ -0,0 +1,104 @@
...
+    var body: some View {
+        VStack(spacing: 3) {
+            #warning("Handle optional stream name")
+            Text(showStreamName())
+                .font(Font.system(size: 13))
+            
+            Button(action: {
+                selectedStream = stream
+            }, label: {
+                HStack(spacing: 3){
```
jankrzempek, line 71: Maybe space between HStack(spacing: 3) and brace? I am only suggesting, it's really hard to make review to people who knows much more then me 😅

AirCasting/SessionViews/MeasurementsView.swift
```diff
@@ -0,0 +1,104 @@
...
+                        .font(Font.moderate(size: 14, weight: .regular))
+                }
+            })
+            .buttonStyle(AirBorderedButtonStyle(isSelected: selectedStream == stream,
+                                                thresholdColor: colorBorder(stream: stream)))
+        }
+    }
+    
+    func showStreamName() -> String {
+        guard let streamName = stream.sensorName else {return ""}
```
jankrzempek, line 84: Here: {return ""} to maybe { return " } to stay consistent? Still suggesting only

AirCasting/Utils/ButtonStyles/AirBorderedButtonStyle.swift
```diff
@@ -0,0 +1,31 @@
+// Created by Lunar on 08/06/2021.
+//
+
+import SwiftUI
+
+struct AirBorderedButtonStyle: ButtonStyle {
```
mateuszmackowiak, line 6: can you explain what exactly is `airboarded`?

AirCasting/SessionViews/SessionCellView.swift
```diff
@@ -13,29 +13,37 @@ import Charts
...
-                    if let streams = session.measurementStreams {
-                        pollutionChart(stream: streams[0] as! MeasurementStreamEntity)
+                    if let selectedStream = selectedStream {
+                        pollutionChart(stream: selectedStream)
                     }
                     buttons
                 }
             }
         }
+        .onAppear{
```
jankrzempek, line 44: .onAppear{ with space maybe

jankrzempek (commented):
As I am new here and overall, it's hard to make a good review. I've done my best. Great Code 🥳

AirCasting/Graph/GraphView.swift
```diff
@@ -10,30 +10,33 @@ import SwiftUI
 struct GraphView: View {
     
     let session: SessionEntity
-    let thresholds: [SensorThreshold]
+    let threshold: SensorThreshold
+    @Binding var selectedStream: MeasurementStreamEntity?
```
pawelgil, line 14: binding UI against `NSManagedObject` subclasses is suboptimal, but I think we'll be addressing this holistically across the app in near future so we can leave it for now

AirCasting/Map/AirMapView.swift
```diff
@@ -11,44 +11,40 @@ import Foundation
...
     private var pathPoints: [PathPoint] {
-        measurementStream?.allMeasurements?.compactMap {
-            if let location = $0.location {
-                return PathPoint(location: location, measurement: $0.value)
-            } else {
-                #warning("TODO: Do something with no location points")
-                return nil
-            }
-       } ?? []
+        if let measurementStream = selectedStream {
```
pawelgil, line 19: Since you're making `?? []` anyway in the next statement you can just omit this `if let` and add one more item to optional chaining like this (also note the use of `guard let` to get rid of unneeded indentation): 
``` Swift
return selectedStream?.allMeasurements?.compactMap {
     #warning("...")
     guard let location = $0.location else { return nil }
     return PathPoint(location: location, measurement: $0.value)
} ?? []
```

AirCasting/SessionViews/MeasurementsView.swift
```diff
@@ -0,0 +1,104 @@
+// Created by Lunar on 07/06/2021.
+//
+
+import SwiftUI
+
+struct ABMeasurementsView: View {
+
```
pawelgil, line 7: I think we don't do empty lines after declarations in majority of the codebase

AirCasting/SessionViews/MeasurementsView.swift
```diff
@@ -0,0 +1,104 @@
...
+                          session.pm2Stream,
+                          session.pm10Stream,
+                          session.FStream,
+                          session.HStream]
+        
+        let toShow = allStreams.compactMap { $0 }
+        return toShow
+    }
+    private var hasAnyMeasurements: Bool {
+        streamsToShow.filter { $0.latestValue != nil }.count > 0
```
pawelgil, line 23: I think [contains(where:)](https://developer.apple.com/documentation/swift/array/2297359-contains) would be a better fit here

AirCasting/SessionViews/MeasurementsView.swift
pawelgil, line 27: inverting and using `guard hasAnyMeasurements` would make this code a bit more friendly

AirCasting/SessionViews/MeasurementsView.swift
```diff
@@ -0,0 +1,104 @@
...
+                    .font(Font.moderate(size: 12))
+            }
+            .foregroundColor(.darkBlue)
+        }
+    }
+
+}
+
+struct SingleMeasurementView: View {
+
```
pawelgil, line 56: space again (we should have a vote on this I think - do we make newlines after declarations or no?)

AirCasting/SessionViews/SessionCellView.swift
```diff
@@ -71,11 +97,22 @@ private extension SessionCellView {
         }
         .buttonStyle(GrayButtonStyle())
     }
+    
+    func thresholdFor(selectedStream: MeasurementStreamEntity?) -> SensorThreshold? {
+        thresholds.first { threshold in
+            let streamName = selectedStream?.sensorName?
+                .drop { $0 != "-" }
+                .replacingOccurrences(of: "-", with: "")
+                .lowercased()
```
pawelgil, line 106: drop and replace is the same as in `MeasurementsView.swift`. Maybe it should be moved to an extension?

AirCasting/SessionViews/SessionHeaderView.swift
```diff
@@ -74,46 +77,15 @@ private extension SessionHeaderView {
...
-    func thresholdFor(name: String) -> SensorThreshold? {
-        thresholds.first { $0.sensorName == name }
-    }
-    
     var measurementsMic: some View {
         VStack(alignment: .leading, spacing: 5) {
-            measurementsTitle
-            singleMeasurement(streamName: "db", value: lastMicMeasurement())
+            Text("Most recent measurement:")
+            if let dbStream = session.dbStream {
```
pawelgil, line 83: I'm no expert in SwiftUI, but will this work `guard let dbStream = session.dbStream else { return }`? If yes we can get rid of one level of indentation here.

pawelgil (approved):
No errors found, only stylistic comments

AirCasting/SessionViews/SessionHeaderView.swift
smialko: Unfortunately, it won't work

smialko: Unfortunately, no. Guard cannot be used inside SwiftUI view builder.

AirCasting/SessionViews/MeasurementsView.swift
smialko: Guard cannot be used inside SwiftUI view builder.

AirCasting/Utils/ButtonStyles/AirBorderedButtonStyle.swift
smialko: I couldn't use` BorderedButtonStyle()` because it already existed. I created a custom button style, and like in case of SectionPicker I added "air" at the beginning which stands for AirCasting.

AirCasting/SessionViews/SessionHeaderView.swift
pawelgil: Well, my first steps in swiftui will be painful for sure 😬

AirCasting/SessionViews/MeasurementsView.swift
smialko: When doing as you suggested I have to move the body to `Group()`, and add a return. Convince me if I'm wrong but I think my initial solution looks better. 

 ```
private var streamsToShow: [MeasurementStreamEntity] {
        let allStreams = [session.pm1Stream,
                          session.pm2Stream,
                          session.pm10Stream,
                          session.FStream,
                          session.HStream]
        
        let toShow = allStreams.compactMap { $0 }
        return toShow
    }
    
    var body: some View {
        let streams = streamsToShow
        let hasAnyMeasurements = streams.filter { $0.latestValue != nil }.count > 0
        
        return Group {
            if hasAnyMeasurements {
                VStack(alignment: .leading, spacing: 5) {
                    Text("Most recefnt measurement:")
                    HStack {
                        Group {
                            ForEach(streams) { stream in
                                                      value: stream.latestValue ?? 0,
                                                      threshold: threshold,
                                                      selectedStream: _selectedStream)
                            }
                        }
                        .frame(maxWidth: .infinity)
                    }
                }
            } else {
                VStack(alignment: .leading, spacing: 3) {
                    Text("Your AirBeam is gathering data.")
                        .font(Font.moderate(size: 14))
                    Text("Measurements will appear in 3 minutes.")
                        .font(Font.moderate(size: 12))
                }
                .foregroundColor(.darkBlue)
            }
        }
    }
}
```

Comment (smialko) [reply AirCasting/SessionViews/MeasurementsView.swift:27]: I'm just not sure if tradeoff is worth it. It's not like calling streamsToShow twice would make any performance issues.

File: AirCasting/SessionViews/MeasurementsView.swift (hunk 2 above)
Comment (smialko) [reply AirCasting/SessionViews/MeasurementsView.swift:91]: Aren't those types imposed by the backend? I'm not sure what can I do about that.

File: AirCasting/SessionViews/MeasurementsView.swift (hunk 12 above)
Comment (smialko) [reply AirCasting/SessionViews/MeasurementsView.swift:56]: TBH I like to have a space after struct/class declaration but if it's more readable not to have it, it's fine with me.

File: AirCasting/SessionViews/SessionCellView.swift (hunk 13 above)
Comment (smialko) [reply AirCasting/SessionViews/SessionCellView.swift:106]: Can't move it to the extension, because in first case I'm doing it on `SensorThreshold` and in the second on `MeasurementStreamEntity`. So I would need to do it twice anyways. 
Do you think it would be a good idea to make a global func for that?

**Review by jankrzempek — COMMENTED** [review]
👌🏻

File: AirCasting/SessionViews/SessionCellView.swift (hunk 13 above)
Comment (pawelgil) [reply AirCasting/SessionViews/SessionCellView.swift:97]: When something like this happens it tells us that we should probably create a new object. Currently we're using entities directly which prevents us from doing so, but in the future we should move all `let/var sensorName: String` declarations from being `String` to a custom type like `SensorName` which would implement this logic. For now we can introduce something super simple like:
``` Swift
class SensorNameFormatter {
     func format(_ name:) -> String {
          ...
     }
}
```

and just use it directly here:
``` Swift
let streamName = SensorNameFormatter().format(sensorName)
```

It's not ideal, but will do 🤷‍♂️

AirCasting/SessionViews/MeasurementsView.swift
mateuszmackowiak: noting Is suggested by the backend. json only supports numeric type and has no limits to the values. Ofcourse as a separate task it makes no sense to separate it but worth considering because it can lead to very strange bugs in some corner cases. but should lt should be done sooner than later to avoid propagating the issue further.

AirCasting/SessionViews/MeasurementsView.swift
mateuszmackowiak: you should calculate the potential recreation every second and for every visible cell etc. But if not convinced maybe @lesniakania @pawelgil you can comment.

AirCasting/Utils/ButtonStyles/AirBorderedButtonStyle.swift
mateuszmackowiak: 🤔 imo not a good approach to prefix with air. Same would be if You added `Bordered2ButtonStyle` etc etc. says nothing. @anna1901 @pawelgil @jankrzempek any thoughts on how could we handle the naming convention here? 

[interesing](https://pspdfkit.com/blog/2018/naming-classes-why-it-matters-how-to-do-it-well/)

smialko: Sure, it would be great to come up with some naming convention rules :)

AirCasting/SessionViews/SessionCellView.swift
smialko: O, great tip, thanks. I wouldn't come up with such solution :)

AirCasting/SessionViews/MeasurementsView.swift
smialko: Ok, I won't argue with you, I'm doing it your way. But tbh I'm not 100% convinced ;)

AirCasting/SessionViews/SessionCellView.swift
smialko: I'm adding it as a fix for the future then, after the refactor.

AirCasting/Utils/ButtonStyles/AirBorderedButtonStyle.swift
mateuszmackowiak: @pawelgil

AirCasting/CoreData/SessionEntity.swift
```diff
@@ -76,6 +76,17 @@ public class SessionEntity: NSManagedObject, Identifiable {
         get { SessionType(rawValue:(value(forKey: "type") as! String)) }
         set { setValue(newValue.rawValue, forKey: "type") }
     }
+    
+    public var allStreams: [MeasurementStreamEntity]? {
+        measurementStreams?.array as? [MeasurementStreamEntity]
+    }
+    
+    func streamWith(sensorName: String) -> MeasurementStreamEntity? {
+        return self.allStreams?.first { stream in
```
pawelgil, line 85: `return` and `self` can be omitted

AirCasting/Utils/ButtonStyles/AirBorderedButtonStyle.swift
pawelgil: tough. `BorderedButtonStyle` is perfect. In my opinion this kind of a naming collision is a clear code smell indicating that we need a namespace for our own stylings. To me it looks like the desired way to achieve namespacing in this case would be by creating a swift module for all in-house button styles (or perhals all styling, not only buttons). It's really lightweight and easy to do. We'd have a module called let's say `AirCastingStyling` the sake of this argument, and than in the codebase:
```Swift
import AirCastingStyling

...
.buttonStyle(AirCastingStyling.BorderedButtonStyle(...))
...
```

If not, we're basically in a position where we need to compromise and, conceptually, I think `Air` prefix is as good as any because the best fitting option `BorderedButtonStyle` is off the table anyway and we need to bandaid.

PR #60, opened 2021-06-18
Added settings page

jankrzempek:
https://trello.com/c/3vxiFRbo/4-settings-for-profile-mvp-view


AirCasting/Settings/MyAccountView.swift
```diff
@@ -0,0 +1,58 @@
...
+//
+
+import SwiftUI
+
+struct MyAccountView: View {
+    var body: some View {
+        ZStack {
+            VStack(alignment: .leading) {
+                
+                Text("You aren’t currently logged in")
```
mateuszmackowiak, line 11: in such cases where the logic isn't finished please add a TODO

mateuszmackowiak: Or this screen has an invalid name.

AirCasting/Settings/MyAccountViewSingOut.swift
```diff
@@ -0,0 +1,40 @@
...
+//
+
+import SwiftUI
+
+struct MyAccountViewSingOut: View {
+    var body: some View {
+        ZStack {
+            VStack(alignment: .leading) {
+                
+                Text("You are currently logged in as jan.krzempek@lunarlogic.io")
```
mateuszmackowiak, line 11: please add a todo.

AirCasting/Settings/SettingsView.swift
```diff
@@ -10,26 +10,95 @@ import Foundation
...
-            
-            Spacer()
+            }
+            .listStyle(GroupedListStyle())
+            .navigationBarTitle("Settings")
+        }
+    }
+    
+    
+   private struct FullScreenModalView: View {
```
mateuszmackowiak, line 75: Please think of a better name

AirCasting/Settings/SettingsView.swift
```diff
@@ -10,23 +10,91 @@ import Foundation
 
 struct SettingsView: View {
     let logoutController: LogoutController
-
+    @State private var isToggle : Bool = false
```
pawelgil, line 13: I know it's nit-picky, but the usual styling (in the app too afaict) is to put a colon after the name without any spaces:

```Swift
@State private var isToggle: Bool = false
```

AirCasting/Settings/MyAccountViewSingOut.swift
```diff
@@ -0,0 +1,40 @@
+// Created by Lunar on 18/06/2021.
+//
+
+import SwiftUI
+
+struct MyAccountViewSingOut: View {
```
pawelgil, line 6: typo `SingOut` -> `SignOut`

AirCasting/Settings/MyAccountView.swift
```diff
@@ -0,0 +1,58 @@
...
+                }) {
+                    Group {
+                        HStack {
+                            Text("Create an account")
+                            Spacer()
+                            Image(systemName: "chevron.right")
+                        }
+                        .padding(.horizontal)
+                    }
+                } .buttonStyle(BlueButtonStyle())
```
pawelgil, line 25: remove space between `}` and `.` plz

AirCasting/Settings/MyAccountView.swift
```diff
@@ -0,0 +1,58 @@
...
+                            .stroke(lineWidth: 1.0)
+                    )
+                }
+                .padding()
+                Spacer()
+            }
+        }
+        .navigationTitle("My account")
+    }
+}
```
pawelgil, line 50: I think this body would be much more readable if we got rid of the unneeded details at the `body` level. [Top-level functions](https://stackoverflow.com/questions/633617/whats-the-difference-between-low-level-functions-top-level-functions) (computed var here is really a function with no paramaters, `var` is just Swift syntactic sugar) should really be screaming their intent, hiding any unneeded implementation detail in some private functions below. Anyone that wants to know the details would just scroll down to the part they are interested in and read it and everyone elses time is saved by making this high-level concept really easy to read thru. An example how this particular code can be divided:

```Swift
var body: some View {
    ZStack {
        VStack(alignment: .leading) {
            #warning("Logic here needs to be finished - handle log in")
            loggedOutInformationLabel
            createAccountButton
            logInButton
            Spacer()
        }
    }
    .navigationTitle("My account")
}

private var loggedOutInformationLabel: some View {
    Text("You aren’t currently logged in")
        .padding()
}

private var createAccountButton: some View {
    Button(action: {
    }) {
        Group {
            HStack {
                Text("Create an account")
                Spacer()
                Image(systemName: "chevron.right")
            }
            .padding(.horizontal)
        }
    }
    .padding()
}

private var logInButton: some View {
    Button(action: {
    }) {
        Group {
            HStack {
                Text("Login in")
                Spacer()
                Image(systemName: "chevron.right")
            }
            .padding(.horizontal)
            .padding(.vertical, 12)
        }
        .overlay(
            RoundedRectangle(cornerRadius: 5)
                .stroke(lineWidth: 1.0)
        )
    }
    .padding()
}
```

This simple little division makes it super-clear how this view looks like at just first glance.

AirCasting/Settings/MyAccountViewSingOut.swift
```diff
@@ -0,0 +1,40 @@
...
+    var body: some View {
```
pawelgil, line 7: here the `body` is less complicated, but IMO would still benefit from making same thing as above (your call)

AirCasting/Settings/SettingsView.swift
```diff
@@ -10,23 +10,91 @@ import Foundation
...
+    @State private var showModal = false
     var body: some View {
```
pawelgil, line 15: this is a huge body function over here, let's try to cut it into smaller pieces like in the example above and to make it more readable this way

pawelgil (changes requested):
Lets make it a bit cleaner 🧹
\ also minor typos

AirCasting/Settings/MyAccountView.swift
```diff
@@ -8,47 +8,58 @@ struct MyAccountView: View {
...
                 Spacer()
             }
         }
-        .navigationTitle("My account")
+        .navigationTitle(Strings.MyAccountSettings.title)
     }
 }
 
+private var loggedOutInformationLabel: some View {
+    Text("You aren’t currently logged in")
```
pawelgil, line 22: missed this one, can go to `Strings` struct also :)

AirCasting/Settings/SettingsView.swift
```diff
@@ -9,92 +9,136 @@ import SwiftUI
...
-                                    .accentColor(.black)
-                                Spacer()
-                                Image(systemName: "control")
-                                    .accentColor(.gray).opacity(0.6)
-                            }
-                        }
-                    }.sheet(isPresented: $showModal, content: {
-                        BackendSettingsModalView()
-                    })
+                    showModalButton
```
pawelgil, line 33: good job on dividing stuff over here. One thing that can be better is this name. "showModalButton" is not really carrying any intent description. 'Modal' is just a presentation style, so showModalButton is saying nothing. Maybe lets think of a better name (from what I see here this button navigates the user to some backend settings, so maybe something like `navigateToBackendSettingsButton` ? )

AirCasting/Settings/SettingsView.swift
```diff
@@ -9,92 +9,136 @@ import SwiftUI
...
                 Spacer()
-                Button("OK") {
-                    presentationMode.wrappedValue.dismiss()
-                }.buttonStyle(BlueButtonStyle())
-                
-                Button("Cancel") {
-                    presentationMode.wrappedValue.dismiss()
-                }.buttonStyle(BlueTextButtonStyle())
+                oKButton
+                cancelButton
```
pawelgil, line 121: nice cleanup! 🧹

pawelgil (changes requested):
great job on applying clean code techniques! Let's rename one more thing and cleanup missed string literal and we'll be good to go here

pawelgil (approved):
🔥

PR #61, opened 2021-06-21
Store user profile in keychain

mateuszmackowiak:


AirCasting/AuthViews/UserAuthenticationSession.swift
```diff
@@ -3,34 +3,59 @@
...
-            isLoggedIn = token != nil
+            isLoggedIn = user != nil
         }
     }
 
     init() {
         do {
-            token = try keychainStorage.string(forKey: Self.authenticationTokenKey)
-            isLoggedIn = token != nil
+            if let data = try keychainStorage.data(forKey: Self.userProfileKey) {
```
pawelgil, line 30: I like `guard let` better here because it would eliminate the need for `else`

AirCasting/AuthViews/UserAuthenticationSession.swift
```diff
@@ -3,34 +3,59 @@
...
     }
 
     func deauthorize() throws {
-        try keychainStorage.removeValue(forKey: Self.authenticationTokenKey)
-        self.token = nil
+        try keychainStorage.removeValue(forKey: Self.userProfileKey)
+        self.user = nil
+    }
+
+    private struct CodableUser: Codable {
```
smialko, line 54: Why do we have separate models for the decodable user and codable user?

smialko (approved):
🤠

AirCasting/AuthViews/UserAuthenticationSession.swift
pawelgil, line 54: Interesting. Myself I would probably just make this tradeoff and declare `User` itself as `Codable`. This solution has a downside of double maintenance. If we decide that we need a `name` in `User` struct we would need to remember to change this code too. And if we assume this is set in stone and will never change than why not making `User` Codable after all. I mean - `Codable` is not really doing anything, the responsibility lies on Encoder/Decoder which is controlled by this class, so IMO it's not violating SRP. I don't know, 50/50

pawelgil (approved):
Looking good 👌 super minor, theoretical comments added, nothing important found.

PR #62, opened 2021-06-21
Sessions synchronization mechanism added

pawelgil:
***Still work in progress, please don't review just yet***
# What was done? 🔧
This PR introduces the session synchronization mechanism. It's needed because a single account can be shared across multiple devices introducing differences between local databases. Because of that the app needs to periodically synchronize with the backend. It works by determining what sessions are missing from the database, what sessions had not been uploaded to the server yet and what sessions were removed by external actors and acting accordingly. This Pseudo-UML chart roughly describes the algorithm:
SyncFlow

# Technical notes 🤓
## Combine Usage 🥷
I decided that the core implementation and required interfaces will use Combine extensively. I'm not an advocate for doing Combine everywhere (note that `Database` object in the `App/Frameworks` layer is closure-based, as usual), but there are some good reasons to use it here:
- Asynchronous nature of the problem - the synchronization scenario is particularly asynchronous. There are multiple async actors present:
    - `Downstream` for session downloads
    - `Upstream` for session uploads
    - `SyncStore` for database read/writes
    - `ContextProvider` for receiving sync context
- Such complexity is really hard to implement using the standard callback approach and leads to a `callback hell`
- Combine lets us avoid that problem and furthermore, make the code more readable. For example the entire function responsible for downloading, pooling and saving to the store looks like this:
```Swift
context.needToBeDownloaded
            .publisher
            .flatMap { self.downloadSingleSession(uuid: $0) }
            .collect()
            .flatMap { self.saveSessions(downloadedSessions: $0) }
            .eraseToAnyPublisher()
```
versus the callback-based code which would probably look like this (imagine how hard it would be to perform canceling here which we get out of the box with Combine):
```Swift
context.needToBeDownloaded.forEach {
            var totalDownloadsFinishedCount: Int = 0
            var pooledDownloads: [SessionsSynchronization.SessionDownloadData] = []
            
            self.downloadSingleSession(uuid: $0) { result in
                totalDownloadsFinishedCount += 1
                if case .success(let session) = result else {
                    Log.error("[SYNC] Error downloading session: \(error)")
                } else {
                    pooledDownloads.append(session)
                }
                // Only save after all downloads finish so we don't cause disco on the UI
                if context.needToBeDownloaded.count == totalDownloadsFinishedCount {
                    self.saveSessions(downloadedSessions: pooledDownloads) { result in
                        if case .success(let session) = result else {
                            Log.error("[SYNC] Error saving downloaded data: \(error)")
                        }
                    }
                }
            }
        }
```
And this is just the inner part of the whole system. In my opinion Combine is *perfect* for such cases.
* Future maintenance and extension will be easier. For example if we wish to introduce a retry for downloads/uploads all we need to do is add `.retry(N)` to the pipeline
## Class structure 🎻
The implementation uses some Clean Architecture patterns to ensure modularity and testability:
- The `Controller` part is the heart of the operation
    - It defines `protocols` for its comrades like data sources to implement in order to play with it
    - It uses dedicated data structures for each of those relationships:
        - Data coming from and to the `Store`
        - Data coming from the `RemoteContextProvider`
        - Data coming from the `SessionDownstream`
        - Data coming from the `SessionUpstream`
- Object in the `InterfaceAdapters` part are the concrete implementations of the `protocols` defined by `Controller`
    - Their job is to solely to adjust external framework-dependent, high-abstraction classes like `Database` (introduced in this PR) or `ClientAPI` to fit what the `Controller` wants from us
    - They are simple, mostly translating framework layer data structures into structures defined by the `Controller`
- Compositional helper is added in order to schedule the work done by the `Controller` into other threads. It uses the Proxy pattern in order to do so. It's beneficial from the unit testing standpoint as we can have a synchronous setup in tests, and in the app simply wrap it into the scheduler proxy to ensure no UI thread is affected.
- Simple, conceptual (names won't match probably) diagram:
Diagram bez tytułu

# Testing 🔥
Testing scenarios would include:
- [ ] Can downloaded items be seen on the UI after the sync is finished? _(I'm not saving each download separately to not clutter the UI, only after the sync is finished the batch update is performed)_
- [ ] Are uploaded sessions visible on other devices after syncing?
- [ ] Did the app remove all sessions marked for delete?
- [ ] Is the app triggering synchronization for all required scenarios?
- [ ] If there is an error during downloads/uploads/context fetch, does the app behave as we want it to?
- [ ] Trying to trigger sync when not all conditions are met (I think being logged in is the only one)
- [ ] Is synchronization canceled after on logout?
- [ ] After relog does it still work properly?
# What still needs to be done ☕️
In upcoming days I'll extend this PR with some more code in order to:
* Trigger sync in all required scenarios
* Encapsulate into a separate module (although this would require to modularize a lot of stuff too - Database, ClientAPI, Logger and many commonly used structures/classes. I think this is out of scope of this PR

Some things to remember about this PR:
* Measurements are *not* being fetched for synchronized sessinon streams (tbd in separate task?)


AirCasting/SessionsSynchronization/Controller/Interfaces/SessionSynchronizationStore.swift
```diff
@@ -0,0 +1,67 @@
...
+protocol SessionSynchronizationStore {
+    func getLocalSessionList() -> AnyPublisher<[SessionsSynchronization.Metadata], Error>
+    
+    @discardableResult
+    func addSessions(with: [SessionsSynchronization.SessionStoreSessionData]) -> Future<Void, Error>
+    
+    @discardableResult
+    func removeSessions(with: [SessionUUID]) -> Future<Void, Error>
+    
+
```
jankrzempek, line 23: two enters, don't know if preference or mistake. Just letting You know 😄

AirCastingTests/API/SyncContextServiceTests.swift
```diff
@@ -0,0 +1,175 @@
...
+}
+
+extension Array where Element == SessionsSynchronization.Metadata {
+    static var random: [Element] = [.random, .random]
+}
+
+class RequestAuthorizationServiceMock: RequestAuthorisationService {
+    struct DummyError: Error {
+        
+    }
```
jankrzempek, line 153: It's empty, don't know if it should be.. wondering

jankrzempek (commented):
I am not able to say anything bad, reading this carefully and learning. Great 🥳

AirCastingTests/API/SyncContextServiceTests.swift
pawelgil: since `Error` is a protocol we need any implementation for it to use in test situations. So therefore the need for `DummyError`. But good you mention this since I also created a slightly less dummy DummyError as a universal test utility:
```Swift
struct DummyError: Error, Equatable {
    let errorData: String
    
    init(errorData: String = .default) {
        self.errorData = errorData
    }
}
```
So this one indeed can be removed!

jankrzempek: I didn't know that earlier, good to know. Thanks!

AirCasting/SessionsSynchronization/Adapters/DataStore/SessionSynchronizationDatabase.swift
```diff
@@ -0,0 +1,152 @@
...
+                                                               version: Int(version),
+                                                               longitude: location?.longitude,
+                                                               latitude: location?.latitude,
+                                                               sessionType: type.rawValue,
+                                                               measurementStreams: measurements ?? [])
+    }
+}
+
+extension SessionsSynchronization.Metadata {
+    init(entity: Database.Session) {
```
mateuszmackowiak, line 149: IMo not a good idea. there should be a mapping layer for it to mapping errors spread around the project

AirCasting/SessionsSynchronization/Adapters/Services/SessionDownloadService.swift
```diff
@@ -0,0 +1,68 @@
...
+        return decoder
+    }()
+    
+    init(client: APIClient, authorization: RequestAuthorisationService, responseValidator: HTTPResponseValidator) {
+        self.client = client
+        self.authorization = authorization
+        self.responseValidator = responseValidator
+    }
+    
+    func download(session: SessionUUID) -> AnyPublisher<SessionsSynchronization.SessionDownstreamData, Error> {
```
mateuszmackowiak, line 27: Are You sure that this is possible and easy to test? Imo completion block here and a wrapper on publisher is more reelable/redable/testable

AirCasting/SessionsSynchronization/Controller/Interfaces/SessionDownstream.swift
```diff
@@ -0,0 +1,55 @@
...
+///
+/// Overview of the sync process:
+/// 1. Fetch sessions stored on a device
+/// 2. Diff them against upstream database
+/// 3. Download/Upload/Delete accordingly (this interface)
+protocol SessionDownstream {
+    func download(session: SessionUUID) -> AnyPublisher<SessionsSynchronization.SessionDownstreamData, Error>
+}
+
+// MARK: Data structures
```
mateuszmackowiak, line 17: Can't find it without it ;)

AirCasting/SessionsSynchronization/Controller/Interfaces/SessionUpstream.swift
```diff
@@ -0,0 +1,53 @@
...
+/// 3. Download/Upload/Delete accordingly (this interface)
+protocol SessionUpstream {
+    func upload(session: SessionsSynchronization.SessionUpstreamData) -> Future<Void, Error>
+}
+
+// MARK: Data structures
+
+extension SessionsSynchronization {
+    struct SessionUpstreamData: Equatable, Encodable {
+        typealias SensorName = String
```
mateuszmackowiak, line 21: How about using this type for other models?

AirCasting/SessionsSynchronization/Controller/SessionSynchronizationController.swift
```diff
@@ -0,0 +1,144 @@
+// Created by Lunar on 10/06/2021.
+//
+
+import Foundation
+import Combine
+import CoreLocation
+
+
+class SessionSynchronizationController: SessionSynchronizer {
```
mateuszmackowiak, line 9: final

AirCasting/SessionsSynchronization/Controller/SessionSynchronizationController.swift
```diff
@@ -0,0 +1,144 @@
...
+            .handleEvents(receiveOutput: {
+                Log.verbose("[SYNC] Context retrieved: \($0)")
+            })
+            .eraseToAnyPublisher()
+    }
+    
+    private func processDownloads(context: SessionsSynchronization.SynchronizationContext) -> AnyPublisher<Void, Error> {
+        context.needToBeDownloaded
+            .publisher
+            .flatMap { self.downloadSingleSession(uuid: $0) }
```
mateuszmackowiak, line 94: flatMap(self.downloadSingleSession(uuid:))

AirCasting/SessionsSynchronization/Controller/SessionSynchronizer.swift
```diff
@@ -0,0 +1,35 @@
...
+    func triggerSynchronization() {
+        triggerSynchronization(completion: nil)
+    }
+}
+
+// MARK: Namespacing struct
+
+/// Namespace for session synchronization data objects
+struct SessionsSynchronization {
+    @available(*, unavailable, message: "This struct is not meant to be instantiated. It only provides a namespace (can remove after modularization)")
```
mateuszmackowiak, line 26: If only for namespace rather than using this hack use caseless enum

AirCasting/SessionsSynchronization/Controller/Utils/BodyEncodingError.swift
```diff
@@ -0,0 +1,9 @@
+// Created by Lunar on 21/06/2021.
+//
+
+import Foundation
+
+/// A common `Error` type for session synchronization service adapters
+enum BodyEncodingError: Error {
+    case dataCannotBeStringified
```
mateuszmackowiak, line 8: `struct BodyEncodingError: Error {}`

mateuszmackowiak: But it seams that with this we are missing the underlying errors that can give us lots of information what has happened exactly.

mateuszmackowiak: Please pass it as a associated value or a param

AirCasting/SessionsSynchronization/Controller/Utils/ScheduledSessionSynchronizerProxy.swift
```diff
@@ -0,0 +1,51 @@
...
+// be the case tho)
+import class UIKit.UIApplication
+import struct UIKit.UIBackgroundTaskIdentifier
+
+/// A  _Proxy_ object wrapping any instance of  a `SessionSynchronizationController` and dispatching its work to a passed `Scheduler`. It also wraps it into a background task.
+///
+/// Note: For more information about the _Proxy_ pattern please read:
+/// - The _Gang of Four_ book
+/// - https://refactoring.guru/design-patterns/proxy
+class ScheduledSessionSynchronizerProxy<S: Scheduler>: SessionSynchronizer {
```
mateuszmackowiak, line 18: final

AirCasting/SessionsSynchronization/Controller/Utils/SessionsSynchronizationStoreDataConverter.swift
```diff
@@ -0,0 +1,63 @@
+// Created by Lunar on 16/06/2021.
+//
+
+import Foundation
+
+/// Provides conversion functions between various sessions sync data structures
+struct SynchronizationDataConterter {
+    @available(*, unavailable, message: "This struct is not meant to be instantiated. It only provides static functions")
+    private init() { }
```
mateuszmackowiak, line 9: caseless enum ?

mateuszmackowiak: But imo this should be a standard class with regular methods. This is logic is not a universal defined, never changing logic and I see lots off potential places where it could be parametrised.

AirCasting/SessionsSynchronization/Database/DatabaseMeasurement.swift
```diff
@@ -0,0 +1,50 @@
...
+        public let measurementShortType: String
+        public let unitSymbol: String
+        public let thresholdVeryLow: Int
+        public let thresholdLow: Int
+        public let thresholdMedium: Int
+        public let thresholdHigh: Int
+        public let thresholdVeryHigh: Int
+        public let isDeleted: Bool
+        
+        #warning("Make this private as soon as we move PersistenceController into here")
```
mateuszmackowiak, line 19: Don't quite get this comment. Acutely at all

AirCasting/SessionsSynchronization/Database/DatabaseMeasurement.swift
```diff
@@ -0,0 +1,50 @@
+import Foundation
+
+extension Database {
+    public struct Measurement {
+        public let id: Int
```
mateuszmackowiak, line 5: typealias for ID type

AirCasting/Utils/ArrayUtils.swift
```diff
@@ -0,0 +1,33 @@
+// Created by Lunar on 14/06/2021.
+//
+
+import Foundation
+
+extension Array {
+    static var empty: Self { [] }
```
mateuszmackowiak, line 7: no!

AirCasting/Utils/ArrayUtils.swift
```diff
@@ -0,0 +1,33 @@
...
+extension Array {
+    init(creating: @autoclosure () -> Element, times: Int) {
+        self = (0..<times).map { _ in creating() }
+    }
+}
+
+infix operator ~~
+
+extension Array where Element: Equatable {
+    static func ~~(lhs: Self, rhs: Self) -> Bool {
```
mateuszmackowiak, line 19: NO!

AirCasting/Utils/CLLocationUtils.swift
```diff
@@ -0,0 +1,35 @@
+// Created by Lunar on 14/06/2021.
+//
+
+import CoreLocation
+
+extension CLLocationCoordinate2D: Equatable { }
```
mateuszmackowiak, line 6: Maybe also make it `Hashable`

AirCasting/Utils/CLLocationUtils.swift
```diff
@@ -0,0 +1,35 @@
...
+
+extension CLLocationCoordinate2D {
+    public init?(latitude: Double?, longitude: Double?) {
```
mateuszmackowiak, line 9: I would be glad You wouldn't.. The check in places where coordinates is mapped to CLLocationCoordinate2D it is lots more understandable that it can fail.

AirCasting/Utils/CLLocationUtils.swift
```diff
@@ -0,0 +1,35 @@
...
+        guard let lat = latitude, let long = longitude else { return nil }
+        self.init(latitude: .init(lat), longitude: .init(long))
+    }
+}
+
+public func ==(lhs: CLLocationCoordinate2D, rhs: CLLocationCoordinate2D) -> Bool {
+    return lhs.latitude == rhs.latitude && lhs.longitude == rhs.longitude
```
mateuszmackowiak, line 16: iMO this is a bad assumption. The accuracy of double is lots and lots and lots wider than the coordinates got from the gps. Comparing this values would probably fail most of the times in some places just because of conversion to double. Assume some rounding.

AirCasting/Utils/CLLocationUtils.swift
```diff
@@ -0,0 +1,35 @@
...
+}
+
+extension Optional where Wrapped == Double {
+    var locationDegrees: CLLocationDegrees {
```
mateuszmackowiak, line 20: just use `double.flatMap(CLLocationDegrees.init)` rather than this. Not every double represents coordinate system and most of the times this would lead to confusion. I would also prefer to have the core location to be used in models and logic from the start so this conversion shouldn't be used almost at all.

AirCasting/Utils/CLLocationUtils.swift
```diff
@@ -0,0 +1,35 @@
...
+}
+
+extension CLLocationCoordinate2D {
+    //
+    // https://en.wikipedia.org/wiki/Null_Island
+    //
+    // since 0 is exactly representable as an IEEE754 floating-point number
+    // comparison with 0 should be safe here:
+    var isLocationLess: Bool {
+        return latitude == 0.0 && longitude == 0.0
```
mateuszmackowiak, line 33: This is not location less. 0 0 is a place in the word. Its in the middle of the sea east to Africa. 
Also please consider checking for CLLocationInvalid. coordinate system only accepts values : lat -90...90 and long -180...180. But in some cases the Aircasting logic may return a location 200,200 which is an invalid value and should be used only as fallback and mapped to nil location in our system.

AirCasting/Utils/Publishers/LogErrorAndCompletePublisher.swift
```diff
@@ -0,0 +1,41 @@
...
+    ///   - message: A message that will prefix the `error.localizedDescription` printout.
+    ///   - logFunc: Optional custom logging function.
+    /// - Returns: A publisher that matches the upstream `Output` and has a `Failure` set to `Never`.
+    func logErrorAndComplete(message: String = "", logFunc: @escaping (String) -> Void = { Log.error($0) }) -> Publishers.LogErrorAndComplete<Self> {
+        .init(upstream: self, logMessage: message, logFunc: logFunc)
+    }
+}
+
+extension Publishers {
+    struct LogErrorAndComplete<Upstream: Publisher>: Publisher {
```
mateuszmackowiak, line 18: IMo if there is a need for sturct like that that means that something is wrong. Logs should be more contextual for what has tiggered it. log itself uses #file to get the current file to give more context to the logs and this will be useless in this file

AirCasting/Utils/Publishers/LogErrorPublisher.swift
```diff
@@ -0,0 +1,40 @@
...
+        private let logMessage: String
+        private let logFunc: (String) -> Void
+        
+        init(upstream: Upstream, logMessage: String, logFunc: @escaping (String) -> Void) {
+            self.upstream = upstream
+            self.logMessage = logMessage
+            self.logFunc = logFunc
+        }
+        
+        func receive<S: Subscriber>(subscriber: S) where S.Failure == Upstream.Failure, S.Input == Upstream.Output {
```
mateuszmackowiak, line 32: As above.

AirCasting/Utils/Publishers/LogEvent.swift
```diff
@@ -0,0 +1,68 @@
...
+    ///
+    /// Note: This function is used when one doesn't need an output value to perform logging, for example:
+    /// ~~~
+    /// urlSession
+    ///    .dataTaskPublisher(for: url)
+    ///    .logVerbose("Data received!")
+    ///    .sink()
+    ///~~~
+    func logVerbose(message: StringLiteralType, _ logFunc: @escaping (_ stringToLog: String) -> Void = { Log.verbose($0) }) -> Publishers.LogEvent<Self> {
+        .init(upstream: self, logMessageClosure: { _ in message }, logFunc: logFunc)
```
mateuszmackowiak, line 22: As above

AirCasting/Utils/Publishers/LogEvent.swift
```diff
@@ -0,0 +1,68 @@
...
+    ///
+    /// Note: This function is used when one doesn't need an output value to perform logging, for example:
+    /// ~~~
+    /// urlSession
+    ///    .dataTaskPublisher(for: url)
+    ///    .decode(type: [String].self, decoder: JSONDecoder())
+    ///    .logVerbose("Received \($0.count) values!")
+    ///    .sink()
+    ///~~~
+    func logVerbose(_ messageClosure: @escaping (_ output: Output) -> String, _ logFunc: @escaping (_ stringToLog: String) -> Void = { Log.verbose($0) }) -> Publishers.LogEvent<Self> {
```
mateuszmackowiak, line 41: As above

AirCasting/Utils/Publishers/LogEvent.swift
```diff
@@ -0,0 +1,68 @@
...
+        .init(upstream: self, logMessageClosure: messageClosure, logFunc: logFunc)
+    }
+}
+
+extension Publishers {
+    struct LogEvent<Upstream: Publisher>: Publisher {
```
mateuszmackowiak, line 47: As above

AirCastingTests/API/SyncContextServiceTests.swift
```diff
@@ -0,0 +1,171 @@
+// Created by Lunar on 19/06/2021.
+//
+
+import XCTest
+import Combine
+@testable import AirCasting
+
+class SyncContextServiceTests: XCTestCase {
```
mateuszmackowiak, line 8: final

AirCastingTests/API/SyncContextServiceTests.swift
```diff
@@ -0,0 +1,171 @@
...
+            "download": ["UUID-TO-DOWNLOAD1", "UUID-TO-DOWNLOAD2"],
+            "upload": ["UUID-TO-UPLOAD1", "UUID-TO-UPLOAD2"],
+            "deleted": ["UUID-TO-DELETE1", "UUID-TO-DELETE2"],
+        }
+        """.data(using: .utf8)!
+    let client = APIClientMock()
+    let auth = RequestAuthorizationServiceMock()
+    let responseValidator = HTTPResponseValidatorMock()
+    lazy var service = SessionSynchronizationService(client: client, authorization: auth, responseValidator: responseValidator)
+    private var cancellables: [AnyCancellable] = []
```
mateuszmackowiak, line 21: Make private things private. 
Also XCtest allocate a new instance for each test method and delicate all when all test pass. So please use teardown to remove unnecessary objects. I use a lazy var with declaring the type to be force unwrapped optional so teardown only sets the values to nil and that's it. Also not clearing this values can lead to false positives.

AirCastingTests/API/SyncContextServiceTests.swift
```diff
@@ -0,0 +1,171 @@
...
+                XCTAssertEqual(self.client.callHistory.count, 1)
+                let request = self.client.callHistory.first!
+                XCTAssertEqual(request.url?.absoluteString, "http://aircasting.org/api/user/sessions/sync_with_versioning.json")
+                XCTAssertEqual(request.httpMethod, "POST")
+                XCTAssertEqual(request.allHTTPHeaderFields?["Accept"], "application/json")
+                XCTAssertEqual(request.allHTTPHeaderFields?["Content-Type"], "application/json")
+                exp.fulfill()
+            })
+            .store(in: &cancellables)
+        wait(for: [exp], timeout: 0.2)
```
mateuszmackowiak, line 38: This is probably isn't the most readable code and the entry could be hard for less experienced. maybe we can for with completions for the api level ?

mateuszmackowiak: Also please be advise that this is changing some of conventions that we have accepted. Maybe a good idea it to confront it with others

AirCastingTests/API/SyncDownstreamServiceTests.swift
```diff
@@ -0,0 +1,197 @@
...
+import XCTest
+import Combine
+@testable import AirCasting
+
+class SyncDownstreamServiceTests: XCTestCase {
+    let client = APIClientMock()
+    let auth = RequestAuthorizationServiceMock()
+    let responseValidator = HTTPResponseValidatorMock()
+    lazy var service = SessionDownloadService(client: client, authorization: auth, responseValidator: responseValidator)
+    private var cancellables: [AnyCancellable] = []
```
mateuszmackowiak, line 13: make private private.. and as above

AirCastingTests/Logic/SessionsSynchronization/SynchronizationTestDoubles.swift
```diff
@@ -0,0 +1,238 @@
...
+        recordedHistory.append(.readSession(uuid))
+        return .init {
+            $0(self.readErrorToReturn == nil ? .success(self.readSessionToReturn ?? .mock(uuid: uuid.rawValue)) : .failure(self.readErrorToReturn!))
+        }
+    }
+}
+
+// MARK: Mock data structures
+
+extension SessionsSynchronization.SynchronizationContext {
```
mateuszmackowiak, line 104: Move mocks to sepperate file

mateuszmackowiak: If there is a need for a mark that means its a different file

AirCastingTests/Mocks/APIClientMock.swift
```diff
@@ -7,9 +7,12 @@ import Combine
 @testable import AirCasting
 
 final class APIClientMock: APIClient {
-    var requestTaskStub: ((_ request: URLRequest, _ completion: (Result<(data: Data, response: HTTPURLResponse), Error>, URLRequest) -> Void) -> Void)!
+    var callHistory: [URLRequest] = []
```
mateuszmackowiak, line 10: There is a reason for the stub to exist. also to test if the request was even called. I think its a valid case and should be used inside test case

AirCastingTests/Utils/TestDefaultProviding.swift
```diff
@@ -0,0 +1,40 @@
+// Created by Lunar on 14/06/2021.
+//
+
+import Foundation
+
```
mateuszmackowiak, line 5: LOL

AirCastingTests/Utils/DummyError.swift
```diff
@@ -0,0 +1,13 @@
...
+
+import Foundation
+
+struct DummyError: Error, Equatable, LocalizedError {
+    let errorData: String
+    var errorDescription: String? { errorData }
+    
+    init(errorData: String = .default) {
+        self.errorData = errorData
+    }
```
mateuszmackowiak, line 12: How about using `NSError(domain:"", code:0)` in test rather than defining this case. It would be more contextual in tests where the domain can be specyfic and give more context. Also `UUID().uuidString` would be more reliable to check if a static sting in development faze was not used.

AirCastingTests/Utils/TestDefaultProviding.swift
mateuszmackowiak: Each type has a specifying more random way to generate values. I would prefer You use `UUID().uuidString` for strings, `Double.random()` etc etc.

mateuszmackowiak (dismissed):

AirCastingTests/Utils/TestDefaultProviding.swift
pawelgil: This is an implementation of a `Dummy Value` test-pattern, not a `Generated Value` test pattern. This is ought to be used in places where the test don't care at all what value it gets. I think randomization for this case is suboptimal (can lead to flaky tests) and if needed should be done using another type of static pattern:

```Swift
extension Int: Randomizable {
   static var random: Self { ... } 
}
```

AirCastingTests/Utils/DummyError.swift
pawelgil: This class is used as a stub in testing scenarios that require stubbed dependencies to return an error and where the concrete error type, domain etc. don't matter at all. I think I made a mistake of being lazy and this `DummyError` should actually be completely empty implementation and for the tests where `localizedDescription` actually matter I shoul've introduced a `LocalizedErrorStub`. Will do so.

AirCastingTests/Mocks/APIClientMock.swift
pawelgil: The stub was not removed, it's still there :) github made this line red because I moved it one line down (and added the `@escaping` keyword) 🤷‍♂️

AirCastingTests/API/SyncContextServiceTests.swift
pawelgil: > Also please be advise that this is changing some of conventions that we have accepted. Maybe a good idea it to confront it with others

would be easier for me if you mentioned what are the exact conventions

pawelgil: > This is probably isn't the most readable code and the entry could be hard for less experienced. maybe we can for with completions for the api level ?

I'll create a generic helper for spying on observables to make those tests clean. I think it's not worth cluttering production code in order to make tests more readable, but luckily we can have both here.

AirCastingTests/API/SyncContextServiceTests.swift
pawelgil: I don't think in this case we need any cleanup here. I'll cancel all cancellables in a `tearDown` just to be sure, but honestly think it's redundant

AirCasting/Utils/Publishers/LogErrorAndCompletePublisher.swift
pawelgil: `#file` and `#line` can be passed here without a problem, didn't think of that TBH, will get fixed 👌. Those operators are by all means not **needed**, but I think are super handy and encourage logging in parts of the code that use `Combine`. In order to log errors now instead of this:
```Swift
publisher.
    .filter(someFilter)
    .map(someMap)
    .handleEvents(recieveCompletion: { result in
        if case .failure(let error) = result else { return }
        logger.error("Error in the pipeline! \(error.localizedDescription)")
    })
```

which is not-so-pretty,
you can do:
```Swift
publisher.
    .filter(someFilter)
    .map(someMap)
    .logError("Error in the pipeline!")
```

AirCasting/Utils/CLLocationUtils.swift
pawelgil: I was not sure what locationLess means in API calls, I think I got it wrong, will remove 👌

AirCasting/Utils/CLLocationUtils.swift
pawelgil: this was actually unused code, removed 👌

AirCasting/Utils/CLLocationUtils.swift
pawelgil: this was a leftover after some experimentation, nice find, removed 👌

AirCasting/Utils/CLLocationUtils.swift
pawelgil: Not too convinced since lat/long can be `null` in our API responses it's quite handy to have a failable initializer for that:

```Swift
DatabaseSession(
...
location: CLLocationCoordinate2D(latitude: $0.latitude, longitude: $0.longitude),
...
)
```

But I'm not too attached to that so let's change 👌

AirCasting/Utils/CLLocationUtils.swift
pawelgil: all that is removed now

AirCasting/Utils/ArrayUtils.swift
pawelgil: Well, I think that's quite handy in the unit test target and a "no" would not be sufficient to convince me, but I have a better idea: instead of using this let's create a custom assertion for that. I'll be happy and you'll be less mad 🐻

```Swift
assertContainsSameElements(removedSessions, sessionsToDelete)
```

AirCasting/SessionsSynchronization/Database/DatabaseMeasurement.swift
pawelgil: 👌
Note: currently there are some universal, app-wide structures that live inside the `CoreData` group like `SessionUUID`, `MeasurementID` etc.
I think this needs to be moved to some other place, because it's not only CoreData layer that uses it, it's all over the app (very versatile objects that are used by all app layers)

AirCasting/SessionsSynchronization/Database/DatabaseMeasurement.swift
pawelgil: Good catch, it's a leftover nonsense, removed

AirCasting/SessionsSynchronization/Controller/Interfaces/SessionUpstream.swift
pawelgil: Let's do it

AirCasting/SessionsSynchronization/Adapters/Services/SessionDownloadService.swift
pawelgil: I'm quite sure since it indeed is tested. Completion block would mess the code inside `SessionSynchronizationController` pretty bad too, because basically the purpose of this class is to adjust existing API to the needs of `SessionSynchronizationController` which has Combine all over. We could introduce a new class that would communicate with the API and have completion-block interface and here only adjust for Combine, but I think splitting it into a private method here will have the same effect (will be readable on the same level) and won't clutter the project. I'll also change the Upstream implementation too since it uses the same pattern

AirCasting/SessionsSynchronization/Adapters/DataStore/SessionSynchronizationDatabase.swift
pawelgil: You're right, I'll make a new class responsible for mapping between `Database <-> SessionSyncStore` objects

AirCasting/SessionsSynchronization/Adapters/DataStore/SessionSynchronizationDatabase.swift
```diff
@@ -0,0 +1,152 @@
...
+                                                                             measurementType: stream.measurementType!,
+                                                                             sensorName: stream.sensorName!,
+                                                                             sensorPackageName: stream.sensorPackageName!,
+                                                                             thresholdHigh: Int(stream.thresholdHigh),
+                                                                             thresholdLow: Int(stream.thresholdLow),
+                                                                             thresholdMedium: Int(stream.thresholdMedium),
+                                                                             thresholdVeryHigh: Int(stream.thresholdVeryHigh),
+                                                                             thresholdVeryLow: Int(stream.thresholdVeryLow),
+                                                                             unitName: stream.unitName!,
+                                                                             unitSymbol: stream.unitSymbol!)
```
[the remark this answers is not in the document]
pawelgil: The new class for `Database <-> SessionSyncStore` translation will handle this. But it's quite annoying to deal with CoreData optionals, we should make a list of stuff that is **guaranteed** or smth and make it so that the `Database` layer structs like `Database.MeasurementsStream` have all optionality sorted out correctly. This way the rest of the app can breathe a bit and not worry about this issue at all

AirCasting/SessionsSynchronization/Controller/Interfaces/SessionDownstream.swift
pawelgil: tenor

AirCasting/SessionsSynchronization/Adapters/DataStore/SessionSynchronizationDatabase.swift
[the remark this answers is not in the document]
pawelgil: Well it fitted quite nicely into the existing `SynchronizationDataConverter`. But those bangs are something to talk about, I don't want to pass optionals everywhere and the only reason we have them is `CoreData` which is a technical detail, we should make some rules and not let this overwhelm us with optionals.

AirCasting/SessionsSynchronization/Adapters/DataStore/SessionSynchronizationDatabase.swift
pawelgil: Edit: no new object needed, `SynchronizationDataConverter` will do.

AirCasting/SessionsSynchronization/Adapters/DataStore/SessionSynchronizationDatabase.swift
```diff
@@ -0,0 +1,152 @@
...
+                                                   sensorPackageName: $0.sensorPackageName,
+                                                   measurementType: $0.measurementType,
+                                                   measurementShortType: $0.measurementShortType,
+                                                   unitName: $0.unitName,
+                                                   unitSymbol: $0.unitSymbol,
+                                                   thresholdVeryHigh: Int32($0.thresholdVeryHigh),
+                                                   thresholdHigh: Int32($0.thresholdHigh),
+                                                   thresholdMedium: Int32($0.thresholdMedium),
+                                                   thresholdLow: Int32($0.thresholdLow),
+                                                   thresholdVeryLow: Int32($0.thresholdVeryLow))
```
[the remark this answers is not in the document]
pawelgil: I think we want to just use `Int` in the rest of the app. The only place that those conversions must happen is the `App<->CoreData` boundary. So now that we have the `Database` layer we want those to only happen while translating `Database.Session <-> CoreData.Session`. Which is not here. It's a mistake that `Database.Sessions` here have fixed-width integers, I will change this. But overall we cannot help converting at some point. Thankfully Swift is pretty smart doing those and uses actual numeric values to transform one into the other so the only threat here is an overflow (which crashes immediately)

AirCasting/SessionsSynchronization/Adapters/DataStore/SessionSynchronizationDatabase.swift
```diff
@@ -0,0 +1,152 @@
...
+                                            startTime: $0.startTime,
+                                            contribute: $0.contribute,
+                                            deviceId: nil,
+                                            endTime: $0.endTime,
+                                            followedAt: nil,
+                                            gotDeleted: $0.gotDeleted,
+                                            isIndoor: $0.isIndoor,
+                                            tags: $0.tags,
+                                            urlLocation: $0.urlLocation,
+                                            version: Int16($0.version!), // TODO: Are we safe?
```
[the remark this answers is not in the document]
pawelgil: It's been moved from here to `Database <-> CoreData` boundary

AirCasting/SessionsSynchronization/Adapters/DataStore/SessionSynchronizationDatabase.swift
```diff
@@ -0,0 +1,152 @@
...
+import CoreLocation
+
+final class SessionSynchronizationDatabase: SessionSynchronizationStore {
+    typealias DatabaseType = SessionsFetchable & SessionRemovable & SessionInsertable
+    private enum SessionSynchronizationDatabaseError: Error, LocalizedError {
+        case sessionNotFound
+        
+        var errorDescription: String? {
+            switch self {
+            case .sessionNotFound: return "Requested session was not found in database"
```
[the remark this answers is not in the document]
pawelgil: I think there is no point in doing error localization when we have 0 localized strings in production code

AirCasting/RootAppView.swift
```diff
@@ -8,7 +8,8 @@
 import SwiftUI
 
 struct RootAppView: View {
-    @ObservedObject var userAuthenticationSession = UserAuthenticationSession()
+    @EnvironmentObject var userAuthenticationSession: UserAuthenticationSession
+    let sessionSynchronizer: SessionSynchronizer
 
     let persistenceController = PersistenceController.shared
```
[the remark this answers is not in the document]
pawelgil: I didn't change PersistenceController to being non-singleton, this PR is already too big (noted in my notebook to do this, will put separate PR)

AirCasting/CoreData/Session.swift
```diff
@@ -69,7 +73,7 @@ public struct SessionUUID: Codable, RawRepresentable, Hashable, CustomStringConv
     }
 }
 
-public enum SessionType: RawRepresentable, CustomStringConvertible, Hashable, Codable {
+public enum SessionType: Equatable, RawRepresentable, CustomStringConvertible, Hashable, Codable {
```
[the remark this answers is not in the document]
pawelgil: Hashable already there, no need for Equatable 🤷‍♂️

AirCasting/CoreData/Session.swift
```diff
@@ -46,12 +46,16 @@ public struct Session {
...
 
-public struct SessionUUID: Codable, RawRepresentable, Hashable, CustomStringConvertible {
+public struct SessionUUID: Equatable, Codable, RawRepresentable, Hashable, CustomStringConvertible, ExpressibleByStringLiteral {
     public let rawValue: String
 
     public init() {
         rawValue = UUID().uuidString
     }
+    
+    public init(stringLiteral value: StringLiteralType) {
```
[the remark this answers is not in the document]
pawelgil: I don't get the problem

AirCasting/CoreData/PersistenceController+Database.swift
```diff
@@ -0,0 +1,141 @@
...
+                  thresholdVeryHigh: coreDataEntity.thresholdVeryHigh,
+                  thresholdHigh: coreDataEntity.thresholdHigh,
+                  thresholdMedium: coreDataEntity.thresholdMedium,
+                  thresholdLow: coreDataEntity.thresholdLow,
+                  thresholdVeryLow: coreDataEntity.thresholdVeryLow)
+    }
+}
+
+extension URL {
+    init?(string: String?) {
```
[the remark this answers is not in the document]
pawelgil: The type is preserved perfectly here, it will return `URL?` just like the `init?(string: String)` returns `URL?`. The only difference here is that this init here is also allowing us to pass optional strings in. I think it's pretty handy and doesn't violate any Swift ideas. But this extension was unused anyway 🤣 so I'll remove

AirCasting/AirCastingApp.swift
```diff
@@ -7,18 +7,92 @@
...
+            case .inactive:
+                break
+            @unknown default:
+                fatalError()
+            }
         }
     }
 }
+
+class SynchronizationScheduler {
```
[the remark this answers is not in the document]
pawelgil: btw. I think this class is garbage, but let's stick to it until all scenarios that triggers re-sync will be finished so we can implement it properly

jankrzempek (approved):

PR #63, opened 2021-06-25
Backend settings port

jankrzempek:
https://trello.com/c/8r6icqJH/122-settings-logout-functionality-back

https://trello.com/c/kAzjpy2N/117-settings-backend-url

It was decided into two separated tickets, to make it more readable. Although the one of the tasks was small enough I think to combine it.


AirCasting/Settings/BackendURLValidator.swift
```diff
@@ -0,0 +1,70 @@
...
+    private func getComponents(url: String) throws -> URLComponents {
+        guard let components = URLComponents(string: url) else {
+            throw ValidationError.invalidURL
+        }
+        return components
+    }
+    
+    private func validateHost(for components: URLComponents) throws {
+        let host = try getHost(from: components)
+        if host.contains(".") {
```
pawelgil, line 36: great idea with the `separatedBy` usage! It's a bit convoluted tho, will this work ok perhaps?:

```Swift
let hostComponents = getHost(from: components).components(separatedBy: ".")
guard components.count > 1, components.allSatisfy({ $0.count > 0 }) else {
    throw ValidationError.invalidURL
    return
}
```

In essence this code:
1. Checks if we have at least 2 components separated by dots in a host string
2. Checks if all of those components are at last one character

Basically this requires each dot to have something before and something after. 
Note this won't protect against all unwanted URL's (for this we'd need a complex regex), but this is not our goal here I think. Our goal is to not enable the button until user types in some text, a dot, and some more text. So I think this will work fine:

`test.c` -> button enabled
`test.` -> button disabled
`.test` -> button disabled
etc.

AirCasting/APICommunicator/CreateSessionAPI.swift
```diff
@@ -92,15 +93,16 @@ final class CreateSessionAPIService {
...
         self.authorisationService = authorisationService
         self.apiClient = apiClient
         self.responseValidator = responseValidator
+        self.urlProvider = baseUrlProvider
     }
 
     @discardableResult
     func createEmptyFixedWifiSession(input: CreateSessionApi.Input, completion: @escaping (Result<CreateSessionApi.Output, Error>) -> Void) -> Cancellable {
-        let url = URL(string: "http://aircasting.org/api/realtime/sessions.json")!
+        let url = urlProvider.baseAppURL.appendingPathComponent("realtime/sessions.json")
```
pawelgil, line 105: just a thought:

we can apply similar pattern to this with UI-facing strings here, for example
```Swift
struct Endpoints {
    static let createFixedSession = "realtime/sessions.json"
}
```

or even adding a method there:
```Swift
struct Endpoint {
    let path: String
    let method: String
}

struct AppEndpoints {
    ...
    static let createFixedSession = Endpoint(path: "realtime/sessions.json", method: "POST")
    ...
}
```

but TBH I don't really see how this would benefit us 🤣 so maybe let's not

AirCasting/APICommunicator/CreateSessionAPI.swift
```diff
@@ -130,3 +132,33 @@ final class CreateSessionAPIService {
         }
     }
 }
+
+protocol BaseURLProvider {
```
pawelgil, line 136: this should have its own file

AirCasting/APICommunicator/CreateSessionAPI.swift
```diff
@@ -130,3 +132,33 @@ final class CreateSessionAPIService {
...
+    var authorizationURL: URL { get }
+}
+
+extension BaseURLProvider {
+    var authorizationURL: URL {
+        URL(string: "http://aircasting.org/api")!
+    }
+}
+
+class UserDefaultsBaseURLProvider: BaseURLProvider {
```
pawelgil, line 147: this should have its own file

AirCasting/APICommunicator/CreateSessionAPI.swift
```diff
@@ -130,3 +132,33 @@ final class CreateSessionAPIService {
...
+            userDefaults.url(forKey: "baseURL") ?? URL(string: "http://aircasting.org/api")!
+        }
+    }
+    private let userDefaults: UserDefaults
+    init(userDefaults: UserDefaults = .standard) {
+        self.userDefaults = userDefaults
+    }
+}
+
+struct DummyURLProvider: BaseURLProvider {
```
pawelgil, line 162: can we wrap this inside 
```Swift
#if DEBUG
#endif
```
??
And perhaps also add it to some other file (maybe the same as the `protocol`? I feel like this dummy implementation doesn't deserve its own file)

AirCasting/CreateSessionViews/ABConnectedView.swift
```diff
@@ -14,6 +14,7 @@ struct ABConnectedView: View {
     @EnvironmentObject var userAuthenticationSession: UserAuthenticationSession
     @EnvironmentObject var sessionContext: CreateSessionContext
     @Binding var creatingSessionFlowContinues : Bool
+    let baseURL: BaseURLProvider
```
pawelgil, line 17: can this be `private`?

AirCasting/CreateSessionViews/ChooseSessionTypeView.swift
```diff
@@ -16,6 +16,7 @@ struct ChooseSessionTypeView: View {
     @State private var isMobileLinkActive = false
     @State private var didTapFixedSession = false
     @EnvironmentObject var bluetoothManager: BluetoothManager
+    let urlProvider: BaseURLProvider
```
pawelgil, line 19: can this be `private`?

AirCasting/CreateSessionViews/ConnectingABView.swift
```diff
@@ -11,7 +11,8 @@ import CoreBluetooth
 struct ConnectingABView: View {
     
     var bluetoothManager: BluetoothManager
-    var selecedPeripheral: CBPeripheral
+    var selectedPeripheral: CBPeripheral
+    let baseURL: BaseURLProvider
```
pawelgil, line 15: can those be `private`?

AirCasting/CreateSessionViews/PowerABView.swift
```diff
@@ -11,6 +11,7 @@ import AirCastingStyling
 struct PowerABView: View {
     @Binding var creatingSessionFlowContinues : Bool
     @EnvironmentObject private var sessionContext: CreateSessionContext
+    let urlProvider: BaseURLProvider
```
pawelgil, line 14: can this be `private`?

AirCasting/CreateSessionViews/SelectDeviceView.swift
```diff
@@ -21,6 +21,8 @@ struct SelectDeviceView: View {
     
     @Binding var creatingSessionFlowContinues : Bool
     
+    let urlProvider: BaseURLProvider
```
pawelgil, line 24: can this be `private`?

AirCasting/CreateSessionViews/TurnOnBluetoothView.swift
```diff
@@ -16,6 +16,8 @@ struct TurnOnBluetoothView: View {
     
     @Binding var creatingSessionFlowContinues : Bool
     
+    let urlProvider: BaseURLProvider
```
pawelgil, line 19: can this be `private`?

AirCasting/MainTabBarView.swift
```diff
@@ -11,6 +11,7 @@ import Firebase
 
 struct MainTabBarView: View {
     let measurementUpdatingService: MeasurementUpdatingService
+    let urlProvider: BaseURLProvider
```
pawelgil, line 14: can this be `private`? (perhaps this measurementUpdatingService too?)

AirCasting/RootAppView.swift
```diff
@@ -13,21 +13,23 @@ struct RootAppView: View {
     let persistenceController = PersistenceController.shared
     let bluetoothManager = BluetoothManager(mobilePeripheralSessionManager: MobilePeripheralSessionManager(measurementStreamStorage: CoreDataMeasurementStreamStorage(persistenceController: PersistenceController.shared)))
     let microphoneManager = MicrophoneManager(measurementStreamStorage: CoreDataMeasurementStreamStorage(persistenceController: PersistenceController.shared))
+    let urlProvider = UserDefaultsBaseURLProvider()
```
pawelgil, line 16: can this be `private`? (and maybe other props here too?)

AirCasting/Settings/BackendURLValidator.swift
```diff
@@ -0,0 +1,70 @@
+//
+//  BackendURLValidator.swift
+//  URLValidation
+//
+//  Created by Pawel Gil on 25/06/2021.
+//
```
pawelgil, line 6: let's be consistent with those header comments (btw. I hate those and would rather not have them at all ;) ) and put something this here:
```Swift
// Created by Lunar on 25/06/2021.
//
```
AFAIK this is the current meta ;)

AirCasting/Settings/BackendURLValidator.swift
```diff
@@ -0,0 +1,70 @@
...
+
+import Foundation
+
+class BackendURLBuilder {
```
pawelgil, line 10: I think we can make it `final`. Or perhaps even make it a `struct` as this includes behavior only

AirCasting/Settings/BackendURLValidator.swift
```diff
@@ -0,0 +1,70 @@
...
+    }
+    
+    func createURL(url: String, port: String) throws -> URL? {
+        if url.count == 0 && port.count == 0 { return nil }
+        if url.count == 0 && port.count > 0 { throw ValidationError.noURL }
+        let url = appenURLWithSchemeIfNotPresent(url: url)
+        var components = try getComponents(url: url)
+        try validateHost(for: components)
+        components.port = try getPortNumber(from: port)
+        return components.url!
```
pawelgil, line 24: even tho we know at this point this `components.url` really should be non-nil, let's just get rid of the bang here and do:
```Swift
return components.url
```
we won't harm any logic and perhaps avoid some unpredicted crashes here

AirCasting/Settings/BackendURLValidator.swift
```diff
@@ -0,0 +1,70 @@
...
+        if port.count == 0 { return nil }
+        guard let portNumber = Int(port) else {
+            throw ValidationError.invalidPort
+        }
+        return portNumber
+    }
+    
+    private func appenURLWithSchemeIfNotPresent(url: String) -> String {
+        var newUrl = url
+        if !url.hasPrefix("http") {
```
pawelgil, line 65: there is a pitfall here. Any url with prefix `http` will pass. Even this one: `httpSOMETHING.com`. Which is not ideal. I think we should change this `if` statement to this:

```Swift
if !(url.hasPrefix("http://") || url.hasPrefix("https://")) {
...
}
```

This solves the problem (which I believe is a problem indeed, I can imagine hosts which start with `http` like `httpfordummies.com` or whatever...

AirCasting/Settings/MyAccountViewSignOut.swift
```diff
@@ -2,8 +2,11 @@
 //
 
 import SwiftUI
+import AirCastingStyling
 
 struct MyAccountViewSignOut: View {
+    let logoutController: LogoutController
```
pawelgil, line 8: can this be `private` ?

AirCasting/Settings/SettingsView.swift
```diff
@@ -6,11 +6,10 @@
 //
 
 import SwiftUI
-import Foundation
 import AirCastingStyling
 
 struct SettingsView: View {
-    
+    let urlProvider: BaseURLProvider
```
pawelgil, line 12: can this be `private`?

AirCasting/Settings/SettingsView.swift
```diff
@@ -82,71 +74,74 @@ struct SettingsView: View {
...
+}
+
+private struct BackendSettingsModalView: View {
+    @State var urlProvider: BaseURLProvider
+    let backendURLBuilder = BackendURLBuilder()
+    @Environment(\.presentationMode) var presentationMode
+    @State var pathText: String = ""
+    @State var portText: String = ""
+    @State var url: URL?
+    @State var buttonEnabled: Bool = false
```
pawelgil, line 89: can all (or at least some) of those be `private`?

AirCasting/Settings/SettingsView.swift
```diff
@@ -82,71 +74,74 @@ struct SettingsView: View {
                 }
             }
         }.sheet(isPresented: $showModal, content: {
-            BackendSettingsModalView()
+            BackendSettingsModalView(urlProvider: urlProvider)
         })
     }
+}
+
+private struct BackendSettingsModalView: View {
```
pawelgil, line 82: IMO it would be better to have this in a separate file (but not a strong opinion)

AirCasting/Utils/Strings.swift
```diff
@@ -31,7 +31,7 @@ struct Strings {
     
     struct SignOutSettings {
         static let title: String = "My account"
-        static let Logged: String = "You are currently logged in as jan.krzempek@lunarlogic.io"
+        static let Logged: String = "You are currently logged in as "
```
pawelgil, line 34: all other static props here start with lower case, let's keep it this way

AirCastingTests/Settings/URLValidationTests.swift
```diff
@@ -0,0 +1,119 @@
+//
+//  URLValidationTests.swift
+//  URLValidationTests
+//
+//  Created by Pawel Gil on 25/06/2021.
+//
```
pawelgil, line 6: same as other comment about this headers. Let's change to:
```
// Created by Lunar on 25/06/2021.
//
```

AirCastingTests/Settings/URLValidationTests.swift
```diff
@@ -0,0 +1,119 @@
...
+
+import XCTest
+@testable import AirCasting
+
+class URLBuilderTests: XCTestCase {
+    let builder = BackendURLBuilder()
+    static let validURL = "https://superstronka.com/"
+    static let validPort = "23"
```
pawelgil, line 14: can we make all those `private`?

pawelgil (changes requested):

AirCasting/CreateSessionViews/ABConnectedView.swift
jankrzempek: I cannot to this or at least I am not able to because than preview has some problem esp. "'ABConnectedView' initializer is inaccessible due to 'private' protection level"

AirCasting/CreateSessionViews/ChooseSessionTypeView.swift
jankrzempek: Same as above, preview is screaming at me :/

AirCasting/CreateSessionViews/ConnectingABView.swift
jankrzempek: 'ConnectingABView' initializer is inaccessible due to 'private' protection level' this is the comment which I get whenever I try to do it

AirCasting/CreateSessionViews/PowerABView.swift
jankrzempek: same error :/

AirCasting/CreateSessionViews/ABConnectedView.swift
pawelgil: Note I only point those `private`s out, so if it's not possible just don't do it :)

AirCasting/URLProviders/BaseURLProtocol.swift
```diff
@@ -0,0 +1,21 @@
+// Created by Lunar on 28/06/2021.
+//
+
+import Foundation
+
+protocol BaseURLProvider {
```
pawelgil, line 6: why not name the file exactly as the protocol ? (BaseURLProvider.swift)

AirCasting/URLProviders/BaseURLProvider.swift
```diff
@@ -0,0 +1,19 @@
+// Created by Lunar on 28/06/2021.
+//
+
+import Foundation
+
+class UserDefaultsBaseURLProvider: BaseURLProvider {
```
pawelgil, line 6: ... and this file as the class name (UserDefaultsBaseURLProvider.swift)

AirCasting/Settings/ModalSettings.swift
```diff
@@ -0,0 +1,66 @@
+// Created by Lunar on 28/06/2021.
+//
+
+import SwiftUI
+import AirCastingStyling
+
+struct BackendSettingsModalView: View {
```
pawelgil, line 7: It's usually a common practice to name files after classes/structs/protocols they contain. So here it would be `BackendSettingsModalView.swift`. I also think that the `Modal` part in the name is not needed because the view itself does not have a knowledge on how it's being presented

AirCasting/Settings/BackendURLValidator.swift
```diff
@@ -0,0 +1,63 @@
+//
+//  BackendURLValidator.swift
+//  URLValidation
+//
+// Created by Lunar on 25/06/2021.
+//
+
+import Foundation
+
+class BackendURLBuilder {
```
pawelgil, line 10: I missed this one last time :( can you change the name of this file to match the class name?

pawelgil (approved):

PR #69, opened 2021-07-06
Loading sessioon view & differentiate session carts

smialko:
**SUMMARY**
- Added saving the thresholds in AirBeam sessions, 
- When there's no streams or measurements there's a view with info that AB is gathering data, 
- Carts are different for mobile active, dormant, and fixed sessions.


AirCasting/SessionViews/SessionLoadingView.swift
```diff
@@ -0,0 +1,29 @@
+// Created by Lunar on 24/06/2021.
+//
+
+import SwiftUI
+
+struct SessionLoadingView: View {
+    var body: some View {
+        HStack {
+            Image("ABLoading")
+                .frame(width: 60, height: 60)
```
jankrzempek, line 10: Is this working well all the time? Wondering 💭 because .resizable modifier is not here

AirCasting/SessionViews/SessionLoadingView.swift
```diff
@@ -0,0 +1,29 @@
...
+            VStack(alignment: .leading) {
+                Text("Your AirBeam is gathering data.")
+                    .font(Font.moderate(size: 15))
+                Text("Measurements will appear in 3 minutes.")
+            }
+            .font(Font.moderate(size: 13))
+            .lineSpacing(5)
```
jankrzempek, line 17: I would place it into HStack(spacing: 5) to make it one liner. Only my weird point of seeing 😄

AirCasting/SessionViews/StreamsView.swift
```diff
@@ -0,0 +1,67 @@
+// Created by Lunar on 29/06/2021.
+//
+
+import SwiftUI
+
+struct StreamsView: View {
+    
+    @Binding var selectedStream: MeasurementStreamEntity?
+    @ObservedObject var session: SessionEntity
+    var threshold: SensorThreshold
```
jankrzempek, line 10: Can this be private too?

AirCasting/SessionViews/StreamsView.swift
```diff
@@ -0,0 +1,67 @@
...
+    @EnvironmentObject private var microphoneManager: MicrophoneManager
+    let showValues: Bool
```
jankrzempek, line 12: also Can this be private too? 😇

jankrzempek (commented):
BUM 👌🏻👌🏻 (added only some ideas and so on)

AirCasting/SessionViews/SessionCartView.swift
```diff
@@ -56,7 +66,17 @@ struct SessionCellView: View {
     }
 }
 
-private extension SessionCellView {
+private extension SessionCartView {
+    
+    var header: some View {
+        SessionHeaderView(
+            action:  {
```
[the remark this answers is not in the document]
smialko: Nice catch, thanks!

AirCasting/SessionViews/ABMeasurementsView.swift
```diff
@@ -58,25 +61,29 @@ struct SingleMeasurementView: View {
     let value: Double
     var threshold: SensorThreshold
     @Binding var selectedStream: MeasurementStreamEntity?
+    let showValues: Bool
     
     var body: some View {
-        VStack(spacing: 3) {
+        
+        return VStack(spacing: 3) {
```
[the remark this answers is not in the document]
smialko: I'm almost 100% sure I added the return because of some errors, but they don't appear so thanks for finding this :D

AirCasting/SessionViews/SessionLoadingView.swift
smialko: It's not a HStack, it's VStack of two texts. The lineSpacing is for situations where on the smaller phones (like mine) the text has more than one line :)

AirCasting/SessionViews/StreamsView.swift
smialko: Unfortunately nope, because it's then inaccessible in AirMapViewm GraphView and SessionCartView :)

AirCasting/SessionViews/StreamsView.swift
smialko: Unfortunately nope, because StreamsView is then inaccessible in AirMapViewm GraphView and SessionCartView :)

AirCasting/Assets.xcassets/ABLoading.imageset/Contents.json
```diff
@@ -0,0 +1,23 @@
+{
+  "images" : [
+    {
+      "filename" : "Group 2.png",
+      "idiom" : "universal",
+      "scale" : "1x"
```
pawelgil, line 6: I don't think we support 1x scale devices

AirCasting/CoreData/MeasurementStreamStorage.swift
```diff
@@ -76,6 +76,16 @@ final class CoreDataMeasurementStreamStorage: MeasurementStreamStorage {
         newStream.gotDeleted = false
 
         session.addToMeasurementStreams(newStream)
+        
+        let existingThreshold: SensorThreshold? = try context.existingObject(sensorName: stream.sensorName ?? "")
```
pawelgil, line 80: I'm not sure I like those `?? ""` silent fallbacks here. I know those `nil` values are often a PITA, but maybe since this is already a `throwing` function we should guard against it and handle errors?

```Swift
guard let sensorName = stream.sensorName else {
    throw StorageError.malformedStream(context: "Stream does not have a sensor name!")
}
```

Or is it purposeful here? Than maybe to emphasize it:

```Swift
let sensorName = stream.sensorName ?? ""
```

AirCasting/Services/UpdateSessionParams.swift
```diff
@@ -54,7 +54,17 @@ final class UpdateSessionParamsService {
             oldStream.thresholdHigh = streamOutput.threshold_high
             oldStream.thresholdVeryHigh = streamOutput.threshold_very_high
             oldStream.gotDeleted = streamOutput.deleted ?? false
-
+            
+            let existingThreshold: SensorThreshold? = try context.existingObject(sensorName: streamOutput.sensor_name ?? "")
```
pawelgil, line 58: same as before with the `?? ""`

AirCasting/SessionViews/ABMeasurementsView.swift
```diff
@@ -8,6 +8,7 @@ struct ABMeasurementsView: View {
     @ObservedObject var session: SessionEntity
     var threshold: SensorThreshold
     @Binding var selectedStream: MeasurementStreamEntity?
+    let showValues: Bool
```
pawelgil, line 11: `Bool` flags are usually not the best for such purposes. It's clunky and slightly beclouds the call site. Much better and future-proof option (even for such binary situations) would be a dedicated enum:

```Swift
enum MeasurementPresentationStyle {
    case showValues
    case hideValues
}
```

AirCasting/SessionViews/ABMeasurementsView.swift
```diff
@@ -58,25 +61,29 @@ struct SingleMeasurementView: View {
     let value: Double
     var threshold: SensorThreshold
     @Binding var selectedStream: MeasurementStreamEntity?
+    let showValues: Bool
```
pawelgil, line 64: Same as above with the `enum`

AirCasting/SessionViews/SessionCartView.swift
```diff
@@ -11,39 +11,49 @@ import CoreData
...
     
-    let session: SessionEntity
+    @ObservedObject var session: SessionEntity
     let thresholds: [SensorThreshold]
     
+    var shouldShowValues: Bool {
+        isCollapsed && (
+            (session.type == .fixed) ||                                  // Fixed
+                (session.type == .mobile && session.status == .FINISHED) // Dormant
+        ) ? false : true
```
pawelgil, line 26: This is super hard to read and understand, lets use named helper variables to make it readable (compiler would optimize it anyway so it's generally a good idea to do so):

```Swift
let isExpanded = !isCollapsed
let isFixed = (session.type == .fixed)
let isDormant = (session.type == .mobile && session.status == .FINISHED)

return isExpanded || !(isFixed || isDormant)
```

Still not perfect, but at least somehow readable. Maybe it can be perfected further or even extracted to some kind of a `Decider` object?

AirCasting/SessionViews/SessionCartView.swift
```diff
@@ -56,7 +66,17 @@ struct SessionCellView: View {
...
 }
 
-private extension SessionCellView {
+private extension SessionCartView {
+    
+    var header: some View {
+        SessionHeaderView(
+            action: {
+                withAnimation {
+                    isCollapsed = !isCollapsed
```
pawelgil, line 75: you can use [toggle](https://developer.apple.com/documentation/swift/bool/2994863-toggle) here

AirCasting/SessionViews/SessionLoadingView.swift
```diff
@@ -0,0 +1,30 @@
...
+import SwiftUI
+
+struct SessionLoadingView: View {
+    var body: some View {
+        HStack {
+            Image("ABLoading")
+                .resizable()
+                .frame(width: 60, height: 60)
+            VStack(alignment: .leading) {
+                Text("Your AirBeam is gathering data.")
```
pawelgil, line 13: do we have this static `Strings` struct on `development` branch? If yes we should put such string over there (applies to all literal strings, I wont point those in further comments)

AirCasting/SessionViews/StreamsView.swift
```diff
@@ -0,0 +1,67 @@
...
+    
+    var body: some View {
+        if session.deviceType == .MIC {
+            HStack {
+                measurementsMic
+                Spacer()
+                //This is a temporary solution for stopping mic session recording until we implement proper session edition menu
+                if microphoneManager.session?.uuid == session.uuid, microphoneManager.isRecording && (session.status == .RECORDING || session.status == .DISCONNETCED) {
```
pawelgil, line 20: as it's temporary I'm not commenting ;)

AirCasting/SessionViews/StreamsView.swift
```diff
@@ -0,0 +1,67 @@
...
+                                      value: lastMicMeasurement(),
+                                      threshold: threshold,
+                                      selectedStream: .constant(dbStream),
+                                      showValues: showValues)
+            }
+        }
+    }
+    var stopRecordingButton: some View {
+        Button(action: {
+            try! microphoneManager.stopRecording()
```
pawelgil, line 46: I believe this error should be handled by some user-facing popup ("Unable to stop session" or smth?) instead of crashing.

AirCasting/SessionViews/StreamsView.swift
```diff
@@ -0,0 +1,67 @@
...
+        }, label: {
+            Text("Stop recording")
+                .foregroundColor(.blue)
+        })
+    }
+    
+    func lastMicMeasurement() -> Double {
+        return session.dbStream?.latestValue ?? 0
```
pawelgil, line 54: Not sure (really not sure, maybe it's ok, just pointing out) this silent unwrap fail is good here.

pawelgil (changes requested):
Let's change some stuff and fix small issues

jankrzempek (approved):
Nice 😎

AirCasting/Assets.xcassets/ABLoading.imageset/Contents.json
smialko: True, I removed the 1x scale and added a task to Trello to review other images and remove 1x scaled images & add missing scales. Thanks!

AirCasting/CoreData/MeasurementStreamStorage.swift
smialko: Hmm, dunno how it happened but we don't need it at all. It's not optional and when I'm using guard let I get an error that stream.sensorName isn't optional. 
I removed silent fallbacks, too :)

AirCasting/Services/UpdateSessionParams.swift
smialko: I switched to guard let here, thanks.

AirCasting/CoreData/MeasurementStreamStorage.swift
smialko: I know why, it's of a different type `StreamOutput` while the stream below is of a type `MeasurementStream`.

AirCasting/SessionViews/StreamsView.swift
smialko: It's not mine, but it'll be changed in the future :)

AirCasting/SessionViews/StreamsView.swift
smialko: It's a temporary solution for stopping recording, and it's not mine. 
Ania created it to be able to stop recording but this is to be done in the future (we have a task in trello for it) :)

AirCasting/SessionViews/SessionLoadingView.swift
smialko: Changed :)

pawelgil (approved):
Splendid 🎉

PR #70, opened 2021-07-07
Worked around a bug that casued dashboard heading to switch to inline

pawelgil:
## What was done?

Workaround for a bug that caused dashboard heading to switch to `inline`-like after popping from other view.

Before             |  After
:-------------------------:|:-------------------------:
  |


jankrzempek (approved):
👌🏻

smialko (approved):
Nice 🥷🏻

PR #75, opened 2021-07-12
added delete view

jankrzempek:
_What was done_:

Added View to be able to handle then deleting sessions

https://trello.com/c/N2RN5aV7/136-delete-session-view


pawelgil (changes requested):

AirCasting/SessionViews/DeleteViewModal.swift
```diff
@@ -0,0 +1,86 @@
+// Created by Lunar on 09/07/2021.
+//
+
+import SwiftUI
+import AirCastingStyling
+
+struct DeleteViewModal: View {
+    @Binding var deleteModal: Bool
+    @State private var allMeasurements = true
```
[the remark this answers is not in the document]
smialko: And I think the better name would be `allStreams`

smialko (changes requested):
Well, I know it's just a view but I think it's important to change the logic for selecting streams, so it's not hardcoded, and so the streams are passed from the session (db for mic sessions and all 5 for AB sessions). 🍉

AirCasting/Models/StreamsType.swift
```diff
@@ -0,0 +1,11 @@
+// Created by Lunar on 13/07/2021.
+//
+
+import Foundation
+
+final class SessionStreamsType {
```
pawelgil, line 6: is this used?

AirCasting/SessionViews/DeleteSessionViewModel.swift
```diff
@@ -0,0 +1,28 @@
...
+    mutating func toggleSelection() {
+        isSelected.toggle()
+    }
+    mutating func changeSelection(newSelected: Bool) {
+        isSelected = newSelected
+    }
+}
+
+protocol DeleteSessionViewModel: ObservableObject {
+    var isMicrophone: Bool { get }
```
pawelgil, line 21: I don't really fancy this kind of shortcuts. We should aim at making this kind of viewmodels/views agnostic of session types. It's not super easy and not super important here and since we're on a hard deadline let's leave it like this and put a `#warning("Abstract microphone sessions away")` here and perhaps we can schedule a pair programming session in the future so I can show you how to do this

AirCasting/SessionViews/DeleteView.swift
```diff
@@ -0,0 +1,83 @@
...
+        Button {
+            deleteModal.toggle()
+        } label: {
+            Text(Strings.deleteSession.cancelButton)
+        }
+        .buttonStyle(BlueTextButtonStyle())
+    }
+}
+
+extension View {
```
pawelgil, line 70: please move this to a separate file named `View+Print.swift` in some Utils folder

AirCasting/Utils/Strings.swift
```diff
@@ -40,4 +40,11 @@ struct Strings {
         static let description: String = "Select a stream to share"
         static let emailDescription: String = "Or email a CSV file with your session data"
     }
+    
+    struct deleteSession {
```
pawelgil, line 44: UpperCamelCase

Podfile.lock
```diff
@@ -1,83 +1,101 @@
 PODS:
-  - Firebase/Analytics (7.5.0):
+  - Firebase/Analytics (8.3.0):
```
pawelgil, line 2: this updates seem kind of random, we shouldn't update libs randomly without a proper retest

pawelgil (changes requested):

smialko (changes requested):
Didn't look into the code because I found a bug while testing it on the phone. I can't mark any of the streams or all streams because it unmarks but itself after a ~second

AirCasting/SessionViews/DefaultDeleteSessionViewModel.swift
```diff
@@ -0,0 +1,64 @@
+// Created by Lunar on 16/07/2021.
+//
+
+import Foundation
+
+class DefaultDeleteSessionViewModel: DeleteSessionViewModel {
+    
+    func isNotMicrophoneToggle() {
```
smialko, line 8: You don't use it, what it's going to be doing? :D

AirCasting/SessionViews/DefaultDeleteSessionViewModel.swift
```diff
@@ -0,0 +1,64 @@
...
+        isMicrophone = false
+    }
+    
+    func isMicrophoneToggle() {
```
smialko, line 12: Same as above ;)

smialko (approved):
🤔 just wondering about those two variables

AirCasting/SessionViews/DefaultDeleteSessionViewModel.swift
```diff
@@ -0,0 +1,64 @@
...
+    
+    var options: [DeleteSessionOptionViewModel] {
+        willSet {
+            objectWillChange.send()
+        }
+    }
+    
+    init() {
+        #warning("TODO: When implementing please provide real data")
+        options =  !isMicrophone ? [.init(id: 0, title: "All", isSelected: false, isEnabled: false),
```
pawelgil, line 58: now I see that `isMicrophone` is only used here to mock. I thought it's really affecting a UI, not only list of streams. I don't think we need it in that case. Since this task is only a view development we can put static mock data here for now. The way it will work when connecting to real data is that we'll be getting some object here representing session and we'll take streams from there and format them accordingly. We don't really need to know what are they (there is a question on how to display single-stream sessions, but that's out of scope here I think). Or am I wrong?

AirCasting/SessionViews/DeleteSessionViewModel.swift
pawelgil: see my second comment - this might be outdated if `isMicrophone` and other microphone-related are only here for mocking purposes

pawelgil (changes requested):

AirCasting/SessionViews/DefaultDeleteSessionViewModel.swift
jankrzempek: You are right I think, will delete this

AirCasting/SessionViews/DeleteSessionViewModel.swift
```diff
@@ -0,0 +1,27 @@
...
+        isSelected.toggle()
+    }
+
+    mutating func changeSelection(newSelected: Bool) {
+        isSelected = newSelected
+    }
+}
+
+protocol DeleteSessionViewModel: ObservableObject {
+    #warning("Abstract microphone sessions away")
```
pawelgil, line 22: since we removed microphone stuff this warning is no longer needed

pawelgil (changes requested):
let's remove unneeded warning and we're good to go :)

pawelgil (approved):

PR #77, opened 2021-07-12
App version added to settings

jankrzempek:
What was made:

I have added app version and build number at the bottom of the settings app

https://trello.com/c/AxL1DVTV/161-add-app-version-at-the-bottom-of-settings-view


AirCasting/Utils/AppInformations.swift
```diff
@@ -0,0 +1,15 @@
+// Created by Lunar on 12/07/2021.
+//
+
+import Foundation
+import SwiftUI
+
+extension UIApplication {
```
pawelgil, line 7: as this is an extension the convention is to name files in format: `WhatIsBeingExtended+Description.swift`. In this case it might be `UIApplication+AppInfo.swift`

pawelgil (changes requested):

AirCasting/Utils/AppInformations.swift
jankrzempek: Didn't know, thank You! 😄

pawelgil (approved):
🥳

smialko (approved):
🥇

PR #78, opened 2021-07-12
added changes to settings and main Dashboard title

jankrzempek:
What was done:

Title and fonts and colours were changes to be the same as on design.
Additional: I have changed the main "Dashboard" title color to dark Blue/

https://trello.com/c/CNDhTlVP/127-ui-settings-view-has-different-fonts-than-it-supposed-to-have


pawelgil (approved):

smialko (approved):
Not sure if they should be .bold but I believe you checked that with Android :)

jankrzempek:
> Not sure if they should be .bold but I believe you checked that with Android :)

Checked with the designs and they are .bold there :/

PR #79, opened 2021-07-13
changes to shwoing dashboard

jankrzempek:
_What was done:_

When You select fixes or mobile session and then proceed through the whole flow You will be sent back to the matching card:
for exp.
selecting mobile session -> shows mobile active 

https://trello.com/c/kbaxOGqG/155-display-proper-view-after-creating-a-new-session


pawelgil (approved):

AirCasting/MainTabBarView.swift
```diff
@@ -73,6 +76,10 @@ class TabBarSelection: ObservableObject {
     }
 }
 
+class SelectSection: ObservableObject {
+    @Published var selectedSection = SelectedSection.mobileDormant
```
smialko, line 80: Why it's dormant by default? 🤔

smialko (commented):
🤔

AirCasting/MainTabBarView.swift
jankrzempek: my bad, changing to mobile active

PR #83, opened 2021-07-14
Location permission changes

jankrzempek:
What was done:

Changes to Location permission function, now it is used using function

Location permissions appears as on the android app

Alert if location is denied

https://trello.com/c/XFiRwcgv/171-location-denied-handle-the-state


AirCasting/CreateSessionViews/PowerABView.swift
```diff
@@ -23,11 +25,17 @@ struct PowerABView: View {
             }
             continueButton
                 .buttonStyle(BlueButtonStyle())
-        }
+        }.alert(isPresented: $showAlert, content: {
+            Alert(title: Text("Location alert"), message: Text("Please go to settings and allow location first!"), dismissButton: .default(Text("OK")))
```
smialko, line 29: Please, remove the exclamation mark, it feels like we're shouting at users :D

smialko, line 29: "Ok" button should redirect to the settings view (it's similar to when you have Bluetooth off or Bluetooth access denied).

AirCasting/CreateSessionViews/SelectDeviceView.swift
```diff
@@ -32,8 +34,16 @@ struct SelectDeviceView: View {
             Spacer()
             chooseButton
             
-        }
+        }.alert(isPresented: $showAlert, content: {
+            Alert(title: Text("Location alert"), message: Text("Please go to settings and allow location first!"), dismissButton: .default(Text("OK")))
```
smialko, line 38: Same as above :)

AirCasting/Utils/LocationStates.swift
```diff
@@ -0,0 +1,16 @@
+// Created by Lunar on 14/07/2021.
+//
+
+import Foundation
+
+enum LocationSates {
+    case granted
+    case denied
+    
+    var bool: Bool {
```
smialko, line 10: Please rename this variable to sth more descriptive

smialko (changes requested):
🕵🏻‍♂️nice work

AirCasting/Utils/LocationStates.swift
jankrzempek: I think that when I am using it .bool making it clear what I want, I don't have an idea for the better one. Could You please suggest 😇?

pawelgil: if this one is here to stay i must have a descriptive name. Currently bool is pretty clear, but if/when we add another case, for example `case undetermined` than `var bool` says nothing. Maybe let's look at how it's used and adjust the name accordingly and also move this into an `extension` placed in the same file as it's being used:

PowerABView.swift
```Swift
extension LocationStates {
    var continueButtonEnabled: Bool { ... }
}
```

The above seems all right, but this usage is very specific so what I would suggest is to really move this responsibility of determining `enabled` state to where it belongs. So in this case to some `private func` inside `PowerABView`:

```Swift
struct PowerABView: View {
    ...
    private var continueButtonEnabled: Bool { ... } 
    ...
}
```

jankrzempek: Thanks, I will try to change it well 😄

AirCasting/Utils/SystemSettings.swift
```diff
@@ -0,0 +1,23 @@
+// Created by Lunar on 15/07/2021.
+//
+
+import Foundation
+import SwiftUI
+
+func goToLocationAuthSettings() {
```
pawelgil, line 7: let's not use global functions without a *really* good reason. Please create some protocol+class pair for it and inject using environment objects where needed

Podfile.lock
```diff
@@ -1,83 +1,101 @@
 PODS:
-  - Firebase/Analytics (7.5.0):
+  - Firebase/Analytics (8.3.0):
```
pawelgil, line 2: same as in other PR - updates (especially such big ones like 8.x -> 7.x) should not be performed without a good reason or a task for it

pawelgil (changes requested):

smialko (approved):
+1 to Paweł about requested changes. 
But I didn't find anything else that's why I'm giving you the approval, don't merge before fixing things from Paweł ;D

AirCasting/Utils/LocationAlert.swift
```diff
@@ -0,0 +1,15 @@
+// Created by Lunar on 21/07/2021.
+//
+
+import SwiftUI
+
+var locationAlert: Alert {
```
pawelgil, line 6: global var is not good, please change it :)

AirCasting/Utils/LocationStates.swift
```diff
@@ -0,0 +1,9 @@
+// Created by Lunar on 14/07/2021.
+//
+
+import Foundation
+
+enum LocationSates {
```
pawelgil, line 6: typo "Sates" -> "States". Also plural form is not really a thing with enums. `LocationState` would be perfect

AirCasting/Utils/SystemSettings.swift
```diff
@@ -0,0 +1,32 @@
+// Created by Lunar on 15/07/2021.
+//
+
+import Foundation
+import SwiftUI
+
+protocol SettingsRedirectionProtocol {
```
pawelgil, line 7: `Protocol` suffix is redundant here. Otherwise this is exactly what I've been asking for. Now it will be super easy to unit test this beahavior inside viewmodels and/or change implementation to something else if needed (perhaps iOS16 will give us some more options to do so and it will come handy, who knows, always good to have things well separated)

AirCasting/Utils/SystemSettings.swift
```diff
@@ -0,0 +1,32 @@
...
+    func goToLocationAuthSettings()
+    func goToBluetoothAuthSettings()
+}
+
+final class SettingsRedirection: SettingsRedirectionProtocol, ObservableObject {
```
pawelgil, line 12: I think `ObservableObject` is not needed here since there is no state whatsoever

AirCasting/CreateSessionViews/TurnOnBluetoothView.swift
```diff
@@ -13,7 +13,7 @@ struct TurnOnBluetoothView: View {
     
     @State private var isPowerABLinkActive = false
     @EnvironmentObject var bluetoothManager: BluetoothManager
-    
+    @EnvironmentObject var settingsRedirection: SettingsRedirection
```
pawelgil, line 16: It should not use a concrete class here. It should use the protocol you've created for this purpose. When using a concrete class we're loosing all the flexibility that protocols give us

pawelgil (changes requested):

AirCasting/Utils/SystemSettings.swift
jankrzempek: Screenshot 2021-07-22 at 10 36 30
I think it is needed here to use EnviromentObject but maybe I am wrong 😳

jankrzempek: Ok, next comments changes a bit

pawelgil (approved):
good stuff

PR #85, opened 2021-07-15
Bluetooth time alert

jankrzempek:
What was done:

Now whenever connecting to bluetooth device, the attempt takes 10 second, if it fails than message is popped up

https://trello.com/c/9fmXcS1i


AirCasting/CreateSessionViews/SelectPeripheralView.swift
```diff
@@ -61,16 +60,22 @@ struct SelectPeripheralView: View {
                     } else {
                         connectButton.disabled(true)
                     }
-                }
-                .padding()
-                .frame(maxWidth: .infinity, minHeight: geometry.size.height, alignment: .top)
+                }.alert(isPresented: $showBluetoothAlert, content: {
+                    Alert(title: Text("Connection error"), message: Text("Bluetooth connection failed. Please toggle the power on your device and try again."), dismissButton: .default(Text("Got it!")))
+                })
+                    .padding()
```
smialko, line 66: Please, fix the indention :)

AirCasting/CreateSessionViews/SelectPeripheralView.swift
```diff
@@ -61,16 +60,22 @@ struct SelectPeripheralView: View {
...
+                    .frame(maxWidth: .infinity, minHeight: geometry.size.height, alignment: .top)
             }
-        }
-        
+        }.onChange(of: bluetoothManager.isConnected, perform: { value in
+            showBluetoothAlert = !value
+        })
+            .onAppear(perform: {
+                bluetoothManager.isConnected = true
```
smialko, line 73: Same as above :)

smialko (changes requested):
Generally, the code looks good, but it's not working as expected, it's constantly showing me a "connecting" event though it's been more than 10 s :)

smialko (approved):
Still, fix the indentations ;D

AirCasting/CreateSessionViews/ConnectingABView.swift
```diff
@@ -5,18 +5,18 @@
 //  Created by Lunar on 04/02/2021.
 //
 
-import SwiftUI
 import CoreBluetooth
+import SwiftUI
 
 struct ConnectingABView: View {
-    
+    @Environment(\.presentationMode) var presentationMode
```
pawelgil, line 12: A view should not manage its own dismissal or presentation. Let's leave that in here, but in the future let's handle such stuff from a parent view.

AirCasting/CreateSessionViews/ConnectingABView.swift
```diff
@@ -39,12 +39,21 @@ struct ConnectingABView: View {
...
+                )
+            )
+        }
         .padding()
         .onAppear(perform: {
             bluetoothManager.centralManager.connect(selectedPeripheral,
                                                     options: nil)
+            DispatchQueue.main.asyncAfter(deadline: .now() + 10) {
+                if selectedPeripheral.state == .connecting {
+                    Log.info("Connecting to bluetooth device failed")
```
pawelgil, line 52: `View` is not a great place for app logic. Please consider moving it out - be that a viewmodel or a controller-viewmodel-service combination (which would be perfect for separating concerns, code clarity and testability). Note that it will also require other logic to be moved away from here.

AirCasting/CreateSessionViews/SelectPeripheralView.swift
```diff
@@ -61,16 +60,22 @@ struct SelectPeripheralView: View {
                     } else {
                         connectButton.disabled(true)
                     }
-                }
+                }.alert(isPresented: $showBluetoothAlert, content: {
+                    Alert(title: Text("Connection error"), message: Text("Bluetooth connection failed. Please toggle the power on your device and try again."), dismissButton: .default(Text("Got it!")))
```
pawelgil, line 64: `Strings`

pawelgil (changes requested):

AirCasting/CreateSessionViews/ConnectingABViewModel.swift
```diff
@@ -0,0 +1,29 @@
+// Created by Lunar on 21/07/2021.
+//
+
+import CoreBluetooth
+import Foundation
+
+protocol ConnectingABProtocol {
```
pawelgil, line 7: That's a good direction here for sure, you're getting the hang of it! Couple of kinks still to work out here:

`ConnectingABProtocol` - it looks like a Controller/UseCase/Interactor layer object for me. It handles connecting which is our feature here. Great job on spotting it.

The protocol itself doesn't require `bluetoothManager` property to be exposed tho. It's an implementation detail that clients of this protocol won't ever use or have a need to know about. So it should only be required by an implementing class for this.  The other issue with this is separation - now the `ConnectingABProtocol` is depending on a `BluetoothManager` class which is clearly a service layer (it's bluetooth, very low-level detail of communication)

Now for the implementing class you've decided to go with making a `ViewModel` conform to `ConnectingABProtocol`. Remember that our goal with all those protocols and layers is to separate concerns, responsibilities and make classes do only one thing. The job of the ViewModel is simple - get stuff from the controller/usecase/interactor and make it easily digestable by the View. So making `ConnectingABViewModel` conform to `ConnectingABProtocol` denies all this work you've put into separating the two :)

It's always super helpful to spend 2mins on a simple pseudo-uml diagram:

image

There are a few problems here with the design. As we can see:
- `ConnectingABViewModel` is conforming to `ConnectingABProtocol` and it's really meant to be using one instead. So we're missing a class that would implement `ConnectingABProtocol` on its own.
- `ConnectingABProtocol` is using `BluetoothManager` directly and classes/interfaces in this layer are prohibited from depending on external stuff, it should define it's expectations and the rest of the app should conform to those.
- `ConnectingABProtocol` knows about bluetooth at all. It should define some connection protocol agnostic of connection type. This way if for example AB4 will have a new connection type (wifi or whatever) we won't have to change any existing code, only introduce a new class for connecting with the new protocol. It also makes unit testing easy and shortens compile times which doesn't seem important now but can get really ugly when the project gets fatter

It should really look like this:

image

But also please rename `ConnectingABProtocol` to something:
- without a `Protocol` in name (it's redundant, everyone knows it's a protocol ;) )
- not using a verb
- perhaps not using AB acronym (AirBeam would be a bit more descriptive)

AirCasting/CreateSessionViews/ConnectingABViewModel.swift
```diff
@@ -0,0 +1,29 @@
...
+    var bluetoothManager: BluetoothManager { get }
+    func performConnectingWithin10Second(peripheral: CBPeripheral, completion: @escaping () -> Void)
+}
+
+class ConnectingABViewModel: ConnectingABProtocol {
+    var bluetoothManager: BluetoothManager
```
pawelgil, line 13: see - now ViewModel is mixed with Bluetooth stuff and everything is connected with each other, no chance to test in isolation and any business requirement change would require everything to be changed. Not too great.

AirCasting/CreateSessionViews/ConnectingABViewModel.swift
```diff
@@ -0,0 +1,29 @@
...
+    
+    init(bluetoothManager: BluetoothManager) {
+        self.bluetoothManager = bluetoothManager
+    }
+    
+    func performConnectingWithin10Second(peripheral: CBPeripheral, completion: @escaping () -> Void) {
+        bluetoothManager.centralManager.connect(peripheral, options: nil)
+        DispatchQueue.main.asyncAfter(deadline: .now() + 10) { [self] in
```
pawelgil, line 21: All this is business and service logic that should be separated from views (see: giant comment ;) )

pawelgil (changes requested):

AirCasting/CreateSessionViews/ConnectingABView/AirBeamConnectionController.swift
```diff
@@ -0,0 +1,33 @@
...
+            }
+        }
+    }
+    
+    init(connectingAirBeamServices: ConnectingAirBeamServices) {
+        self.connectingAirBeamServices = connectingAirBeamServices
+    }
+}
+
+protocol AirBeamConnectionController {
```
pawelgil, line 31: protocol should go before the implementing `class` (the rule is that we put high-level concepts above low-level details - for readability)

AirCasting/CreateSessionViews/ConnectingABView/AirBeamConnectionController.swift
```diff
@@ -0,0 +1,33 @@
+// Created by Lunar on 30/07/2021.
+//
+
+import Foundation
+import CoreBluetooth
+
+enum AirBeamServicesConnectionResult {
```
pawelgil, line 7: let's move this to the same file as `ConnectionAirBeamServices` since this detail is not important for `AirBeamConnectionController` users at all

AirCasting/CreateSessionViews/ConnectingABView/AirBeamServices.swift
```diff
@@ -0,0 +1,35 @@
...
+import CoreBluetooth
+
+protocol ConnectingAirBeamServices {
+    func connect(to peripheral: CBPeripheral, timeout: TimeInterval, completion: @escaping (AirBeamServicesConnectionResult) -> Void)
+}
+
+class ConnectingAirBeamServicesBluetooth: ConnectingAirBeamServices {
+    
+    private var connectionInProgress = false
+    let bluetoothConnector: BluetoothConnector
```
pawelgil, line 14: private

AirCasting/CreateSessionViews/ConnectingABView/AirBeamServices.swift
```diff
@@ -0,0 +1,35 @@
...
+        guard !connectionInProgress else { completion(.deviceBusy); return }
+        bluetoothConnector.connect(to: peripheral)
+        DispatchQueue.main.asyncAfter(deadline: .now() + .seconds(Int(timeout))) {
+            if peripheral.state == .connecting {
+                Log.info("Connecting to bluetooth device failed")
+                self.bluetoothConnector.cancelPeripheralConnection(for: peripheral)
+                completion(.timeout)
+            }
+        }
+        var token = NotificationCenter.default.addObserver(forName: Notification.Name(rawValue: "DeviceConnected"), object: nil, queue: nil) { _ in
```
pawelgil, line 30: you have just removed the code sending this notification so this will never be triggered

AirCasting/CreateSessionViews/ConnectingABView/ConnectingABView.swift
```diff
@@ -82,8 +78,9 @@ struct ConnectingABView: View {
         }
     }
 }
-//struct ConnectingABView_Previews: PreviewProvider {
+
+// struct ConnectingABView_Previews: PreviewProvider {
```
pawelgil, line 82: we should uncomment and fix this ([The Boy Scout Rule](https://deviq.com/principles/boy-scout-rule), remember ;) I know, I know, it's annoying)

AirCasting/CreateSessionViews/ConnectingABView/ConnectingABViewModel.swift
```diff
@@ -0,0 +1,26 @@
...
+    }
+    
+    func connectToAirBeam(peripheral: CBPeripheral) {
+        self.airBeamConnectionController.connectToAirBeam(peripheral: peripheral) { success in
+            if success {
+                self.isDeviceConnected = true
+            } else {
+                self.isDeviceConnected = false
+                self.shouldDismiss = true
+            }
```
pawelgil, line 23: if/else block is not needed, we can just assign:
```Swift
self.isDeviceConnected = success
self.shouldDismiss = !success
```

AirCasting/CreateSessionViews/ConnectingABView/ConnectingABViewModel.swift
```diff
@@ -0,0 +1,26 @@
+// Created by Lunar on 21/07/2021.
+//
+
+import Foundation
+import CoreBluetooth
+
+class ConnectingABViewModel: ObservableObject {
```
pawelgil, line 7: I really like how this works out 🧹

AirCasting/CreateSessionViews/SelectPeripheralView.swift
```diff
@@ -61,16 +61,22 @@ struct SelectPeripheralView: View {
...
                 .padding()
                 .frame(maxWidth: .infinity, minHeight: geometry.size.height, alignment: .top)
             }
-        }
-        
+        }.onChange(of: bluetoothManager.isConnected, perform: { value in
+            showBluetoothAlert = !value
+        })
+        .onAppear(perform: {
+            bluetoothManager.isConnected = true
```
pawelgil, line 74: why is this here? It seems unrelated and if needed I feel we need another way since this won't be clear to readers (also we should not use BluetoothManager in View objects). Can you elaborate?

AirCasting/RootAppView.swift
```diff
@@ -7,38 +7,44 @@
...
-            .environmentObject(microphoneManager)
-            .environmentObject(userAuthenticationSession)
-            .environmentObject(persistenceController)
-            .environment(\.managedObjectContext, persistenceController.viewContext)
+                        authorisationService: dependancies.userAuthenticationSession,
+                        persistenceController: dependancies.persistenceController,
+                        baseUrl: dependancies.urlProvider), urlProvider: dependancies.urlProvider)
+            .environmentObject(dependancies.bluetoothManager)
+            .environmentObject(dependancies.microphoneManager)
+            .environmentObject(dependancies.userAuthenticationSession)
```
pawelgil, line 44: nice! In the future we can change this to only pass `dependencies` thru. But now it's not a good time, just making a remark ;)

pawelgil (changes requested):

AirCasting/CreateSessionViews/SelectPeripheralView.swift
jankrzempek: I think the reason that the BT manager is here is not given by us. Mostly it was done by someone else. I am a bit confused now, going to try to delete .onApper but the rest will stay as if it is not mine

pawelgil: All right 👌

AirCasting/CreateSessionViews/ConnectingABView/ConnectingABView.swift
jankrzempek: For sure You are right, I will try to remember then (a bit annoying 😛)

PR #90, opened 2021-07-20
Fix threshold

smialko:
Before: the threshold for the selected stream was used for all streams views (f.e. dots next to measurement value). 

Now: Threshold for the selected stream is used for graph/map/chart that should display data of this selected stream. The dots' colors next to measurement values are based on the threshold for each stream. 

I wasn't able to test this, I'll test as soon as I charge AB3 (I'll have the cable today :) )

### DIFF
--- a/AirCasting.xcodeproj/project.pbxproj
+++ b/AirCasting.xcodeproj/project.pbxproj
@@ -181,6 +181,7 @@
 		DD430B062670FCC000F1BF2E /* GreenButtonStyle.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD430B052670FCC000F1BF2E /* GreenButtonStyle.swift */; };
 		DD430B082670FE6100F1BF2E /* GreenTextButtonStyle.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD430B072670FE6100F1BF2E /* GreenTextButtonStyle.swift */; };
 		DD430B0C26710E0600F1BF2E /* PrivacyOnboarding.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD430B0B26710E0600F1BF2E /* PrivacyOnboarding.swift */; };
+		DD839E0F269EF324006A8625 /* LifeTimeEventsProvider.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD839E0E269EF324006A8625 /* LifeTimeEventsProvider.swift */; };
 		DD700CF22695CE9C0048A17D /* MailView.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD700CF12695CE9C0048A17D /* MailView.swift */; };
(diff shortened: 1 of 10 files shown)


AirCasting/Dashboard/DashboardView.swift
```diff
@@ -40,7 +40,7 @@ struct DashboardView: View {
                 ScrollView(.vertical) {
                     LazyVStack(spacing: 20) {
                         ForEach(sessions, id: \.uuid) { session in
-                            SessionCartView(session: session, thresholds: thresholds)
+                            SessionCartView (session: session, thresholds: thresholds)
```
pawelgil, line 43: let's get rid of the space here

AirCasting/Utils/ThresholdSelector.swift
```diff
@@ -0,0 +1,13 @@
+// Created by Lunar on 16/07/2021.
+//
+
+import Foundation
+
+func thresholdFor(thresholds: [SensorThreshold], stream: MeasurementStreamEntity?) -> SensorThreshold? {
```
pawelgil, line 6: let put it inside a class, a struct or an extension, global funcs should be reserved for a truely global stuff like logging, asserts etc.:

```Swift
extension Array where Element == SensorThreshold {
    func threshold(for stream: MeasurementStreamEntity?) -> SensorThreshold? {
        ...
    }
}
```

AirCasting/SessionViews/StreamsView.swift
```diff
@@ -36,14 +36,27 @@ struct StreamsView: View {
...
+                                          value: lastMicMeasurement(),
+                                          threshold: threshold,
+                                          selectedStream: .constant(dbStream),
+                                          measurementPresentationStyle: measurementPresentationStyle)
+                }
             }
         }
     }
+    
+    func thresholdFor(stream: MeasurementStreamEntity?) -> SensorThreshold? {
```
pawelgil, line 50: I see this in multiple places, let's try to unify since the logic is basically the same everywhere. See my proposition below 😀

AirCasting/Utils/ThresholdSelector.swift
```diff
@@ -0,0 +1,13 @@
...
+    let match = thresholds.first { threshold in
+        let streamName = stream?.sensorName?
+            .lowercased()
+        return threshold.sensorName?.lowercased() == streamName
+    }
+    return match
```
pawelgil, line 12: the `match` is not needed at all, you can just leave `thresholds.first { ... }` as the expression type is matching func return type

pawelgil (commented):
Small improvements suggested. 
Not ideal that we're getting deeper into the "`NSManagedObject` subclasses in `View` layer" rabbit hole, but I guess it's not the PR to fight status quo :)

jankrzempek (approved):
👌🏻🥳

PR #97, opened 2021-07-22
Average measurements for dormant sessions

smialko:
**SUMMARY**
Dormant sessions should display average values.


jankrzempek (approved):
😎

pawelgil (changes requested):
Let's move some strings around

PR #98, opened 2021-07-22
Add ScrollViewReader

smialko:
The bug was: https://trello.com/c/P4N33Xha/75-tabs-for-following-mobile-active-dormant-fixed-in-dashboard

We have a custom section picker and I created a workaround for this, hope it'll work ok. 
I added a ScrollViewReader that scrolls section picker to show the "fixed" tab when the user chooses "dormant", and to show the "following" tab when the user chooses the "active" :)


jankrzempek (approved):
👌

pawelgil (approved):
🔧

PR #102, opened 2021-07-26
No connection alert for sync

pawelgil:
## What was done?

Added an alert that informs the user that synchronization could not be finished if device if offline.

## Tech notes

* Added a `SessionSynchronizerErrorStream` as a plugin point for the `SessionSynchronizer`. This way we can redirect the flow of errors completely different way than standard output goes. This allows for a better separation - "normal" views don't need to worry about failure cases, it's handled by separate code in separate place.
* Temporarily commented out tests that were preventing test target to build @jankrzempek please take care of this


jankrzempek (approved):
👌

smialko (approved):
🥷🏻

PR #103, opened 2021-07-26
Follow session

jankrzempek:
https://trello.com/c/yUexRDqt/41-follow-unfollow-session

What was done:
You are now able to follow and unfollow session. When You create fixed session, at first You should be directed to following tab.

// For now it can be only fully tested with AirBeam


AirCasting/CoreData/Session.swift
```diff
@@ -118,7 +122,12 @@ public enum SessionStatus: Int {
     case NEW = -1
     case RECORDING = 0
     case FINISHED = 1
-    case DISCONNETCED = 2
+    case DISCONNECTED = 2
```
smialko, line 125: 👏

AirCasting/CoreData/Session.swift
```diff
@@ -118,7 +122,12 @@ public enum SessionStatus: Int {
...
+}
+
+enum SessionFollowing: Int {
```
smialko, line 128: Suggestion mode! Maybe change cases to "followed" & "unfollowed"? :)

AirCasting/Services/UpdateSessionParams.swift
```diff
@@ -54,7 +54,7 @@ final class UpdateSessionParamsService {
             oldStream.thresholdVeryHigh = streamOutput.threshold_very_high
             oldStream.gotDeleted = streamOutput.deleted ?? false
             
-            let existingThreshold: SensorThreshold? = try context.existingObject(sensorName: streamOutput.sensor_name)
+            let existingThreshold: SensorThreshold? = try context.existingObject(sensorName: "AirBeam3-PM10")
```
smialko, line 57: ? Why ?

AirCasting/SessionViews/SessionCartView.swift
```diff
@@ -5,60 +5,69 @@
...
     }
     
     var body: some View {
         VStack(alignment: .leading, spacing: 13) {
-            
             header
+            //            if let threshold = thresholdFor(selectedStream: selectedStream) {
+            StreamsView(selectedStream: $selectedStream,
+                        session: session,
+                        threshold: thresholds[0],
```
smialko, line 40: threshold is hardcoded, it shouldn't be.

AirCasting/SessionViews/SessionCartView.swift
```diff
@@ -5,60 +5,69 @@
...
-                        pollutionChart(stream: selectedStream,
-                                       threshold: threshold)
-                    }
-                    if !isCollapsed {
-                        displayButtons(threshold: threshold)
-                    }
+            VStack(alignment: .trailing, spacing: 40) {
+                if let selectedStream = selectedStream, showChart {
+                    pollutionChart(stream: selectedStream,
+                                   threshold: thresholds[0])
```
smialko, line 46: Same a above. 
I know you did that for mocking, for the future please add lots of comments&warnings to avoid situation like this when you missed hardcoded thingies.

AirCasting/SessionViews/SessionCartView.swift
```diff
@@ -5,60 +5,69 @@
...
+                }
+                if !isCollapsed {
+                    displayButtons(threshold: thresholds[0])
                 }
-            } else {
-                SessionLoadingView()
             }
+            //            } else {
+            //                SessionLoadingView()
+            //            }
```
smialko, line 54: it's commented & it shouldn't be

AirCasting/SessionViews/SessionCartView.swift
```diff
@@ -5,60 +5,69 @@
...
         }
         .onChange(of: session.allStreams) { _ in
             selectedStream = session.allStreams?.first
         }
-        .onAppear() {
-            selectedStream = session.allStreams?.first
```
smialko, line 61: please, don't remove that

AirCasting/SessionViews/SessionCartView.swift
```diff
@@ -107,7 +128,13 @@ private extension SessionCartView {
     }
     
     func displayButtons(threshold: SensorThreshold) -> some View {
-        HStack(spacing: 20){
+        HStack(spacing: 20) {
+            if isFollowing && !(session.type == .mobile) {
+                unFollowButton
+            } else if session.type == .fixed || session.type == .following && !isFollowing {
```
smialko, line 134: I don't get the `session.type == .following && !isFollowing` <-- doesn't .following mean that we're following this session?

AirCasting/SessionViews/SessionCartView.swift
```diff
@@ -107,7 +128,13 @@ private extension SessionCartView {
     }
     
     func displayButtons(threshold: SensorThreshold) -> some View {
-        HStack(spacing: 20){
+        HStack(spacing: 20) {
+            if isFollowing && !(session.type == .mobile) {
```
smialko, line 132: I think `isFollowing` would be enough as we only set "isFollowing" on fixed sessions. But if you want to stick with the second part, why do you use !`(session.type == .mobile)` instead of `(session.type == .fixed)`?

AirCasting/SessionViews/SessionCartView.swift
smialko: Wouldn't `session.type == .fixed && !isFollowing` be enough?

smialko (changes requested):
I have some questions & a couple of suggestions ;)

AirCasting/SessionViews/SessionCartView.swift
jankrzempek: Yes, you are right. It was for nothing. Thanks!

AirCasting/SessionViews/SessionCartView.swift
jankrzempek: again, you are right 😄

AirCasting/SessionViews/SessionCartView.swift
jankrzempek: Of course, I misled myself - not the best

AirCasting/Services/UpdateSessionParams.swift
jankrzempek: I probably were sleeping when doing this 😅

AirCasting/CoreData/Session.swift
jankrzempek: I will allow myself to leave it as it is now, because I think that fixed is entry state. And then you can additionally make follow state, but when we go back to entry state (fixed) I think there is no need to say it is unfollow or etc. If it's ok I will leave it 😄

AirCasting/SessionViews/SessionCartView.swift
jankrzempek: Changing right now!

AirCasting/SessionViews/SessionCartView.swift
```diff
@@ -5,42 +5,41 @@
...
+import CoreLocation
+import SwiftUI
 
 struct SessionCartView: View {
-    
+    @State private var isFollowing = false
     @State private var isCollapsed = true
     @State private var selectedStream: MeasurementStreamEntity?
-    
+    @EnvironmentObject var persistenceController: PersistenceController
```
pawelgil, line 18: `PersistenceController` should not be present in here. The whole purpose of `MeasurementStreamStorage` protocol is so we don't use `CoreData` outside of its implementation. If you really need to use storage here it should be injected from outside and used as a `MeasurementStreamStorage`

AirCasting/SessionViews/SessionCartView.swift
```diff
@@ -57,7 +56,20 @@ struct SessionCartView: View {
         .onChange(of: session.allStreams) { _ in
             selectedStream = session.allStreams?.first
         }
-        .onAppear() {
+        .onChange(of: isFollowing) { _ in
+            let measurementStreamStorage = CoreDataMeasurementStreamStorage(persistenceController: persistenceController)
```
pawelgil, line 60: 1. If you really need to use storage here this should at least be injected and used as `MeasurementStreamStorage` protocol, not this concrete class. The protocol is designed so that clients don't know anything about coredata, persistence controllers etc. Use [injection](https://freecontent.manning.com/understanding-constructor-injection/) patterns

2. It would also be good if we didn't put databases inside views (at least with new code, like here) - there **really** should be some middle man objects in between. At least an orchestrating `ViewModel`, but I encourage more separation (ViewModel is still a view layer after all). Consider writing some tests for this - it looks like an important piece of logic.

AirCasting/SessionViews/SessionCartView.swift
```diff
@@ -124,14 +154,14 @@ private extension SessionCartView {
...
-    static var previews: some View {
-        EmptyView()
-        SessionCartView(session: SessionEntity.mock, thresholds: [.mock, .mock])
-            .padding()
-            .previewLayout(.sizeThatFits)
-            .environmentObject(MicrophoneManager(measurementStreamStorage: PreviewMeasurementStreamStorage()))
-    }
-}
-#endif
+// #if DEBUG
```
pawelgil, line 157: Please adjust this code instead of commenting it out

Modules/AirCastingStyling/Sources/AirCastingStyling/FollowButtonStyle.swift
```diff
@@ -0,0 +1,22 @@
+//
+//  File.swift
+//  
+//
+//  Created by Lunar on 19/07/2021.
+//
```
pawelgil, line 6: those headers don't seem consistent with the rest of the codebase

Modules/AirCastingStyling/Sources/AirCastingStyling/UnFollowButtonStyle.swift
```diff
@@ -0,0 +1,24 @@
+//
+//  File.swift
+//  
+//
+//  Created by Lunar on 19/07/2021.
+//
```
pawelgil, line 6: those headers don't seem consistent with the rest of the codebase

AirCasting/SessionViews/SessionCartView.swift
```diff
@@ -5,42 +5,41 @@
 //  Created by Lunar on 08/01/2021.
 //
 
-import SwiftUI
-import CoreLocation
-import CoreData
-import Charts
 import AirCastingStyling
+import Charts
+import CoreData
```
pawelgil, line 10: I know it's been here, but just to point it out as the opportunity presented itself :) : such imports structure is already a pretty big [code smell](https://en.wikipedia.org/wiki/Code_smell) telling us that we probably have some serious architectural problems (no change required here ofc, just pointing out):
```Swift
import CoreData
import SwiftUI
```

pawelgil (changes requested):

AirCasting/CoreData/Session.swift
smialko: Sure, no need to explain, it was just a suggestion :)

AirCasting/Services/UpdateSessionParams.swift
smialko: So you're coding while you sleep! This is how you make lots of views in such a short time ;D

Modules/AirCastingStyling/Sources/AirCastingStyling/UnFollowButtonStyle.swift
smialko: What do you mean? This header is the same as other headers in the AirCastingStyling module.

Modules/AirCastingStyling/Sources/AirCastingStyling/FollowButtonStyle.swift
smialko: Same, they're the same as other headers in the AirCastingStyling module

smialko (approved):
Looks fine to me except for those things Paweł found :D

AirCasting.xcodeproj/project.pbxproj
```diff
@@ -1616,7 +1614,7 @@
 				);
 				PRODUCT_BUNDLE_IDENTIFIER = org.habitatmap.AirCasting;
 				PRODUCT_NAME = "$(TARGET_NAME)";
-				PROVISIONING_PROFILE_SPECIFIER = "AirCasting Ad hoc distribution";
+				PROVISIONING_PROFILE_SPECIFIER = "";
```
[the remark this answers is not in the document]
jankrzempek: For sure, no. I really didn't know about this ☹️

AirCasting/CoreData/MeasurementStreamStorage.swift
```diff
@@ -100,6 +101,21 @@ final class CoreDataMeasurementStreamStorage: MeasurementStreamStorage {
         sessionEntity.status = sessionStatus
         try context.save()
     }
+    
+    func updateSessionIfFollowing(_ sessionFollowing: SessionFollowing, for sessionUUID: SessionUUID) throws {
+        let context = persistenceController.editContext()
+        let sessionEntity = try context.existingSession(uuid: sessionUUID)
+        let copySession = sessionEntity
```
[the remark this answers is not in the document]
jankrzempek: I think it does so, I need to make a copy of the following cell to be able to see it on the following tab and on the fixed tab too. copySession making it great

AirCasting/CoreData/MeasurementStreamStorage.swift
```diff
@@ -112,6 +128,10 @@ final class CoreDataMeasurementStreamStorage: MeasurementStreamStorage {
 #if DEBUG
 /// Only to be used for swiftui previews
 final class PreviewMeasurementStreamStorage: MeasurementStreamStorage {
+    func updateSessionIfFollowing(_ sessionStatus: SessionFollowing, for sessionUUID: SessionUUID) throws {
+        print("Nothing happened")
```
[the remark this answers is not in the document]
jankrzempek: Coppied from other function to make it fancy, changing now

AirCasting/Models/AirBeamFixedSessionCreator.swift
```diff
@@ -38,6 +38,7 @@ final class AirBeamFixedSessionCreator: SessionCreator {
                               deviceType: sessionContext.deviceType,
                               location: sessionContext.startingLocation,
                               startTime: Date(),
+                              followedAt: Date(),
```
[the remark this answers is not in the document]
jankrzempek: Yes, Michael said that we should

Modules/AirCastingStyling/Sources/AirCastingStyling/FollowButtonStyle.swift
jankrzempek: Can't see really a difference too, could you please point it out?

AirCasting/CoreData/MeasurementStreamStorage.swift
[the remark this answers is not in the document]
pawelgil: please try without it and see if it changes anything. I'd be very surprised if it did. Reason is that `SessionEntity` is a class, so a reference type (unlike structs which are value types, [see this](https://developer.apple.com/swift/blog/?id=10)), assigning does not make a copy of it.

AirCasting/CoreData/MeasurementStreamStorage.swift
```diff
@@ -101,6 +103,27 @@ final class CoreDataMeasurementStreamStorage: MeasurementStreamStorage {
         try context.save()
     }
 
+    func updateSessionIfFollowing(_ sessionFollowing: SessionFollowing, for sessionUUID: SessionUUID) throws {
+        let context = persistenceController.editContext()
+        let sessionEntity = try context.existingSession(uuid: sessionUUID)
+        let copySession = sessionEntity
+        if sessionFollowing.rawValue == 1 {
+            sessionEntity.followedAt = Date()
```
pawelgil, line 111: all this should also be inside `performBlockAndWait`. Any reads/writes for CoreData-managed classes (`NSManagedObject` subclasses) should be wrapped with it.

AirCasting/SessionViews/SessionCartView.swift
```diff
@@ -57,7 +56,15 @@ struct SessionCartView: View {
...
             selectedStream = session.allStreams?.first
         }
-        .onAppear() {
+        .onChange(of: isFollowing) { _ in
+            sessionCartViewModel.toggleFollowing(for: session)
+        }
+        .onAppear {
+            if session.followedAt != nil {
+                isFollowing = true
+            }
```
pawelgil, line 65: no need for if, just assign:
`isFollowing = session.followedAt != nil`

AirCasting/SessionViews/SessionCartView.swift
```diff
@@ -98,6 +105,18 @@ private extension SessionCartView {
         }
     }
     
+    var followButton: some View {
+        Button("follow") {
```
pawelgil, line 109: Strings

AirCasting/SessionViews/SessionCartView.swift
```diff
@@ -98,6 +105,18 @@ private extension SessionCartView {
...
+            isFollowing.toggle()
+        }.buttonStyle(FollowButtonStyle())
+    }
+    
+    var unFollowButton: some View {
+        Button("unfollow") {
```
pawelgil, line 115: Strings

AirCasting/SessionViews/SessionCartView.swift
```diff
@@ -107,7 +126,13 @@ private extension SessionCartView {
     }
     
     func displayButtons(threshold: SensorThreshold) -> some View {
-        HStack(spacing: 20){
+        HStack(spacing: 20) {
+            if isFollowing {
+                unFollowButton
+            } else if session.type == .fixed, !isFollowing {
```
pawelgil, line 132: we just created a viewmodel for this class, maybe we should use it (not a must tho)

AirCastingTests/Logic/SessionCartTests/SessionCartViewModelTest.swift
```diff
@@ -0,0 +1,9 @@
+// Created by Lunar on 29/07/2021.
+//
+
+import XCTest
+@testable import AirCasting
+
+class SessionCartViewModelTest: XCTestCase {
```
pawelgil, line 7: That's a start ;)

Modules/AirCastingStyling/Sources/AirCastingStyling/FollowButtonStyle.swift
pawelgil: Ok, didn't consider `AirCastingStyling`. We should change it some day to the same format we're using in the app module (no File.swift nonsense)

pawelgil (changes requested):
small stuff left to correct and we'll be ready to go :)

AirCasting/CoreData/MeasurementStreamStorage.swift
jankrzempek: for this "please try without it and see if it changes anything. I'd be very surprised if it did. Reason is that SessionEntity is a class, so a reference type (unlike structs which are value types, see this), assigning does not make a copy of it." @pawelgil the copy needs to be made to make it appear on following like as on the fixed at the same time. Tested it now

AirCastingTests/Logic/SessionCartTests/SessionCartViewModelTest.swift
jankrzempek: And I think for now and only, because I've tried and lose. Will try again in near feature if allowed?

AirCasting/CoreData/MeasurementStreamStorage.swift
```diff
@@ -100,6 +101,21 @@ final class CoreDataMeasurementStreamStorage: MeasurementStreamStorage {
         sessionEntity.status = sessionStatus
         try context.save()
     }
+    
+    func updateSessionIfFollowing(_ sessionFollowing: SessionFollowing, for sessionUUID: SessionUUID) throws {
+        let context = persistenceController.editContext()
```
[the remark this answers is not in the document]
pawelgil: reminder: this is still pending

AirCasting/SessionViews/SessionCartViewModel.swift
```diff
@@ -0,0 +1,47 @@
+// Created by Lunar on 29/07/2021.
+//
+
+import Foundation
+
+class SessionCartViewModel: ObservableObject {
+    private let sessionCartFollowing: SessionCartFollowing
+    @Published var isFollowing: Bool = false
```
pawelgil, line 8: I don't think we should have a default here, we need to determine correct value in `init`

AirCasting/SessionViews/SessionCartViewModel.swift
```diff
@@ -0,0 +1,47 @@
...
+    
+    init(sessionCartFollowing: SessionCartFollowing) {
+        self.sessionCartFollowing = sessionCartFollowing
```
pawelgil, line 11: here - add this code instead of defaulting to `false`:

```Swift
self.isFollowing = sessionCartFollowing == .following
```
(or whatever the enum value is for followed state)

AirCasting/SessionViews/SessionCartViewModel.swift
```diff
@@ -0,0 +1,47 @@
...
+    
+    func toggleFollowing(for session: SessionEntity) {
+        session.followedAt != nil ? sessionCartFollowing.makeFollowing(for: session) : sessionCartFollowing.makeNotFollowing(for: session)
+        switch showFollowingButton(for: session) {
+            case true: isFollowing = true
+            case false: isFollowing = false
+        }
+    }
+    
+    func showFollowingButton(for session: SessionEntity) -> Bool {
```
pawelgil, line 22: I'm not sure what this function is doing. For sure not showing any buttons ;) please rename (and also make sure all the things that can be `private` in this class are)

AirCasting/CoreData/MeasurementStreamStorage.swift
```diff
@@ -1,16 +1,17 @@
...
+import CoreLocation
+import Foundation
 
 protocol MeasurementStreamStorage {
     func addMeasurement(_ measurement: Measurement, toStreamWithID id: MeasurementStreamLocalID) throws
     func createMeasurementStream(_ stream: MeasurementStream, for sessionUUID: SessionUUID) throws -> MeasurementStreamLocalID
     func createSession(_ session: Session) throws
     func createSessionAndMeasurementStream(_ session: Session, _ stream: MeasurementStream) throws -> MeasurementStreamLocalID
     func updateSessionStatus(_ sessionStatus: SessionStatus, for sessionUUID: SessionUUID) throws
+    func updateSessionIfFollowing(_ sessionStatus: SessionFollowing, for sessionUUID: SessionUUID) throws
```
pawelgil, line 14: I think the name is not correct anymore. What would you say to a rename (maybe just `updateSessionFollowing`?)

pawelgil (changes requested):

AirCasting/SessionViews/SessionCartViewModel.swift
jankrzempek: we don't have any enum here :/. I tried with this:   _self.session.followedAt != nil ? (isFollowing = true) : (isFollowing = false)_  but I cannot get session from anywhere. Could You please help me with this?

pawelgil: @jankrzempek this will require some more structural changes to the solution (which would also making it cleaner). Take a look at my proposition : #126

pawelgil (changes requested):

AirCasting/SessionViews/SessionCartViewModel.swift
jankrzempek: I went through it, awesome code. I imagine it would be hard for me to write it in that way at leat for now, but Thank You for showing me 😇

pawelgil (approved):

PR #105, opened 2021-07-27
Naive dBFS -> dB convertion mechanism added

pawelgil:
This is so bad, but converting dbfs values to db is not possible without at least some calibration (experimented with many other "tricks", but it always boil down to this, so no point in doing any advanced algo here if we don't have this much required anchor)


jankrzempek (approved):
🥳 Stay with us!

smialko (approved):
Good job doing it and not leaving the project 🥷🏻

PR #110, opened 2021-07-28
Changing backend settings - additional changes

jankrzempek:
What was done:

Now there is only one url and if it changes then the whole app, including signing in depends on it.
Done things from the checkboxes from the ticket.

https://trello.com/c/kAzjpy2N/117-settings-backend-url

### DIFF
--- a/AirCasting/Settings/BackendSettingsView.swift
+++ b/AirCasting/Settings/BackendSettingsView.swift
@@ -13,6 +13,7 @@ struct BackendSettingsView: View {
     @State private var pathText: String = ""
     @State private var portText: String = ""
     @State private var url: URL?
+    @State private var alertPresented = false
     private var urlWithoutPort: String? {
         let components = URLComponents(url: urlProvider.baseAppURL, resolvingAgainstBaseURL: false)!
         return components.host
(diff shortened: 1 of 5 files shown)


AirCasting/Settings/BackendSettingsView.swift
```diff
@@ -53,6 +54,7 @@ struct BackendSettingsView: View {
         Button {
             urlProvider.baseAppURL = url ?? URL(string: "http://aircasting.org/api")!
             presentationMode.wrappedValue.dismiss()
+            try! logoutController.logout()
```
pawelgil, line 57: We really should not crash on error here, let's handle it somehow

pawelgil (changes requested):

AirCasting/Settings/BackendSettingsView.swift
jankrzempek: For sure, nonsense, trying now

AirCasting/Settings/BackendSettingsView.swift
```diff
@@ -53,6 +54,11 @@ struct BackendSettingsView: View {
         Button {
             urlProvider.baseAppURL = url ?? URL(string: "http://aircasting.org/api")!
             presentationMode.wrappedValue.dismiss()
+            do {
+                try logoutController.logout()
+            } catch {
+                Log.info("Error when loging out")
```
pawelgil, line 60: We need this log , but let's also log this error so we can see why it failed. 
But I think we also need an alert here to inform the user that logging out failed (sorry I didn't make it clear how to handle this error, my bad)

pawelgil (changes requested):

pawelgil (approved):

PR #136, opened 2021-08-10
Fixed contribute switch

pawelgil:
The switch was starting from the `On` position which was not reflecting actual setting state.

PR #135, opened 2021-08-10
Uploading measurements along session synchronization

pawelgil:
* Datastructs updated
* Unit tests adjusted

PR #137, opened 2021-08-10
Session uploads and data baselined with android platform

pawelgil:
Baselining so that sessions uploaded from the iOS application are visible on the web app:
* indoor flag fixed for mobile sessions (@smialko please confirm this is correct)
* device name for microphone changed from `Phone Microphone-dB` to `Phone Microphone` to match Android

Note that for this PR to take effect we'll need to also have #135 and #136

### DIFF
--- a/AirCasting/CreateSessionViews/CreateSessionDetailsView.swift
+++ b/AirCasting/CreateSessionViews/CreateSessionDetailsView.swift
@@ -143,7 +143,7 @@ private extension CreateSessionDetailsView {
 
     func getAndSaveStartingLocation() {
         let fakeLocation = CLLocationCoordinate2D(latitude: 200.0, longitude: 200.0)
-        if isIndoor {
+        if isIndoor && sessionContext.sessionType == .fixed {
             sessionContext.startingLocation = fakeLocation
         } else {
             sessionContext.obtainCurrentLocation()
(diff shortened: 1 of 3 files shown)


jankrzempek (approved):
😎

AirCasting/CreateSessionViews/CreateSessionDetailsView.swift
```diff
@@ -143,7 +143,7 @@ private extension CreateSessionDetailsView {
 
     func getAndSaveStartingLocation() {
         let fakeLocation = CLLocationCoordinate2D(latitude: 200.0, longitude: 200.0)
-        if isIndoor {
+        if isIndoor && sessionContext.sessionType == .fixed {
```
smialko, line 146: Mobile sessions can't be indoor so it's not necessary, on the other hand, we can leave it as it is to make sure that we're not sending fake locations for mobile sessions flagged as indoor by mistake 🤔

smialko (approved):
👏

AirCasting/CreateSessionViews/CreateSessionDetailsView.swift
pawelgil: turns out this flag starts as `true` in this view and the switch is not visible for mobile sessions so the execution was indeed getting here and sending false location (200/200) to server :) I consider this a hotfix because all this should really be handled elsewhere

PR #138, opened 2021-08-10
Removed dead logic for locally updating session end time

pawelgil:
[Trello](https://trello.com/c/94mcyT22)

Take a look. I think this is already how it works, just removed the unused code.


smialko (approved):
good catch 🥷🏻

jankrzempek (approved):
🥳

PR #140, opened 2021-08-11
Doubled streams issue fixed

pawelgil:
Matching issue caused by the `id` field not being set during session persistence


jankrzempek (approved):
🥳🥳🥳

smialko:
Thanks so much, I used that in my PR so we can delete this branch :)

PR #141, opened 2021-08-12
Microphone and bluetooth bug

jankrzempek:
There were two bugs:

One of them was with mobile bluetooth session -> bluetooth popup has not appeared before going to searching devices -- Done

Second was with microphone session -> When you have proceed and then try to go back, app crashed (there was unused line of code) -- Done


pawelgil (approved):
Let's change a name of this PR, otherwise all good!

PR #142, opened 2021-08-12
Search location screen wizard

jankrzempek:
What was done:

Added screen to fix session - when using outdoor user now is able to choose custom location or swipe map and center of it always will be chosen. 

Additional change: Matched the house icon in the TabBar

https://trello.com/c/99hIb9qu/194-fixed-session-wizard-match-screens-to-android-search-screen

### DIFF
--- a/AirCasting/AuthViews/UserSettings.swift
+++ b/AirCasting/AuthViews/UserSettings.swift
@@ -6,17 +6,15 @@ import Foundation
 class UserSettings: ObservableObject {
     private let userDefaults: UserDefaults
     
-    var contributingToCrowdMap: Bool {
-        set {
-            objectWillChange.send()
-            userDefaults.setValue(newValue, forKey: "crowdMap")
-        }
-        get {
-            userDefaults.bool(forKey: "crowdMap")
+    @Published var contributingToCrowdMap: Bool {
+        didSet {
+            Log.info("Changed crowdmap contribution setting to \(contributingToCrowdMap ? "ON" : "OFF")")
+            userDefaults.setValue(contributingToCrowdMap, forKey: "crowdMap")
         }
     }
     
     init(userDefaults: UserDefaults = .standard) {
+        self.contributingToCrowdMap = userDefaults.bool(forKey: "crowdMap")
         self.userDefaults = userDefaults
     }
 }
--- a/AirCasting/AuthViews/UserSettings.swift
+++ b/AirCasting/AuthViews/UserSettings.swift
@@ -16,26 +16,6 @@ class UserSettings: ObservableObject {
         }
     }
     
-    var lat: Double {
-        set {
-            objectWillChange.send()
-            userDefaults.setValue(newValue, forKey: "lat")
-        }
-        get {
-            userDefaults.double(forKey: "lat")
-        }
-    }
-    
-    var lon: Double {
-        set {
-            objectWillChange.send()
-            userDefaults.setValue(newValue, forKey: "lon")
-        }
-        get {
-            userDefaults.double(forKey: "lon")
-        }
-    }
-    
     init(userDefaults: UserDefaults = .standard) {
         self.userDefaults = userDefaults
     }
--- a/AirCasting/AuthViews/UserSettings.swift
+++ b/AirCasting/AuthViews/UserSettings.swift
@@ -16,6 +16,26 @@ class UserSettings: ObservableObject {
         }
     }
     
+    var lat: Double {
+        set {
+            objectWillChange.send()
+            userDefaults.setValue(newValue, forKey: "lat")
+        }
+        get {
+            userDefaults.double(forKey: "lat")
+        }
+    }
+    
+    var lon: Double {
+        set {
+            objectWillChange.send()
+            userDefaults.setValue(newValue, forKey: "lon")
+        }
+        get {
+            userDefaults.double(forKey: "lon")
+        }
+    }
+    
     init(userDefaults: UserDefaults = .standard) {
         self.userDefaults = userDefaults
     }
--- a/AirCasting/CreateSessionViews/ChooseCustomLocationView.swift
+++ b/AirCasting/CreateSessionViews/ChooseCustomLocationView.swift
@@ -49,7 +49,7 @@ struct ChooseCustomLocationView: View {
     }
 
     var mapGoogle: some View {
-        GoogleMapView()
+        GoogleMapView(pathPoints: [])
     }
 
     var dot: some View {
--- a/AirCasting/CreateSessionViews/ChooseCustomLocationView.swift
+++ b/AirCasting/CreateSessionViews/ChooseCustomLocationView.swift
@@ -1,24 +1,22 @@
 // Created by Lunar on 11/08/2021.
 //
 
-import SwiftUI
 import AirCastingStyling
-import CoreLocation
+import SwiftUI
 
-struct LocationPopUpView: View {
+struct ChooseCustomLocationView: View {
     let sessionCreator: SessionCreator
-    @State var sessionName: String = ""
     @State private var isConfirmCreatingSessionActive: Bool = false
-    @State private var text = ""
     @State private var location = ""
-    @State private var lan: Double = 50.0
-    @State private var long: Double = 19.0
     @State var isLocationPopupPresented = false
     @Binding var creatingSessionFlowContinues: Bool
+    @Binding var sessionName: String
+
     var body: some View {
-        VStack() {
+        VStack(spacing: 40) {
+            ProgressView(value: 0.85)
             titleLabel
-            createTextfield(placeholder: "Session location", binding: $location)
+            createTextfield(placeholder: Strings.ChooseCustomLocationView.sessionLocation, binding: $location)
                 .disabled(true)
                 .onTapGesture {
                     isLocationPopupPresented.toggle()
@@ -30,7 +28,7 @@ struct LocationPopUpView: View {
             Button(action: {
                 isConfirmCreatingSessionActive.toggle()
             }, label: {
-                Text(Strings.ConfirmCreatingSessionView.startRecording)
+                Text(Strings.ChooseCustomLocationView.continueButton)
                     .bold()
             }).background(
                 NavigationLink(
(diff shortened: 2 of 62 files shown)


AirCasting/CreateSessionViews/ChooseCustomLocationView.swift
```diff
@@ -0,0 +1,74 @@
...
+struct ChooseCustomLocationView: View {
+    let sessionCreator: SessionCreator
+    @State private var isConfirmCreatingSessionActive: Bool = false
+    @State private var location = ""
+    @State var isLocationPopupPresented = false
+    @Binding var creatingSessionFlowContinues: Bool
+    @Binding var sessionName: String
+
+    var body: some View {
+        VStack(spacing: 40) {
```
pawelgil, line 16: this is quite a big body, let's split this up a bit to make it more clear

AirCasting/CreateSessionViews/ConfirmCreatingSessionView.swift
```diff
@@ -37,40 +38,54 @@ struct ConfirmCreatingSessionView: View {
...
     private var someText: some View {
         Text(Strings.ConfirmCreatingSessionView.contentViewText_1)
             + Text(sessionType)
             .foregroundColor(.accentColor)
             + Text(Strings.ConfirmCreatingSessionView.contentViewText_2)
             + Text(sessionName)
             .foregroundColor(.accentColor)
-            + Text(Strings.ConfirmCreatingSessionView.contentViewText_3)
-//            + Text(Strings.ConfirmCreatingSessionView.contentViewText_4)
+            + text
```
pawelgil, line 53: let's take the opportunity and make this name a little more descriptive since `text` is a bit frugal

AirCasting/CreateSessionViews/ConfirmCreatingSessionView.swift
```diff
@@ -92,8 +107,13 @@ struct ConfirmCreatingSessionView: View {
                 Text(Strings.ConfirmCreatingSessionView.startRecording)
                     .bold()
             })
-            .buttonStyle(BlueButtonStyle())
-        }.padding()
+                .buttonStyle(BlueButtonStyle())
+        }.onAppear(perform: {
+            let lat = locationTracker.googleLocation[0].location.latitude
```
pawelgil: is it guaranteed that we'll have at least one location here so we don't crash on `[0]` ?

AirCasting/CreateSessionViews/CreateSessionDetailsView.swift
```diff
@@ -67,7 +70,11 @@ private extension CreateSessionDetailsView {
                 sessionContext.isIndoor = isIndoor
             }
             getAndSaveStartingLocation()
-            isConfirmCreatingSessionActive = true
+            if isIndoor {
+                isConfirmCreatingSessionActive = true
+            } else {
+               isLocationSessionDetailsActive = true
+            }
```
pawelgil, line 96: we can do:

```Swift
isConfirmCreatingSessionActive = isIndoor
isLocationSessionDetailsActive = !isIndoor
```
?
It's 2LOC vs 4LOC ;)

AirCasting/CreateSessionViews/CreateSessionDetailsView.swift
```diff
@@ -145,8 +159,12 @@ private extension CreateSessionDetailsView {
         let fakeLocation = CLLocationCoordinate2D(latitude: 200.0, longitude: 200.0)
         if isIndoor {
             sessionContext.startingLocation = fakeLocation
+            locationTracker.googleLocation = [PathPoint(location: CLLocationCoordinate2D(latitude: 200.0, longitude: 200.0), measurement: 20.0)]
         } else {
-            sessionContext.obtainCurrentLocation()
+            let lat = (locationTracker.locationManager.location?.coordinate.latitude)!
```
pawelgil, line 172: are you sure it's safe to force unwrap?

AirCasting/CreateSessionViews/PlacePicker.swift
```diff
@@ -0,0 +1,60 @@
...
+        
+        let autocompleteController = GMSAutocompleteViewController()
+        autocompleteController.delegate = context.coordinator
+
+        let fields: GMSPlaceField = GMSPlaceField(rawValue: UInt(GMSPlaceField.name.rawValue) | UInt(GMSPlaceField.coordinate.rawValue))
+        autocompleteController.placeFields = fields
+
+        let filter = GMSAutocompleteFilter()
+        filter.type = .address
+        filter.type = .geocode
```
pawelgil, line 23: you assign to `type` twice here. I think this will be equivalent to just `filter.type = .geocode` ?

AirCasting/Map/LocationTracker.swift
```diff
@@ -13,13 +13,18 @@ class LocationTracker: NSObject, ObservableObject, CLLocationManagerDelegate {
...
     
     init(locationManager: CLLocationManager) {
         self.locationManager = locationManager
         self.locationManager.desiredAccuracy = kCLLocationAccuracyBest
         switch locationManager.authorizationStatus {
-            case .authorizedAlways, .authorizedWhenInUse: self.locationGranted = .granted
-            case .denied, .notDetermined, .restricted:  self.locationGranted = .denied
+            case .authorizedAlways, .authorizedWhenInUse:
+                self.locationGranted = .granted
+                googleLocation = [PathPoint(location: CLLocationCoordinate2D(latitude: (locationManager.location?.coordinate.latitude)!, longitude: (locationManager.location?.coordinate.longitude)!), measurement: 20)]
```
pawelgil, line 24: we 100% sure force unwrapping location here is safe?

AirCasting/Models/SessionContext.swift
```diff
@@ -28,13 +28,13 @@ final class CreateSessionContext: ObservableObject {
...
+    func obtainCurrentLocation(lat: Double, log: Double) {
         locationProvider = LocationProvider()
         locationProvider?.requestLocation()
-        
-        locationSink = locationProvider?.$currentLocation
-            .sink(receiveValue: { [weak self] (location) in
-                self?.startingLocation = location?.coordinate
-            })
+        startingLocation = CLLocationCoordinate2D(latitude: lat, longitude: log)
+//        locationSink = locationProvider?.$currentLocation
```
pawelgil, line 35: let's not leave commented out code

pawelgil (changes requested):
Not a Coordinator expert so not a very deep review. Found some minor stuff

AirCasting/CreateSessionViews/ConfirmCreatingSessionView.swift
jankrzempek: It is not in the code anymore 😄

AirCasting/CreateSessionViews/CreateSessionDetailsView.swift
jankrzempek: Nice, thank You 🥳

AirCasting/CreateSessionViews/CreateSessionDetailsView.swift
jankrzempek: I can say that I am never 100% sure, I will handle that

AirCasting/Map/LocationTracker.swift
jankrzempek: I think yes, because it will be only triggered when ` case .authorizedAlways, .authorizedWhenInUse:` so current location should be available then

AirCasting/CreateSessionViews/CreateSessionDetailsView.swift
```diff
@@ -143,10 +155,21 @@ private extension CreateSessionDetailsView {
 
     func getAndSaveStartingLocation() {
         let fakeLocation = CLLocationCoordinate2D(latitude: 200.0, longitude: 200.0)
-        if isIndoor {
-            sessionContext.startingLocation = fakeLocation
+        if sessionContext.sessionType == .fixed {
+            if isIndoor {
+                sessionContext.startingLocation = fakeLocation
+                locationTracker.googleLocation = [PathPoint(location: CLLocationCoordinate2D(latitude: 200.0, longitude: 200.0), measurement: 20.0)]
```
smialko, line 161: Hmm, I'm not sure about using `measurement: 20.0`, could you leave a comment with some explanation why it's hard coded to be 20?

smialko (commented):
Nice! I have some questions though :D

AirCasting/Map/LocationTracker.swift
smialko: +1 to what Paweł wrote, I'm not a fan of force unwrapping so are 100% sure? :D

AirCasting/CreateSessionViews/CreateSessionDetailsView.swift
jankrzempek: I should say how it is so: when I was doing it, I didn't know what to write there, so that's why 20. Could you please tell me what should be there because I don't know really

smialko: Maybe we could create something like "fakeMeasurement" (similar to fake location) for this, what do you say? Because as far as I see it influence only the color of the dot, am I right? :D

jankrzempek: Yes, for the point which I have some knowledge. So what measurement point will we choose for fake one?

smialko: Hmm, dunno, I guess it would be best if we don't use it for anything, even coloring the dot (color should always be green at this point, right?). Then we can do whatever we want with this measurement, just describe it well in comment for future generations :D

AirCasting/AuthViews/UserSettings.swift
```diff
@@ -16,6 +16,26 @@ class UserSettings: ObservableObject {
         }
     }
     
+    var lat: Double {
+        set {
+            objectWillChange.send()
+            userDefaults.setValue(newValue, forKey: "lat")
```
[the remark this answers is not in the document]
pawelgil: I'm also wondering - is it really a setting? `UserSettings` should just handle actual user settings that need to persist across app launches and which are settable by the user. I don't think it's the case here with lat/long?

AirCasting/CreateSessionViews/CreateSessionDetailsView.swift
pawelgil: +1, it's super unclear why the measurement is here. We should aim with the code quality to not be forced to put such explanation comments at all. I also don't like exposing `googleLocation`. 3rd party stuff should be hidden away (but let's not change it since we have app to finish)

AirCasting/AuthViews/UserSettings.swift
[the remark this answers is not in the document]
pawelgil: after CR: Is it even used anywhere?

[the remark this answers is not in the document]
jankrzempek: no. it is not. Thank You!

AirCasting/CreateSessionViews/CreateSessionDetailsView.swift
smialko: Like Paweł wrote, let it stay, just put a warning here for future ;)

smialko (approved):
After you follow Paweł's suggestions, you're good to merge :)

AirCasting/CreateSessionViews/ConfirmCreatingSessionView.swift
```diff
@@ -37,40 +38,58 @@ struct ConfirmCreatingSessionView: View {
...
     private var someText: some View {
         Text(Strings.ConfirmCreatingSessionView.contentViewText_1)
             + Text(sessionType)
             .foregroundColor(.accentColor)
             + Text(Strings.ConfirmCreatingSessionView.contentViewText_2)
             + Text(sessionName)
             .foregroundColor(.accentColor)
-            + Text(Strings.ConfirmCreatingSessionView.contentViewText_3)
-//            + Text(Strings.ConfirmCreatingSessionView.contentViewText_4)
+            + contentViewText_Ending
```
pawelgil, line 53: as per standard Swift conventions (for production code, tests sometimes have those) we shouldn't use snake_case at all.

pawelgil (approved):

PR #148, opened 2021-08-16
Loading indicator fetched session

jankrzempek:
https://trello.com/c/CgoklEjD/203-loading-indicator-when-sessions-are-being-fetched-after-login


AirCasting/Dashboard/EmptyDashboardView.swift
```diff
@@ -10,13 +10,19 @@ import SwiftUI
...
+    @EnvironmentObject var selectedSection: SelectSection
+    var settings: BaseURLProvider
     var body: some View {
         emptyState
     }
 
     private var emptyState: some View {
         VStack(spacing: 45) {
             VStack {
+                if (selectedSection.selectedSection == .mobileDormant || selectedSection.selectedSection == .fixed) && settings.sessionSynced == false {
```
pawelgil, line 22: this is a complicated boolean expression. Let's put it in a helper variable with descriptive name so we don't need to decipher this each time:
```Swift
let someThing = ...
if someThing {
    ....
}
```

AirCasting/Dashboard/EmptyDashboardView.swift
```diff
@@ -10,13 +10,19 @@ import SwiftUI
...
+                    ProgressView("Fetching...")
```
pawelgil, line 23: Strings

AirCasting/SessionsSynchronization/Controller/SessionSynchronizationController.swift
```diff
@@ -13,6 +13,7 @@ final class SessionSynchronizationController: SessionSynchronizer {
     private let downstream: SessionDownstream
     private let upstream: SessionUpstream
     private let store: SessionSynchronizationStore
+    private var settings: BaseURLProvider
```
pawelgil, line 16: well it's not settings, it's `urlProvider` :) how about changing this name to be more precise?

Also - this should really not be shoved in here like that. This class should *only* be responsible for what it's meant to be doing - synchronization. It's really not a good idea to put side effects like that here. Try to change it so that we don't introduce this dependency and let's be a little bit more careful where we put logic in

AirCasting/URLProviders/BaseURLProvider.swift
```diff
@@ -5,10 +5,12 @@ import Foundation
 
 protocol BaseURLProvider {
     var baseAppURL: URL { get set }
+    var sessionSynced: Bool { get set }
```
pawelgil, line 8: why should an object designed for providing us with URL have such a flag also? :) Not a good place for that

pawelgil (changes requested):

AirCasting/Dashboard/EmptyDashboardView.swift
jankrzempek: I was not able to make it into var because of ` @EnvironmentObject var selectedSection: SelectSection` but I put it into function for now at least

AirCasting/URLProviders/BaseURLProvider.swift
jankrzempek: Yeah, you are right :/

AirCasting/SessionsSynchronization/Controller/SessionSynchronizationController.swift
jankrzempek: I know that this probably won't be a great idea but I made a public var for now - because I really can't think about something simple yet so efficient

AirCasting/SessionsSynchronization/Adapters/DataStore/SessionSynchronizationDatabase.swift
```diff
@@ -18,7 +18,7 @@ final class SessionSynchronizationDatabase: SessionSynchronizationStore {
     }
     
     private let database: DatabaseType
-    private let dataConverter = SynchronizationDataConterter()
+    private let dataConverter = SynchronizationDataContainer()
```
pawelgil, line 21: it should be `converter` ;)

AirCasting/SessionsSynchronization/Controller/SessionSynchronizationController.swift
```diff
@@ -5,15 +5,18 @@ import Foundation
 import Combine
 import CoreLocation
 
+public var isSyncing = false
```
pawelgil, line 8: no no, this cannot be done this way. Consider global vars as strictly forbidden

pawelgil (changes requested):

AirCasting/SessionsSynchronization/Controller/SessionSynchronizationController.swift
jankrzempek: I thought that You will say something like this 😅. Could You please tell me, give me some clue how to do it well?

AirCasting/SessionsSynchronization/Adapters/DataStore/SessionSynchronizationDatabase.swift
jankrzempek: My bad 🥶

AirCasting/SessionsSynchronization/Controller/SessionSynchronizationController.swift
```diff
@@ -36,12 +38,16 @@ final class SessionSynchronizationController: SessionSynchronizer {
     func triggerSynchronization(completion: (() -> Void)?) {
         lock.lock(); defer { lock.unlock() }
         if syncInProgress { return }
-        syncInProgress = true
+        DispatchQueue.main.async {
```
pawelgil, line 41: this flag is guarding us against nasty race condition bugs, it needs to be set here in a synchronous manner. It's not designed to be a UI-driver. I would suggest leaving the class as it was and build a ViewModel on top instead.

AirCasting/SessionsSynchronization/Controller/Utils/SessionsSynchronizationStoreDataConverter.swift
```diff
@@ -4,7 +4,7 @@
 import Foundation
 
 /// Provides conversion functions between various sessions sync data structures
-struct SynchronizationDataConterter {
+struct SynchronizationDataConverter {
```
pawelgil, line 7: 👌

AirCasting/Dashboard/EmptyDashboardView.swift
```diff
@@ -10,15 +10,25 @@ import SwiftUI
 
 struct EmptyDashboardView: View {
     @EnvironmentObject private var tabSelection: TabBarSelection
+    @EnvironmentObject var selectedSection: SelectSection
+    @EnvironmentObject var sessionSynchronizationController: SessionSynchronizationController
     var body: some View {
         emptyState
     }
+    var shouldSessionFetch: Bool {
+        (selectedSection.selectedSection == .mobileDormant || selectedSection.selectedSection == .fixed) && sessionSynchronizationController.syncInProgress
```
pawelgil, line 19: I think it would be better to keep `SessionSynchronizationController` as a non-ObservableObject entity and create a ViewModel that in turn will be a proper `ObservableObject` and will contain all this UI logic. Two reasons:
* any time the synchronization state changes it will currently cause to reload basically all apps UI since it's used by `AirCastingApp.swift`. We can avoid that and only reload the UI when and where it's actually needed by creating a VM
* code clarity - we move out the logic part of the UI to a special-purposed class (VM) which makes `View` only about layout as it should.
* mitigating conflicts - any logic put inside views will generate lots of conflicts for us as all departments are meeting in the same files instead of being separated. We're experiencing this already a little.

pawelgil (changes requested):
I would strongly suggest getting a VM in the middle here. If you want I can take over @smialko

smialko (approved):

AirCasting/SessionsSynchronization/Controller/SessionSynchronizationController.swift
```diff
@@ -6,18 +6,20 @@ import Combine
...
     /// A plugin point for error handlers
     var errorStream: SessionSynchronizerErrorStream?
     
     private let synchronizationContextProvider: SessionSynchronizationContextProvidable
     private let downstream: SessionDownstream
     private let upstream: SessionUpstream
     private let store: SessionSynchronizationStore
-    private let dataConverter = SynchronizationDataConterter()
+    private let dataConverter = SynchronizationDataConverter()
+    private let viewModel: SessionSynchronizationViewModel
```
pawelgil, line 18: view model cannot be used by the controller object, it's not how it's supposed to work. This class should be totally independent from other layers of the code like UI or the networking

AirCasting/SessionsSynchronization/Controller/SessionSynchronizationController.swift
```diff
@@ -6,18 +6,20 @@ import Combine
...
     
     // Progress tracking for filtering requests while already syncing
     // (can this be somehow moved to a custom operator or something?)
-    private var syncInProgress: Bool = false
+//    var syncInProgress: Bool = false
```
pawelgil, line 22: this flag and how it's used in the `triggerSynchronization(completion:)` function is **super crucial** for avoiding race conditions, it cannot be removed.

Basically this class should not be modified in order to accomplish the task. I would probably use a [decorator](https://refactoring.guru/pl/design-patterns/decorator) pattern in order to add something like a published bool to indicate if synchronization is in progress. The other way is to ever so slightly change this just by making this `syncInProgress` public and also `@Published` and make our way from there. 
The dependency needs to be the exact oposite as here - `ViewModel` should hold a `SessionSynchronizer`, not the other way around (dependency inversion principle, also single responsibility principle)

AirCasting/SessionsSynchronization/Controller/SessionSynchronizationViewModel.swift
```diff
@@ -0,0 +1,25 @@
+// Created by Lunar on 09/09/2021.
+//
+
+import Foundation
+
+protocol DefaultSessionSynchronizer {
```
pawelgil, line 6: it should have a "ViewModel" in the name. I suggest protocol named `SessionSynchronizationViewModel` and implementation with `Default` prefix

AirCasting/SessionsSynchronization/Controller/SessionSynchronizationController.swift
```diff
@@ -26,22 +28,23 @@ final class SessionSynchronizationController: SessionSynchronizer {
     init(synchronizationContextProvider: SessionSynchronizationContextProvidable,
          downstream: SessionDownstream,
          upstream: SessionUpstream,
-         store: SessionSynchronizationStore) {
+         store: SessionSynchronizationStore,
+         viewModel: SessionSynchronizationViewModel) {
```
pawelgil, line 32: leaving aside the problem that it should not be here in the first place - you should be using a protocol type here, not concrete class (why even create protocols if we inject concrete objects?)

pawelgil (changes requested):

AirCasting/SessionsSynchronization/Controller/SessionSynchronizationController.swift
jankrzempek: My bad, I was doing this and the idea was to use the protocol. Something has happened, I cannot explain I need to say

AirCasting/SessionsSynchronization/Controller/SessionSynchronizationController.swift
```diff
@@ -35,13 +36,17 @@ final class SessionSynchronizationController: SessionSynchronizer {
...
+        if _syncInProgress { return }
+        DispatchQueue.main.async {
+            self._syncInProgress = true
+        }
         
         let onFinish = {
             Log.info("[SYNC] Ending synchronization")
             completion?()
-            self.syncInProgress = false
+            DispatchQueue.main.async {
```
pawelgil, line 47: it cannot be scheduled to main thread due to race condition hazard

AirCasting/SessionsSynchronization/Controller/SessionSynchronizationController.swift
pawelgil: this issue stands

AirCasting/SessionsSynchronization/Controller/SessionSynchronizationViewModel.swift
```diff
@@ -0,0 +1,20 @@
...
+protocol SessionSynchronizationViewModel: ObservableObject {
+    var syncInProgress: Bool { get }
+}
+
+class DefaultSessionSynchronizationViewModel: SessionSynchronizationViewModel {
+    @Published public var syncInProgress: Bool = true
+    private var cancellables = [AnyCancellable]()
+    
+    init(syncSessionController: SessionSynchronizer) {
+        syncSessionController.syncInProgress.sink { [weak self] value in
```
pawelgil, line 16: when you change those DispatchQueue.main.async from controller, lets add `.receive(on: DispatchQueue.main)` (before `.sink`)

pawelgil (changes requested):

pawelgil (approved):

PR #149, opened 2021-08-17
App stopped - mobile session finish

jankrzempek:
What was done:

Handeled AppWillTerminate state, so whenever it crash it saved the mobile active session

https://trello.com/c/IkuofDdr/110-if-the-app-has-stopped-crashed-the-mobile-sessions-should-be-finished-on-app-start-check-if-there-are-any-recording-sessions-and

### DIFF
--- a/AirCasting/AirCastingApp.swift
+++ b/AirCasting/AirCastingApp.swift
@@ -21,9 +21,9 @@ struct AirCastingApp: App {
     private var cancellables: [AnyCancellable] = []
 
     init() {
-        #if !DEBUG
+//        #if !DEBUG
         FirebaseApp.configure()
-        #endif
+//        #endif
         self.authorization = UserAuthenticationSession()
         let synchronizationContextProvider = SessionSynchronizationService(client: URLSession.shared, authorization: authorization, responseValidator: DefaultHTTPResponseValidator())
         let downloadService = SessionDownloadService(client: URLSession.shared, authorization: authorization, responseValidator: DefaultHTTPResponseValidator())
--- a/AirCasting/AirCastingApp.swift
+++ b/AirCasting/AirCastingApp.swift
@@ -8,18 +8,21 @@
 import SwiftUI
 import Firebase
 import Combine
+
 @main
 struct AirCastingApp: App {
     @Environment(\.scenePhase) var scenePhase
     private let authorization: UserAuthenticationSession
     private let syncScheduler: SynchronizationScheduler
     private let microphoneManager: MicrophoneManager
+    private let appState: AppState
     private var sessionSynchronizer: SessionSynchronizer
     private let persistenceController = PersistenceController.shared
     private let appBecameActive = PassthroughSubject<Void, Never>()
     @ObservedObject private var offlineMessageViewModel: OfflineMessageViewModel
     private var cancellables: [AnyCancellable] = []
-
+//    let appState = AppState()
+    
     init() {
 //        #if !DEBUG
         FirebaseApp.configure()
@@ -41,7 +44,7 @@ struct AirCastingApp: App {
                               appBecameActive: appBecameActive.eraseToAnyPublisher(),
                               periodicTimeInterval: 300,
                               authorization: authorization)
-        
+        appState = AppState(microphoneManager: microphoneManager)
         offlineMessageViewModel = .init()
         sessionSynchronizer.errorStream = offlineMessageViewModel
     }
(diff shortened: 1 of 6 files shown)


AirCasting/AirCastingApp.swift
```diff
@@ -8,22 +8,25 @@
...
     private let syncScheduler: SynchronizationScheduler
     private let microphoneManager: MicrophoneManager
+    private let appState: AppState
     private var sessionSynchronizer: SessionSynchronizer
     private let persistenceController = PersistenceController.shared
     private let appBecameActive = PassthroughSubject<Void, Never>()
     @ObservedObject private var offlineMessageViewModel: OfflineMessageViewModel
     private var cancellables: [AnyCancellable] = []
-
+//    let appState = AppState()
```
smialko, line 24: Why did you leave this comment? :)

AirCasting/AirCastingApp.swift
```diff
@@ -8,22 +8,25 @@
...
+    
     init() {
-        #if !DEBUG
+//        #if !DEBUG
         FirebaseApp.configure()
-        #endif
+//        #endif
```
smialko, line 29: Why did you comment `#if !DEBUG` and `#endif`?

AirCasting/AppState.swift
```diff
@@ -0,0 +1,26 @@
+// Created by Lunar on 17/08/2021.
+//
+import SwiftUI
+
+class AppState: ObservableObject {
```
smialko, line 5: it's only for the microphone sessions while mobile sessions can be both mic or AB :)

AirCasting/Settings/MyAccountViewSignOut.swift
```diff
@@ -32,11 +32,12 @@ private extension MyAccountViewSignOut {
...
-                try logoutController.logout()
-            } catch {
-                assertionFailure("Failed to deauthorize \(error)")
-            }
+            fatalError("CRASHLYTICS TEST")
+//            do {
+//                try logoutController.logout()
+//            } catch {
+//                assertionFailure("Failed to deauthorize \(error)")
+//            }
```
smialko, line 40: Same as above - commented code, why? :D

AirCasting/AirCastingApp.swift
```diff
@@ -51,6 +54,7 @@ struct AirCastingApp: App {
             RootAppView(sessionSynchronizer: sessionSynchronizer, persistenceController: persistenceController)
                 .environmentObject(authorization)
                 .environmentObject(microphoneManager)
+                .environmentObject(appState)
```
smialko, line 57: Why making appState `environmentObject` if you don't use it anywhere?

smialko (changes requested):
Good job, although I have some remarks.
The most important:
1. You missed mobile AB sessions. 
2. You don't cover the "On app start check if there are any recording sessions and finish them if so." part of the task.

AirCasting/AirCastingApp.swift
jankrzempek: I cannot see it, maybe from earlier commit? 🙁

AirCasting/AirCastingApp.swift
jankrzempek: same: I cannot see it, maybe from earlier commit? 🙁

AirCasting/AppState.swift
jankrzempek: Yes You are totally right!

AirCasting/Settings/MyAccountViewSignOut.swift
jankrzempek: same: I cannot see it, maybe from earlier commit? 🙁

AirCasting/AirCastingApp.swift
jankrzempek: same: I cannot see it, maybe from earlier commit? 🙁

AirCasting/AirCastingApp.swift
smialko: it's weird, why github is marking it as changes if it's different now. Sorry, I can see in the code that it's not commented anymore 🤷🏻‍♀️

AirCasting/Settings/MyAccountViewSignOut.swift
smialko: Weird, but true, the code is different from what github is suggesting it is.

AirCasting/AirCastingApp.swift
jankrzempek: Important that we agreed on this 😛

PR #151, opened 2021-08-17
Handle app crash

jankrzempek:
App stopped - check if session finished
What was done:

Handeled AppWillTerminate state, so whenever it crash it saved the mobile active session

https://trello.com/c/IkuofDdr/110-if-the-app-has-stopped-crashed-the-mobile-sessions-should-be-finished-on-app-start-check-if-there-are-any-recording-sessions-and


pawelgil (changes requested):
Not doing CR since I think this code is redundant. The `PersistenceController` is already finishing all mobile sessions at it's startup phase (its not great design to put such important logic as side effects in obscure places like that, but it does the job):
```Swift
init(inMemory: Bool = false) {
    ...
    finishMobileSessions()
}
```

Where it fetches all ongoing mobile sessions and stopping them. Can you confirm @jankrzempek @smialko ?

jankrzempek:
> Not doing CR since I think this code is redundant. The `PersistenceController` is already finishing all mobile sessions at it's startup phase (its not great design to put such important logic as side effects in obscure places like that, but it does the job):
> 
> ```swift
> init(inMemory: Bool = false) {
>     ...
>     finishMobileSessions()
> }
> ```
> 
> Where it fetches all ongoing mobile sessions and stopping them. Can you confirm @jankrzempek @smialko

PR #157, opened 2021-08-18
Wizard text formatting

jankrzempek:
https://trello.com/c/0Ii0jCDy/223-fixed-mobile-wizard-start-recording-screen-update-text-formatting

PR #158, opened 2021-08-18
Map button - outdoor

jankrzempek:
In addition to this: https://github.com/HabitatMap/AirCastingiOS/pull/157 

After those two tickets isIndoor will be handled in every case, so that in this ticket we are able to shorten IF

https://trello.com/c/jp738Egu/221-fixed-outdoor-dashboard-map-button-is-missing


smialko (approved):
Screenshot 2021-08-18 at 08 59 28

PR #164, opened 2021-08-19
changed map to graph

jankrzempek:
There were two maps button instead of graph and map -> now it should be ok

### DIFF
--- a/AirCasting/SessionViews/SessionCartView.swift
+++ b/AirCasting/SessionViews/SessionCartView.swift
@@ -101,6 +101,8 @@ private extension SessionCartView {
             isMapButtonActive = true
         } label: {
             Text(Strings.SessionCartView.map)
+                .font(Font.muli(size: 13, weight: .semibold))
+                .padding(.horizontal, 8)
         }
     }
     
@@ -108,7 +110,7 @@ private extension SessionCartView {
         Button {
             isMapButtonActive = true
         } label: {
-            Text(Strings.SessionCartView.map)
+            Text(Strings.SessionCartView.graph)
                 .font(Font.muli(size: 13, weight: .semibold))
                 .padding(.horizontal, 8)
         }
--- a/AirCasting/SessionViews/SessionCartView.swift
+++ b/AirCasting/SessionViews/SessionCartView.swift
@@ -100,7 +100,7 @@ private extension SessionCartView {
         Button {
             isGraphButtonActive = true
         } label: {
-            Text(Strings.SessionCartView.map)
+            Text(Strings.SessionCartView.graph)
                 .font(Font.muli(size: 13, weight: .semibold))
                 .padding(.horizontal, 8)
         }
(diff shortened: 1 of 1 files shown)


AirCasting/SessionViews/SessionCartView.swift
```diff
@@ -101,14 +101,16 @@ private extension SessionCartView {
             isMapButtonActive = true
```
smialko, line 101: this shouldn't be `isMapButtonActive` but `isGraphButtonActive`. 
It's rebase's fault. I hate rebases.

smialko (changes requested):

AirCasting/SessionViews/SessionCartView.swift
jankrzempek: Feel You, Hate rebases too. Fixing now!

pawelgil: <grumpy 👴🏻 Paweł>
The conflicts emerging in each PR are because we're shoving everything in the same files (mainly views are our god objects, but not only). We need to split responsibilities and make our architecture obey clean code principles sooner than later if we don't want to push our lives into a misery ;)
</grumpy 👴🏻 Paweł>

pawelgil (approved):

AirCasting/SessionViews/SessionCartView.swift
jankrzempek: I love your comments and I know it is serious one. I think that as soon as after MVP release we can focus on that mostly 😄

smialko: Could we make a deal that after we're done with MVP (essential) we'll spend a day on a call refactoring code structure? I'd love to learn more about code structure from someone experienced :D

AirCasting/SessionViews/SessionCartView.swift
```diff
@@ -98,17 +98,19 @@ private extension SessionCartView {
...
+                .padding(.horizontal, 8)
         }
     }
     
     var mapButton: some View {
         Button {
             isMapButtonActive = true
         } label: {
-            Text(Strings.SessionCartView.map)
+            Text(Strings.SessionCartView.graph)
```
smialko, line 113: map button should have` Text(Strings.SessionCartView.map)` 😅

AirCasting/SessionViews/SessionCartView.swift
```diff
@@ -98,17 +98,19 @@ private extension SessionCartView {
     
     var graphButton: some View {
         Button {
-            isMapButtonActive = true
+            isGraphButtonActive = true
         } label: {
             Text(Strings.SessionCartView.map)
```
smialko, line 103: Graph button should have `Text(Strings.SessionCartView.graph)` 🙃

smialko (changes requested):
🙃

smialko (approved):
🥷🏻

PR #166, opened 2021-08-19
differentiate views for sessions

jankrzempek:
https://trello.com/c/52VOtYg4/142-change-loading-view-for-mic-sessions

What was done:

The goal is to differentiate different session and show different views

### DIFF
--- a/AirCasting/MicrophoneSession/MicrophoneManager.swift
+++ b/AirCasting/MicrophoneSession/MicrophoneManager.swift
@@ -110,7 +110,7 @@ private extension MicrophoneManager {
 
     func createMeasurementStream(for session: Session) throws -> MeasurementStreamLocalID {
         let stream = MeasurementStream(id: nil,
-                                       sensorName: "Phone Microphone",
+                                       sensorName: Strings.SensorsData.microphone,
                                        sensorPackageName: "Builtin",
                                        measurementType: "Sound Level",
                                        measurementShortType: "db",
--- a/AirCasting/MicrophoneSession/MicrophoneManager.swift
+++ b/AirCasting/MicrophoneSession/MicrophoneManager.swift
@@ -110,7 +110,7 @@ private extension MicrophoneManager {
 
     func createMeasurementStream(for session: Session) throws -> MeasurementStreamLocalID {
         let stream = MeasurementStream(id: nil,
-                                       sensorName: Strings.SensorsData.microphone,
+                                       sensorName: Constants.SensorName.microphone,
                                        sensorPackageName: "Builtin",
                                        measurementType: "Sound Level",
                                        measurementShortType: "db",
--- a/AirCasting/SessionViews/ABMeasurementsView.swift
+++ b/AirCasting/SessionViews/ABMeasurementsView.swift
@@ -45,27 +45,25 @@ struct ABMeasurementsView: View {
                         .frame(maxWidth: .infinity)
                     }
                 }
-            } else if session.type == .fixed {
-               SessionLoadingView()
-            } else if session.allStreams?.count == 1 {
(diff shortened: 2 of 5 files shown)


AirCasting/SessionViews/ABMeasurementsView.swift
```diff
@@ -95,12 +105,18 @@ struct SingleMeasurementView: View {
     
     func showStreamName() -> String {
         guard let streamName = stream.sensorName else { return "" }
-        return streamName
-            .drop { $0 != "-" }
-            .replacingOccurrences(of: "-", with: "")
+        if streamName == "Phone Microphone" {
```
pawelgil, line 108: this string needs to go to a constant. Reason being - it's used in two completely different places:
* Here
* `MicrophoneManager` - when creating mic sessions

Now imagine a task that would say "we need to rename mic session stream names from "Phone Microphone" to "Microphone". Someone performing this change would just go to `MicrophoneManager` and change the string there accidentally braking this piece of logic and probably wouldn't spot it because it's not easy to spot.

If we introduce a constant for this on the other hand, the developer would go there instead and change this constants value. Now both places would get the change and no bugs would be introduced by accident.

pawelgil (changes requested):

smialko (approved):
:D

AirCasting/MicrophoneSession/MicrophoneManager.swift
```diff
@@ -110,7 +110,7 @@ private extension MicrophoneManager {
 
     func createMeasurementStream(for session: Session) throws -> MeasurementStreamLocalID {
         let stream = MeasurementStream(id: nil,
-                                       sensorName: "Phone Microphone",
+                                       sensorName: Strings.SensorsData.microphone,
```
smialko, line 113: I'm not sure but shouldn't `Strings` contain only text we display for users? 
Maybe we could make a constant for this sensor name?

AirCasting/SessionViews/ABMeasurementsView.swift
```diff
@@ -41,32 +41,39 @@ struct ABMeasurementsView: View {
...
-                    VStack(alignment: .leading, spacing: 3) {
-                        Text(Strings.LoadingSession.title)
-                            .font(Font.moderate(size: 14))
-                        Text(Strings.LoadingSession.description)
-                            .font(Font.moderate(size: 12))
+                if session.followedAt != nil {
+                    SessionLoadingView()
+                } else {
+                    VStack(alignment: .leading, spacing: 8) {
+                        Text("Parameters:")
```
smialko, line 53: We could move `"Parameters:"` to Strings :)

smialko (approved):
👏

AirCasting/MicrophoneSession/MicrophoneManager.swift
jankrzempek: It needs to be accessible from everywhere, maybe let's just name String struct well, what do You think?

pawelgil: I agree with @smialko . We need to introduce a completely new namespace for this. Something like:
`Constants.SensorNames.microphone` or something

pawelgil (changes requested):

AirCasting/Utils/Constants.swift
```diff
@@ -0,0 +1,10 @@
+// Created by Lunar on 20/08/2021.
+//
+
+import Foundation
+
+struct Constants {
```
smialko, line 6: Love this 💙🚀

smialko (approved):
Nice job!

pawelgil (approved):

PR #167, opened 2021-08-20
changes to action menu

jankrzempek:
Removed for now unused button from session menu

### DIFF
--- a/AirCasting/CoreData/SessionEntity.swift
+++ b/AirCasting/CoreData/SessionEntity.swift
@@ -132,9 +132,3 @@ extension SessionEntity {
     @NSManaged public func removeFromMeasurementStreams(_ values: NSSet)
 
 }
-
-extension SessionEntity {
-       var isActive: Bool {
-           type == .mobile && status == .RECORDING
-       }
-}
--- a/AirCasting/CoreData/SessionEntity.swift
+++ b/AirCasting/CoreData/SessionEntity.swift
@@ -132,3 +132,9 @@ extension SessionEntity {
     @NSManaged public func removeFromMeasurementStreams(_ values: NSSet)
 
 }
+
+extension SessionEntity {
+       var isActive: Bool {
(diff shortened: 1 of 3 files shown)


smialko (changes requested):
Fixed session shouldn'y have "finish recording" as it can't be finished. We can remove this action button from fixed session for now :)

AirCasting/SessionViews/SessionHeaderView.swift
```diff
@@ -30,6 +30,7 @@ struct SessionHeaderView: View {
                 Spacer()
                 if session.type == .fixed {
                     actionsMenuFixed
+                        .disabled(true)
```
smialko, line 33: maybe it would be better to remove this actionsMenuFixed at all for the time being?

smialko (changes requested):
🙈

AirCasting/SessionViews/SessionHeaderView.swift
jankrzempek: I really like this idea - done

AirCasting/CoreData/SessionEntity.swift
```diff
@@ -132,3 +132,9 @@ extension SessionEntity {
     @NSManaged public func removeFromMeasurementStreams(_ values: NSSet)
 
 }
+
+extension SessionEntity {
```
smialko, line 136: wait, we already have such extension with different session types

smialko: Screenshot 2021-08-20 at 14 48 10

smialko: could you move isActive there?

jankrzempek: for sure, it is a bit mixed 😅

smialko (approved):

PR #170, opened 2021-08-20
Fix downloading measurements for fixed session

smialko:


jankrzempek (approved):
🥳

smialko (commented):
@pawelgil  I'm merging it beacuse without it fixed sessions don't fetch measurement. Please review it anyway and I'll add fixed from your CR in separate branch

PR #171, opened 2021-08-23
ContributToCrowdMap fix 

jankrzempek:
What was done:

https://trello.com/c/XkbtqPbo/116-settings-contribute-to-crowd-map

Now the default state of the toggle is "ON"
UI - text is now well laid
Ensured that contriuteToCrowdMap changes all the time and the proper state is sent during the session creation process

### DIFF
--- a/AirCasting/AuthViews/UserSettings.swift
+++ b/AirCasting/AuthViews/UserSettings.swift
@@ -6,15 +6,18 @@ import Foundation
 class UserSettings: ObservableObject {
     private let userDefaults: UserDefaults
     
-    @Published var contributingToCrowdMap: Bool {
-        didSet {
-            Log.info("Changed crowdmap contribution setting to \(contributingToCrowdMap ? "ON" : "OFF")")
-            userDefaults.setValue(contributingToCrowdMap, forKey: "crowdMap")
+    var contributingToCrowdMap: Bool {
+        get {
+            userDefaults.bool(forKey: "crowdMap")
+        }
+        set {
+            userDefaults.setValue(newValue, forKey: "crowdMap")
+            Log.info("Changed crowdMap contribution setting to \(contributingToCrowdMap ? "ON" : "OFF")")
         }
     }
     
     init(userDefaults: UserDefaults = .standard) {
-        self.contributingToCrowdMap = userDefaults.bool(forKey: "crowdMap")
         self.userDefaults = userDefaults
+        contributingToCrowdMap = true
     }
 }
--- a/AirCasting/CreateSessionViews/CreateSessionDetailsView.swift
+++ b/AirCasting/CreateSessionViews/CreateSessionDetailsView.swift
@@ -161,21 +161,19 @@ private extension CreateSessionDetailsView {
         if sessionContext.sessionType == .fixed {
             if isIndoor {
                 sessionContext.startingLocation = fakeLocation
-                locationTracker.googleLocation = [PathPoint(location: CLLocationCoordinate2D(latitude: 200.0, longitude: 200.0), measurement: 20.0)]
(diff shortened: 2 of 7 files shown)


AirCasting/AuthViews/UserSettings.swift
```diff
@@ -6,15 +6,18 @@ import Foundation
 class UserSettings: ObservableObject {
     private let userDefaults: UserDefaults
     
-    @Published var contributingToCrowdMap: Bool {
-        didSet {
-            Log.info("Changed crowdmap contribution setting to \(contributingToCrowdMap ? "ON" : "OFF")")
-            userDefaults.setValue(contributingToCrowdMap, forKey: "crowdMap")
+    var contributingToCrowdMap: Bool {
+        get {
+            userDefaults.bool(forKey: "crowdMap")
```
smialko, line 11: Why not use @AppStorage here? :)

AirCasting/Map/LocationTracker.swift
```diff
@@ -21,7 +21,11 @@ class LocationTracker: NSObject, ObservableObject, CLLocationManagerDelegate {
         switch locationManager.authorizationStatus {
             case .authorizedAlways, .authorizedWhenInUse:
                 self.locationGranted = .granted
-                googleLocation = [PathPoint(location: CLLocationCoordinate2D(latitude: (locationManager.location?.coordinate.latitude)!, longitude: (locationManager.location?.coordinate.longitude)!), measurement: 20)]
+                if locationManager.location?.coordinate.latitude != nil && locationManager.location?.coordinate.longitude != nil {
+                    googleLocation = [PathPoint(location: CLLocationCoordinate2D(latitude: (locationManager.location?.coordinate.latitude)!, longitude: (locationManager.location?.coordinate.longitude)!), measurement: 20)]
```
smialko, line 25: How about using a variable here for mocked measurement? :)

smialko (approved):
🥷🏻

AirCasting/AuthViews/UserSettings.swift
jankrzempek: We have (I think) about that with Paweł that then it is somehow directly access to UserDefeults and now not exactly but I am not sure and I am learning so don't listen to me. We need to ask @pawelgil

AirCasting/Map/LocationTracker.swift
jankrzempek: Greeeeet. Do we have some or I should create that?

AirCasting/CreateSessionViews/CreateSessionDetailsView.swift
```diff
@@ -161,21 +161,19 @@ private extension CreateSessionDetailsView {
...
             if isIndoor {
                 sessionContext.startingLocation = fakeLocation
-                locationTracker.googleLocation = [PathPoint(location: CLLocationCoordinate2D(latitude: 200.0, longitude: 200.0), measurement: 20.0)]
-                // measurement: 20.0 was designed just to be 'something'. Is should be handle somehow, but for now we are leaving this like it is.
+                locationTracker.googleLocation = [PathPoint.fakePathPoint]
             } else {
                 guard let lat = (locationTracker.locationManager.location?.coordinate.latitude),
                       let lon = (locationTracker.locationManager.location?.coordinate.longitude) else { return }
-                locationTracker.googleLocation = [PathPoint(location: CLLocationCoordinate2D(latitude: lat, longitude: lon), measurement: 20.0)]
+                locationTracker.googleLocation = [PathPoint.fakePathPoint]
```
pawelgil, line 168: btw. I have a strong feeling this googleLocation doesn't need to b PathPoint (it can be just a standard CoreLocation coordinates). But we need to confirm.

pawelgil (approved):

PR #172, opened 2021-08-23
wifi to the left

jankrzempek:
https://trello.com/c/ZRoqJXu3/244-fixed-wizard-move-wifi-to-the-left-make-it-the-default-selection

What was done:
Although the wifi is picked now as the default we are somehow ensure that user type the password because the wifi warning popup will appear 

Changes Wifi and cellular


smialko (approved):
😎

PR #174, opened 2021-08-23
implemented heatmap deault action

jankrzempek:
https://trello.com/c/V8vkUMOU/237-heat-legend-reset-to-default-does-nothing


smialko (changes requested):
It's not how it should work. 
Those thresholds are just mic thresholds. 
Here we need thresholds we get from AB. That's why we have global thresholds for each stream but we also save thresholds we get from AB in the session entity. We have to fetch those thresholds from any AB session :)

smialko:
I'm closing this PR as all this code needs to be removed and done differently anyways, and Janek is off - it makes no sense for me to take over this PR and remake all of this :)

PR #175, opened 2021-08-23
empty onboarding for fixed session added

jankrzempek:
What was done:

Empty onabording should be different for (fixed and following) and (mobile dormant and active)

https://trello.com/c/ADhWO5Uc/243-following-blank-fixed-new-profile-default-view-should-look-like-android


AirCasting/Dashboard/EmptyDashboardViewFixed.swift
```diff
@@ -0,0 +1,57 @@
+// Created by Lunar on 23/08/2021.
+//
+import AirCastingStyling
+import SwiftUI
+
+struct EmptyDashboardViewFixed: View {
```
smialko, line 6: Both `EmptyDashboardViewFixed` and `EmptyDashboardViewMobile` have super crowded body. 
Some suggestions for the future (I'm refactoring this code right now, bc you're vacationing :D) 

1. Don't use VStack with just one element inside, it just doesn't make sense. Also adding "spacing" such VStack does nothing - the spacing is set for elements **inside** the VStack, when it has only one view inside it can't add spacing, bc spacing between what? :D 
2. When you create a variable for a view outside of the body you can add characteristics like `frame` in the variable declaration, too. Keeps the body shorter. Example: "old" `airBeamDescription`
3. You have a button that is used at least in two separate views. Why not extract instead of doubling the code? :D 
4. When you have VStack of two Texts and some of the characteristics are doubled you can move those characteristics to outside the VStack :)

smialko (commented):
It'll be refactored so no need for CR for now @pawelgil :)

smialko:
@pawelgil you can review it now 😎

AirCasting/Dashboard/EmptyDashboardViewFixed.swift
smialko: @jankrzempek I'm tagging you so you won't miss this comment 🦆

smialko (approved):
I'm approving but it's mine so it doesn't really matter 🥷🏻

PR #176, opened 2021-08-23
wizard confirm screen changed

jankrzempek:
https://trello.com/c/QjomS94U/242-wizard-mobile-fixed-outdoor-fixed-indoor-are-you-ready-screen-text-flow-is-incorrect

### DIFF
--- a/AirCasting/CreateSessionViews/ConfirmCreatingSessionView.swift
+++ b/AirCasting/CreateSessionViews/ConfirmCreatingSessionView.swift
@@ -88,7 +88,7 @@ struct ConfirmCreatingSessionView: View {
             ZStack {
                 if sessionContext.sessionType == .mobile {
                     GoogleMapView(pathPoints: [], isMyLocationEnabled: true)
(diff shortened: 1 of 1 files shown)


AirCasting/CreateSessionViews/ConfirmCreatingSessionView.swift
```diff
@@ -93,7 +88,7 @@ struct ConfirmCreatingSessionView: View {
             ZStack {
                 if sessionContext.sessionType == .mobile {
                     GoogleMapView(pathPoints: [], isMyLocationEnabled: true)
-                } else {
+                } else if !sessionContext.isIndoor! {
```
pawelgil, line 91: IDK about the force unwrap here. It will crash the value if not present... Can we add a fallback instead?
```Swift
} else if !(sessionContext.isIndoor ?? false) {
```

pawelgil (changes requested):

smialko (approved):

pawelgil (approved):

PR #178, opened 2021-08-23
changed text depends on the state - sessions

jankrzempek:
What was done:
Michael in comments said the we need to make some changes - I know that iffffying is not pretty but don't have better ideas for know :/

https://trello.com/c/zvSTCjIu/45-differences-design-android-app-vs-ios-app

### DIFF
--- a/AirCasting/Graph/GraphView.swift
+++ b/AirCasting/Graph/GraphView.swift
@@ -21,6 +21,7 @@ struct GraphView: View {
                               session: session,
                               sessionStopperFactory: sessionStoppableFactory).padding()
             StreamsView(selectedStream: $selectedStream,
+                        isCollapsed: Binding.constant(true),
                         session: session,
                         thresholds: thresholds,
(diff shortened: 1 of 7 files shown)


pawelgil:
Basically any if/else can be replaced by some kind of polymorphism or other design solutions. Some reading on the topic:
[1](https://blog.jetbrains.com/idea/2017/09/code-smells-if-statements/)
[2](https://levelup.gitconnected.com/conditional-statements-are-a-code-smell-here-is-your-relief-38e50c023708)
[3](https://levelup.gitconnected.com/treat-if-else-as-a-code-smell-until-proven-otherwise-3bd2c4c577bf)

Here I would *strongly* suggest moving this string resolution logic into a viewmodel (along all the other view logic) and then making further refactorings. For now tho I think we can roll with those `if`s since it's not extremely unreadable

pawelgil (approved):

smialko (approved):

PR #184, opened 2021-08-26
Fix thresholds

smialko:
The bug: when the user changed the threshold on the map or graph view using the slider at the bottom, the color didn't update on the measurement view (the border and dot). 
Fixed by making threshold ObservedObject. 
I had to make an additional struct to make the threshold a non-optional variable to be able to make it observed. I nested this struct into `SingleMeasurementView`. 

finally got something done 🥳


AirCasting/SessionViews/ABMeasurementsView.swift
```diff
@@ -107,56 +107,69 @@ struct SingleMeasurementView: View {
...
-        case threshold.thresholdLow..<threshold.thresholdMedium:
-            return .aircastingYellow
-        case threshold.thresholdMedium..<threshold.thresholdHigh:
-            return .aircastingOrange
-        case threshold.thresholdHigh..<threshold.thresholdVeryHigh:
-            return .aircastingRed
-        default:
-            return .white
+
+    struct _SingleMeasurementButton: View {
```
pawelgil, line 135: not sure why there is an underscore in the name of this struct. It also might be beneficial to put this in a separate file.

pawelgil (approved):

AirCasting/SessionViews/ABMeasurementsView.swift
smialko: Not sure if you want to move `SingleMeasurementView` or `_SingleMeasurementButton` to separate file. I think it makes more sense to move the whole `SingleMeasurementView`. I can also make `_SingleMeasurementButton` a separate struct (it's nested in `SingleMeasurementView`). 
And the underscore. It's just a mark that this struct is private. Some apple frameworks have the underscore in names to mark is as private so it was kind of tribute to them.🙃 But I'll remove it if it's confusing.

PR #196, opened 2021-09-07
Change dashboard texts for fixed followed

smialko:
https://trello.com/c/ADhWO5Uc/243-following-blank-fixed-new-profile-default-view-should-look-like-android


smialko:
It's **super super small** so I'm merging it without approves. I want to send it within today's build 😬

PR #200, opened 2021-09-08
added to string dot and cleanup

jankrzempek:
https://trello.com/c/QjomS94U/242-wizard-mobile-fixed-outdoor-fixed-indoor-are-you-ready-screen-text-flow-is-incorrect

What was done:

Text was doubled because of some unwanted var so it is deleted now.


smialko (approved):
🥳

PR #202, opened 2021-09-08
Map appearing fix

jankrzempek:
https://trello.com/c/7AibKytL/259-wizard-fixed-outdoor-search-the-address-are-you-ready-map-is-missing

Map was not appearing on fixed session, now it is working.

What was done:
Fake location was replaced by the right location and the map shows well now. On fixed session, location which was sent to database was the always the starting location. For now it should be fixed and after moving map location which will be send is changed too.

### DIFF
--- a/AirCasting/CreateSessionViews/ConfirmCreatingSessionView.swift
+++ b/AirCasting/CreateSessionViews/ConfirmCreatingSessionView.swift
@@ -16,6 +16,7 @@ struct ConfirmCreatingSessionView: View {
             isPresentingAlert = error != nil
         }
     }
+
     @State private var isPresentingAlert: Bool = false
     @State private var didStartRecordingSession = false
     @EnvironmentObject var selectedSection: SelectSection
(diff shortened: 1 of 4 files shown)


AirCasting/CreateSessionViews/ConfirmCreatingSessionView.swift
```diff
@@ -118,7 +123,26 @@ struct ConfirmCreatingSessionView: View {
             })
                 .buttonStyle(BlueButtonStyle())
         }
-            .padding()
+        .padding()
+    }
+
+    func getAndSaveStartingLocation() {
+        if sessionContext.sessionType == .fixed {
```
pawelgil, line 130: let's split this block into some more private functions for readability

edit: I see it's just been moved from somewhere, disregard ;)

pawelgil (approved):
would be better to only use xcodes reformatting on parts where you actually edit the code, not whole files. This will result in fewer conflicts

PR #208, opened 2021-09-09
Add dot to mobile session's map

smialko:
Michael asked to make the polyline thinner & and a dot at the end of the polyline. I decided to use GMSMarker instead of GMSCircle because GMSCircle doesn't change its width when scaling the map so it's invisible after zoom out.
Task: https://trello.com/c/jv3Hk6EG/227-mobile-map-improve-mapping-visualization


jankrzempek (approved):
Cool! 👌🏻

PR #209, opened 2021-09-09
time 24h format

jankrzempek:
https://trello.com/c/7ktdjPxZ/260-use-24-hr-time

What was done:

System has the settings which enforces all apps to use 24 or 12 format. For now we should use only 24h format so the conversion needs to be made - and that was the idea of this task.

TO TEST:
Go to system settings and change format to 12 seeing if the time on the session cart changes or not

### DIFF
--- a/AirCasting/SessionViews/SessionHeaderView.swift
+++ b/AirCasting/SessionViews/SessionHeaderView.swift
@@ -53,6 +53,7 @@ private extension SessionHeaderView {
         formatter.timeStyle = .short
         formatter.dateStyle = .medium
         var date = ""
+        var additionalDate = ""
         
             guard let start = session.startTime else { return Text("") }
             let end = session.endTime ?? Date()
@@ -64,7 +65,14 @@ private extension SessionHeaderView {
         } else {
             if string.contains("–") {
                 let time = string.components(separatedBy: "–")
-                let endTime12 = time.last?.trimmingCharacters(in: .whitespaces)
+                var timeLast = time.last
+                // some of the dates on fixed session are different we need to handle that 'out of nowhere (few days sessions)' case
+                if ((time.last?.contains("/")) != nil) {
+                    additionalDate = (time.last?.components(separatedBy: ",").first)!
+                    timeLast = time.last?.components(separatedBy: ",").last
+                }
+                
+                let endTime12 = timeLast!.trimmingCharacters(in: .whitespaces)
             
                 let last = time.first!
                 let time2 = last.components(separatedBy: ",")
@@ -72,10 +80,13 @@ private extension SessionHeaderView {
                 var startTime12 = time2.last?.trimmingCharacters(in: .whitespaces)
                 
                 if !(startTime12!.contains("PM") || startTime12!.contains("AM")) {
-                    (endTime12?.contains("PM"))! ? startTime12?.append(" PM") : startTime12?.append(" AM")
+                    (endTime12.contains("PM")) ? startTime12?.append(" PM") : startTime12?.append(" AM")
                 }
-
-                date.append(", \(timeConversion24(time12: startTime12!))-\(timeConversion24(time12: endTime12!))")
+                if additionalDate == "" {
+                    date.append(", \(timeConversion24(time12: startTime12!))-\(timeConversion24(time12: endTime12))")
+                } else {
+                    date.append(", \(timeConversion24(time12: startTime12!))-\(additionalDate), \(timeConversion24(time12: endTime12))")
+                }
             } else {
                 let time2 = string.components(separatedBy: ",")
                 date = time2.first!
--- a/AirCasting/SessionViews/SessionHeaderView.swift
+++ b/AirCasting/SessionViews/SessionHeaderView.swift
@@ -48,54 +48,7 @@ struct SessionHeaderView: View {
 
 private extension SessionHeaderView {
     var dateAndTime: some View {
-        
-        let formatter = DateIntervalFormatter()
-        formatter.timeStyle = .short
-        formatter.dateStyle = .medium
-        var date = ""
-        var additionalDate = ""
-        
-            guard let start = session.startTime else { return Text("") }
-            let end = session.endTime ?? Date()
-            let string = DateIntervalFormatter().string(from: start, to: end)
-        
-        if is24Hour() {
-            let replaced = string.replacingOccurrences(of: "—", with: "-")
-            return Text(replaced)
-        } else {
-            if string.contains("–") {
-                let time = string.components(separatedBy: "–")
-                var timeLast = time.last
-                // some of the dates on fixed session are different we need to handle that 'out of nowhere (few days sessions)' case
-                if ((time.last?.contains("/")) != nil) {
-                    additionalDate = (time.last?.components(separatedBy: ",").first)!
-                    timeLast = time.last?.components(separatedBy: ",").last
-                }
-                
-                let endTime12 = timeLast!.trimmingCharacters(in: .whitespaces)
-            
-                let last = time.first!
-                let time2 = last.components(separatedBy: ",")
-                date = time2.first!
-                var startTime12 = time2.last?.trimmingCharacters(in: .whitespaces)
-                
-                if !(startTime12!.contains("PM") || startTime12!.contains("AM")) {
-                    (endTime12.contains("PM")) ? startTime12?.append(" PM") : startTime12?.append(" AM")
-                }
-                if additionalDate == "" {
-                    date.append(", \(timeConversion24(time12: startTime12!))-\(timeConversion24(time12: endTime12))")
-                } else {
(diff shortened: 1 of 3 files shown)


AirCasting/SessionViews/SessionHeaderView.swift
```diff
@@ -48,18 +48,54 @@ struct SessionHeaderView: View {
...
         
         let formatter = DateIntervalFormatter()
-        
         formatter.timeStyle = .short
         formatter.dateStyle = .medium
-        let string = DateIntervalFormatter().string(from: start, to: end)
-        let replaced = string.replacingOccurrences(of: "—", with: "-")
-        return Text(replaced)
+        var date = ""
+        var additionalDate = ""
```
smialko, line 56: what's `additionalDate`? This name is kinda confusing :)

AirCasting/Utils/TimeConversion.swift
```diff
@@ -0,0 +1,22 @@
+// Created by Lunar on 09/09/2021.
+//
+
+import Foundation
+
```
smialko, line 5: I think we shouldn't make those functions global but correct me if I'm wrong @pawelgil :) 
You can wrap those funcs into a TimeConverter class f.e. :)

AirCasting/SessionViews/SessionHeaderView.swift
```diff
@@ -48,18 +48,54 @@ struct SessionHeaderView: View {
...
+            let string = DateIntervalFormatter().string(from: start, to: end)
+        
+        if is24Hour() {
+            let replaced = string.replacingOccurrences(of: "—", with: "-")
+            return Text(replaced)
+        } else {
+            if string.contains("–") {
+                let time = string.components(separatedBy: "–")
+                var timeLast = time.last
+                // some of the dates on fixed session are different we need to handle that 'out of nowhere (few days sessions)' case
```
smialko, line 69: what do you mean by "'out of nowhere (few days sessions)'"?

AirCasting/SessionViews/SessionHeaderView.swift
```diff
@@ -48,18 +48,54 @@ struct SessionHeaderView: View {
 
 private extension SessionHeaderView {
     var dateAndTime: some View {
```
smialko, line 50: this variable is super big& complicated! Maybe we could extract it somewhere as a function?

AirCasting/SessionViews/SessionHeaderView.swift
```diff
@@ -48,18 +48,54 @@ struct SessionHeaderView: View {
...
-        let end = session.endTime ?? Date()
         
         let formatter = DateIntervalFormatter()
-        
         formatter.timeStyle = .short
         formatter.dateStyle = .medium
-        let string = DateIntervalFormatter().string(from: start, to: end)
-        let replaced = string.replacingOccurrences(of: "—", with: "-")
-        return Text(replaced)
+        var date = ""
```
smialko, line 55: this name could be a bit more explanatory, is it a start date, end date of the session? :D

smialko (commented):
Well, you were right, it's definitely more complex than anyone expected! 
I'm not sure about those force unwrapping here and there. You sure it's safe?

AirCasting/SessionViews/SessionHeaderView.swift
```diff
@@ -48,18 +48,54 @@ struct SessionHeaderView: View {
...
+        var additionalDate = ""
+        
+            guard let start = session.startTime else { return Text("") }
+            let end = session.endTime ?? Date()
+            let string = DateIntervalFormatter().string(from: start, to: end)
```
pawelgil, line 60: I know you're all tired of me saying this, but we should split responsibilities and this should be handled by the viewmodel. Common interpretation for "single responsibility" is that an object should only have one reason to change. For the case of views it should solely be UI layout. Not formatting, not colors, not strings, just a screen layout.

As for now I think we should just roll with it as deadline is approaching, but in the future please keep that in mind and introduce proper viewmodels so that any data formatting decisions are being made there, not in the view itself.

AirCasting/SessionViews/SessionHeaderView.swift
```diff
@@ -48,18 +48,54 @@ struct SessionHeaderView: View {
...
+        
+        if is24Hour() {
+            let replaced = string.replacingOccurrences(of: "—", with: "-")
```
pawelgil, line 63: I'm not too sure this line is doing anything? It's replacing occurrences of "-" with "-" ?

AirCasting/SessionViews/SessionHeaderView.swift
pawelgil, line 69: unclear comment, still no idea what the purpose is :) can be more explicit

AirCasting/SessionViews/SessionHeaderView.swift
```diff
@@ -48,18 +48,54 @@ struct SessionHeaderView: View {
...
+                var startTime12 = time2.last?.trimmingCharacters(in: .whitespaces)
+                
+                if !(startTime12!.contains("PM") || startTime12!.contains("AM")) {
+                    (endTime12.contains("PM")) ? startTime12?.append(" PM") : startTime12?.append(" AM")
+                }
+                if additionalDate == "" {
+                    date.append(", \(timeConversion24(time12: startTime12!))-\(timeConversion24(time12: endTime12))")
+                } else {
+                    date.append(", \(timeConversion24(time12: startTime12!))-\(additionalDate), \(timeConversion24(time12: endTime12))")
+                }
```
pawelgil, line 89: i feel like this whole block of code can be handled by a well formatted `DateFormatter`. Not sure tho as it's pretty enigmatic in what it does. If it needs to stay like this - please comment it heavily with examples so that anyone that stumbles upon it will have an easier time understanding what it does.

Example:
```Swift
// Date format provided can be either "2021/09/09 10:20PM" or "2021-09-09 22:20" so we
// need to distinguish between those two formats (etc. etc)...
```

AirCasting/Utils/TimeConversion.swift
pawelgil: indeed. Please consider any global functions or variables as a no-go

AirCasting/Utils/TimeConversion.swift
```diff
@@ -0,0 +1,22 @@
...
+func is24Hour() -> Bool {
+    let dateFormat = DateFormatter.dateFormat(fromTemplate: "j", options: 0, locale: Locale.current)!
+    return dateFormat.firstIndex(of: "a") == nil
+}
+
+func timeConversion24(time12: String) -> String {
+    let dateAsString = time12
+    let df = DateFormatter()
```
pawelgil, line 13: initialization time of a `DateFormatter` is a really time consuming task, we probably should be defining this in a static (if you make this an object it will be possible)

AirCasting/Utils/TimeConversion.swift
```diff
@@ -0,0 +1,22 @@
+// Created by Lunar on 09/09/2021.
+//
+
+import Foundation
+
+func is24Hour() -> Bool {
+    let dateFormat = DateFormatter.dateFormat(fromTemplate: "j", options: 0, locale: Locale.current)!
```
pawelgil, line 7: initialization time of a `DateFormatter` is a really time consuming task, we probably should be defining this in a static (if you make this an object it will be possible)

AirCasting/SessionViews/SessionHeaderView.swift
```diff
@@ -95,20 +131,20 @@ private extension SessionHeaderView {
             }
         }.alert(isPresented: $showingFinishAlert) {
             Alert(title: Text(Strings.SessionHeaderView.finishAlertTitle) +
-                Text(session.name ?? Strings.SessionHeaderView.finishAlertTitle_2)
-                +
-                Text(Strings.SessionHeaderView.finishAlertTitle_3),
-                message: Text(Strings.SessionHeaderView.finishAlertMessage_1) +
+                    Text(session.name ?? Strings.SessionHeaderView.finishAlertTitle_2)
```
pawelgil, line 134: unneeded formatting (as per previous CRs - let's not reformat whole files since we're having some issues with conflicts emerging already)

pawelgil (changes requested):

AirCasting/SessionViews/SessionHeaderView.swift
jankrzempek: I will then. Haven't thought about viewmodel really, but it makes sense for sure after your explanation

AirCasting/SessionViews/SessionHeaderView.swift
jankrzempek: The longer "_" with the shorter "-"

AirCasting/SessionViews/SessionHeaderView.swift
jankrzempek: Totally my bad, apologises

pawelgil (approved):
Good comments, now it's clear

PR #212, opened 2021-09-10
changed stream order

jankrzempek:
https://trello.com/c/kT13LHxW/247-change-streams-order


smialko (commented):
Not sure why but it looks like there's additional padding between F stream name and the rest. 
IMG_9CA1590F517E-1

PR #213, opened 2021-09-10
color statistic container

smialko:
Added colors based on value and threshold for the selected stream to statistic container.


jankrzempek (approved):
🥳

AirCasting/Graph/StatisticsContainer/ViewModel/StatisticsContainerViewModelable.swift
```diff
@@ -11,7 +11,7 @@ struct SingleStatViewModel: Identifiable {
 
     let id: Int
     let title: String
-    let value: String
+    let value: Double
```
pawelgil, line 14: The viewmodel should provide already formatted stuff for the UI. Views should only be responsible for the layout, having all provided by the viewmodel

AirCasting/SessionViews/StatisticsContainerView.swift
```diff
@@ -9,6 +9,7 @@ import SwiftUI
 
 struct StatisticsContainerView<ViewModelType>: View where ViewModelType: StatisticsContainerViewModelable {
     @ObservedObject var statsContainerViewModel: ViewModelType
+    @ObservedObject var threshold: SensorThreshold
```
pawelgil, line 12: all things related to this (and this var itself too) should be moved to viewmodel since we're using one for this view. The viewmodel should be the only source of truth for the view and should decide about colors, label values, date formatting etc.

pawelgil (changes requested):
Let's move some logic to viewmodel

AirCasting/Graph/StatisticsContainer/ViewModel/StatisticsContainerViewModelable.swift
smialko: Added a String value, then :)

AirCasting/SessionViews/StatisticsContainerView.swift
smialko: Ok, so I tried, spent like 2 hours or more with different approaches (adding threshold/ thresholds, etc) and it always ends up in some dead end. 
So I decided to leave it as it is for now and switch to other essential tasks 😬

pawelgil (approved):
I'll try to figure it out on my side, for now we should merge this in as it's important.

PR #214, opened 2021-09-10
chart to match android

jankrzempek:
https://trello.com/c/T5EJqaY9/238-mobile-active-following-dashboard-chart-make-it-look-work-like-android

What was done:
Chart to match android - following point which were done here are on the ticker

### DIFF
--- a/AirCasting/SessionViews/SessionCartView.swift
+++ b/AirCasting/SessionViews/SessionCartView.swift
@@ -195,12 +195,12 @@ private extension SessionCartView {
     
     func pollutionChart(thresholds: [SensorThreshold]) -> some View {
         
-        let start = session.startTime
+        let start = session.startTime ?? Date()
         let end = session.endTime ?? Date()
         
         let dateFormatter : DateFormatter = DateFormatter()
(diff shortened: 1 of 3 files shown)


AirCasting/SessionViews/SessionCartView.swift
```diff
@@ -194,11 +194,31 @@ private extension SessionCartView {
...
-                ChartView(stream: selectedStream,
-                          thresholds: thresholds)
-                    .frame(height: 200)
+        
+        let start = session.startTime
+        let end = session.endTime ?? Date()
+        
+        let dateFormatter : DateFormatter = DateFormatter()
+        dateFormatter.dateFormat = "HH:mm"
+        let startTime = dateFormatter.string(from: start!)
```
smialko, line 203: Not a fan of force unwrapping 🤔

smialko (approved):
Seems legit 🦆

AirCasting/SessionViews/SessionCartView.swift
jankrzempek: Mee to, I've changed it

PR #216, opened 2021-09-11
Fix sync

smialko:
This branch was created to make corrections to syncing measurements (from Paweł's CR here: https://github.com/HabitatMap/AirCastingiOS/pull/173). The main change is adding the ViewModel for MeasurementsView. 
But while doing this I think I found the reason for the poor performance of syncing the measurements and did a refactor to make it better. It seems like it's working and it's much faster than it was 🥳


jankrzempek (approved):
Approved 🥳

AirCasting/CoreData/MeasurementStreamStorage.swift
```diff
@@ -31,13 +32,20 @@ final class CoreDataMeasurementStreamStorage: MeasurementStreamStorage {
 
     private let persistenceController: PersistenceController
     private lazy var updateSessionParamsService = UpdateSessionParamsService()
-
+    
+    let context: NSManagedObjectContext
```
pawelgil, line 36: I think this should be private

AirCasting/CoreData/MeasurementStreamStorage.swift
```diff
@@ -15,6 +15,7 @@ protocol MeasurementStreamStorage {
     func updateSessionEndtime(_ endTime: Date, for sessionUUID: SessionUUID) throws
     func updateSessionFollowing(_ sessionStatus: SessionFollowing, for sessionUUID: SessionUUID)
     func existingMeasurementStream(_ sessionUUID: SessionUUID, name: String) throws -> MeasurementStreamLocalID?
+    func save() throws
```
pawelgil, line 18: That's an interesting architectural problem here. It's perfectly fine to have a `save` method on the storage, but it poses a risk of someone forgetting it after making changes. It can be hard to track bugs due to that. I'm wondering if maybe we should be forcing those saves by enclosing other calls in some form of a batch updates block or something. Callsite would look like this:
```Swift
measurementStoarge.performUpdates { updateContext in
    for i in 0..<ONE_BILLION { updateContext.addMeasurement(...) }
}
```

This solution would require MeasurementStreamStorage to be designed a bit differently:

```Swift
protocol MasurementStreamStorage {
    func performUpdates(_ updateBlock: (MeasurementStreamStorageContext) -> Void)
}

protocol MeasurementStreamStorageContext {
    func addMeasurement(_ measurement: Measurement, toStreamWithID id: MeasurementStreamLocalID) throws
}
```

This way we can ensure that `context.save` is called each time we're leaving the `updateBlock` without forcing users of the storage to remember it by themselves.

this is just a thought, it can 100% just stay like this

AirCasting/CoreData/MeasurementStreamStorage.swift
```diff
@@ -31,13 +32,20 @@ final class CoreDataMeasurementStreamStorage: MeasurementStreamStorage {
...
+    let context: NSManagedObjectContext
+    
     init(persistenceController: PersistenceController) {
         self.persistenceController = persistenceController
+        self.context = persistenceController.editContext()
     }
-
+    
+    func save() throws {
+        try context.save()
```
pawelgil, line 44: this should really be dispatched to the context queue (like basically everything public in this class. We have no real control over what thread this func can be called from and NSManagedContexts are really thread sensitive)

AirCasting/SessionViews/SessionCartViewModel/MeasurementsViewModel.swift
```diff
@@ -0,0 +1,60 @@
...
+import CoreData
+
+class ViewModel: NSObject, ObservableObject {
+    
+    @Published var showLoadingIndicator = false
+    let sessionDownloader = SessionDownloadService(client: URLSession.shared,
+                                                   authorization: UserAuthenticationSession(),
+                                                   responseValidator: DefaultHTTPResponseValidator())
+    var measurementStreamStorage: MeasurementStreamStorage?
+    var task: Cancellable?
```
pawelgil, line 15: I think this stuff can be private? And shouldn't they be injected using `init`?

I feel that only the `func syncMeasurements` and `showLoadingIndicator` should be public here to avoid confusion

AirCasting/SessionViews/SessionCartViewModel/MeasurementsViewModel.swift
```diff
@@ -0,0 +1,60 @@
...
+//
+
+import Foundation
+import CoreLocation
+import CoreData
+
+class ViewModel: NSObject, ObservableObject {
+    
+    @Published var showLoadingIndicator = false
+    let sessionDownloader = SessionDownloadService(client: URLSession.shared,
```
pawelgil, line 11: instantiating own dependencies is not a good design, it should be passed in from the outside (probably using init). What's more - it's not protocolozed and as it's boundary-crossing (`sessionDownloader` is in delivery departament and viewmodels are UI) and it really should be. We can create a protocol for it real quick and easy:

```Swift
protocol MeasurementsDownloadable {
    func downloadSessionWithMeasurement(uuid: SessionUUID, completion: @escaping (Result<SessionsSynchronization.SessionDownstreamData, Error>) -> Void) -> Cancellable
}
```

and than just conform SessionDownloadService to it (as it's already doing so):

```Swift
final class SessionDownloadService: SessionDownstream, MeasurementsDownloadable {
...
}
```

This way here we can just add a dependency over this protocol instead of a whole `SessionDownloadService`:

```Swift
private let sessionDownloader: MeasurementsDownloadable
...
init(sessionDownloader: MeasurementsDownloadable ... ) { ... }
```

AirCasting/SessionViews/SessionCartViewModel/MeasurementsViewModel.swift
```diff
@@ -0,0 +1,60 @@
+// Created by Lunar on 11/09/2021.
+//
+
+import Foundation
+import CoreLocation
+import CoreData
+
+class ViewModel: NSObject, ObservableObject {
```
pawelgil, line 8: The name should be probably something more than a `ViewModel`. Maybe `MeasurementsViewModel` or even more verbose `CardMeasurementsViewModel`?

Also is `NSObject` needed?

AirCasting/SessionViews/ABMeasurementsView.swift
```diff
@@ -3,27 +3,26 @@
...
-import CoreData
 
 enum MeasurementPresentationStyle {
     case showValues
     case hideValues
 }
 
 struct ABMeasurementsView: View {
+    
+    @StateObject private var viewModel = ViewModel()
```
pawelgil, line 14: It's best when viewmodels are being passed in instead of created by the view layer itself. But as currently we have no good way of injecting dependencies I'm not sure there is a difference.

AirCasting/SessionViews/SessionCartViewModel/MeasurementsViewModel.swift
pawelgil, line 8: 2nd point here would be that it's really beneficial to have ViewModel protocols. You can see examples already in the app like `StatisticsContainerViewModelable` or `ForgotPasswordViewModel`. It's not much work to do it like that and the benefits are pretty cool. The main one would be that we can stub some static viewmodels instead of "real" ones that connect with the web while automating screenshot taking or implementing UI tests.

Not required tho.

pawelgil (changes requested):
Pretty nice. Some minor stuff found, also some optional tips

AirCasting/SessionViews/SessionCartViewModel/MeasurementsViewModel.swift
smialko: Right, thanks, dunno why I left it named ViewModel :D

AirCasting/SessionViews/ABMeasurementsView.swift
smialko: I had to do it anyways after marking `measurementStreamStorage: MeasurementStreamStorage?` as private.

smialko:
After adding the ViewModel the sync stopped working. It turned out that it didn't go through 
`guard let self = self else { return }` in `func syncMeasurements(for session: SessionEntity)`.
The reason for that was that `CardMeasurementsViewModel` was initialized in the `ABMeasurementsView` so it was created each time the view was refreshed. 
With a little bit of help I came up with the idea of `viewModelProvider` which solved this bug, the view model is being created just once.

AirCasting/CoreData/MeasurementStreamStorage.swift
```diff
@@ -31,13 +33,25 @@ final class CoreDataMeasurementStreamStorage: MeasurementStreamStorage {
...
+    private let context: NSManagedObjectContext
+    
     init(persistenceController: PersistenceController) {
         self.persistenceController = persistenceController
+        self.context = persistenceController.editContext()
     }
-
+    
+    func save() throws {
+        DispatchQueue.main.async {
```
pawelgil, line 44: this cannot be on main queue. The context is created as `.editContext()`, so it's queue will probably not be a main one. Use a [perform(_:)](https://developer.apple.com/documentation/coredata/nsmanagedobjectcontext/1506578-perform) or [performAndWait(_:)](https://developer.apple.com/documentation/coredata/nsmanagedobjectcontext/1506364-performandwait) in order to dispatch properly

AirCasting/CreateSessionViews/ConnectingABView.swift
```diff
@@ -93,9 +93,9 @@ struct ConnectingABView<VM: AirbeamConnectionViewModel>: View {
...
-struct ConnectingABView_Previews: PreviewProvider {
-    static var previews: some View {
-        ConnectingABView(viewModel: NeverConnectingAirbeamConnectionViewModel(), baseURL: DummyURLProvider(), creatingSessionFlowContinues: .constant(true))
-    }
- }
+//struct ConnectingABView_Previews: PreviewProvider {
+//    static var previews: some View {
+//        ConnectingABView<AirbeamConnectionViewModelDefault>(measurementsViewModel: NeverConnectingAirbeamConnectionViewModel(), viewModel: AirbeamConnectionViewModelDefault(airBeamConnectionController: <#AirBeamConnectionController#>, peripheral: ), baseURL: DummyURLProvider(), creatingSessionFlowContinues: .constant(true))
+//    }
+// }
```
pawelgil, line 100: let's remove it or fix it instead of commenting out

AirCasting/Graph/GraphView.swift
```diff
@@ -22,14 +22,18 @@ struct GraphView<StatsViewModelType>: View where StatsViewModelType: StatisticsC
...
                               sessionStopperFactory: sessionStoppableFactory).padding()
-
-            ABMeasurementsView(session: session,
+            
+            ABMeasurementsView(viewModelProvider: {
+                CardMeasurementsViewModel(measurementStreamStorage: measurementStreamStorage,
+                                          sessionDownloader: SessionDownloadService(client: URLSession.shared,
+                                                                                    authorization: UserAuthenticationSession(),
+                                                                                    responseValidator: DefaultHTTPResponseValidator()))
+            },
```
pawelgil, line 31: yeah, this is also not the place where we should be resolving those dependencies, but that's the best we can do currently without any DI rules and tools in place

AirCasting/SessionViews/ABMeasurementsView.swift
```diff
@@ -3,27 +3,51 @@
...
-    @Binding var showLoadingIndicator: Bool
     var thresholds: [SensorThreshold]
     let measurementPresentationStyle: MeasurementPresentationStyle
-    let sessionDownloader = SessionDownloadService(client: URLSession.shared,
-                                                   authorization: UserAuthenticationSession(),
-                                                   responseValidator: DefaultHTTPResponseValidator())
-    let measurementStreamStorage: MeasurementStreamStorage
-    @EnvironmentObject var selectedSection: SelectSection
 
+    @State private var viewModel: CardMeasurementsViewModel?
```
pawelgil, line 20: Try to use `@StateObject` instead of `@State` (ah, SwiftUI...) - it will probably remove the need of `viewModelProvider` (you will just inject ViewModel in init as you would typically do)

AirCasting/SessionViews/ABMeasurementsView.swift
```diff
@@ -3,27 +3,51 @@
...
+                                    measurementPresentationStyle: measurementPresentationStyle)
+            }
+        }
+        .onAppear {
+            viewModel = viewModelProvider()
+        }
+    }
+}
+
+struct _ABMeasurementsView: View {
```
pawelgil, line 39: not sure why the split into `_ABMeasurementsView` and `ABMeasurementsView`  🤔

AirCasting/SessionViews/ABMeasurementsView.swift
```diff
@@ -74,40 +98,9 @@ struct ABMeasurementsView: View {
...
-                                                                           thresholdLow: Int32(stream.thresholdLow),
-                                                                           thresholdVeryLow: Int32(stream.thresholdVeryLow))
-                            
-                        }
-                        showLoadingIndicator = false
-                    case .failure(let error):
-                        Log.info("\(error)")
-                    }
-                }
+                measurementsViewModel.syncMeasurements(for: session)
```
pawelgil, line 103: one of the most urgent things to clean up is usage of `CoreData` inside Views. The proper structure for this should just be:
```Swift
measurementsViewModel.syncMeasurements()
```

and `ViewModel` should have the entity injected on `init`:

```Swift
init(session: SessionEntity, ...) { ... }
```

It's still not perfect as ViewModels should also not talk to CD entities directly, but it's already a huge step forward.
We would also probably still need this entity to be present in the view, but step by step we should remove those dependencies and move it out to viewmodels

AirCasting/SessionViews/SessionCartView.swift
```diff
@@ -256,16 +260,17 @@ private extension SessionCartView {
...
+//        SessionCartView(session: SessionEntity.mock,
+//                                sessionCartViewModel: SessionCartViewModel(followingSetter: MockSessionFollowingSettable()),
+//                                thresholds: [.mock, .mock], sessionStoppableFactory: SessionStoppableFactoryDummy(), measurementStreamStorage: MeasurementStreamStorage.self as! MeasurementStreamStorage)
+//            .padding()
+//            .previewLayout(.sizeThatFits)
+//            .environmentObject(MicrophoneManager(measurementStreamStorage: PreviewMeasurementStreamStorage()))
+//    }
+// }
+// #endif
+
```
pawelgil, line 276: let's remove or fix

AirCasting/SessionViews/SessionCartViewModel/MeasurementsDownloadable.swift
```diff
@@ -0,0 +1,8 @@
+// Created by Lunar on 15/09/2021.
+//
+
+import Foundation
+
+protocol MeasurementsDownloadable {
```
pawelgil, line 6: That's the stuff :tada:

AirCasting/SessionViews/SessionCartViewModel/MeasurementsViewModel.swift
```diff
@@ -0,0 +1,67 @@
+// Created by Lunar on 11/09/2021.
+//
+
+import Foundation
+import CoreLocation
+import CoreData
+
+protocol SyncMeasurements {
```
pawelgil, line 8: `SyncMeasurements` is not a great name. It's usually a convention that imperative verbs are used for function names (like `doThis()`, `doThat()`). For protocols we use nouns or adjectives (which IMO better reflect what protocol actually is, but sometimes there is no easy adjective). Let's rename it? Perhaps make it a protocol for whole VM instead of just this sync function? So we don't directly use `CardMeasurementsViewModel` in a view. Examples can be found in code on how to do that (very tiny generics syntax is involved) like `ForgotPasswordViewModel`

AirCasting/SessionViews/SessionCartViewModel/MeasurementsViewModel.swift
```diff
@@ -0,0 +1,67 @@
...
+    init(measurementStreamStorage: MeasurementStreamStorage?, sessionDownloader: MeasurementsDownloadable) {
+        self.sessionDownloader = sessionDownloader
+        self.measurementStreamStorage = measurementStreamStorage
+    }
+    
+    func syncMeasurements(for session: SessionEntity) {
+        guard let measurementStreamStorage = measurementStreamStorage else { return }
+        showLoadingIndicator = true
+        
+        task = sessionDownloader.downloadSessionWithMeasurement(uuid: session.uuid) { [weak self] result in
```
pawelgil, line 28: thought: should we cancel download if user collapses the cell while download in progress?

pawelgil (changes requested):

AirCasting/SessionViews/ABMeasurementsView.swift
smialko: I might be wrong but I think general advice is that `@StateObject` should be private and not passed from the parent view.
I couldn't find much information about that. But here's one comment from Apple's engineer:
https://developer.apple.com/forums/thread/650618
From what I understand, he confirms that `@StateObject` should be initialized inside the view.
This I think makes sense. Even if passing a ViewModel to the @StateObjct wouldn't override the old object, it would still create that view model on each refresh or parent's view and deinitialize it immediately. It just smells bad 🤷🏻‍♀️

AirCasting/SessionViews/SessionCartViewModel/MeasurementsViewModel.swift
smialko: Dunno, I'll ask

smialko: Nope, they don't do that so I gues we won't be doing that either.

AirCasting/SessionViews/ABMeasurementsView.swift
smialko: It's used for view model provider, to avoid the situation described in the comment above ^

pawelgil (approved):
I see your point with `StateObject`. We need to figure out robust dependency management sooner rather than later

PR #219, opened 2021-09-13
map improvement

jankrzempek:
What was done:

There was a delay on especially long sessions on the map zooming and changing streams.

Should work better compering to develop now.

https://trello.com/c/9kOgl6ZX/278-mobile-active-mobile-dormant-map-pan-zoom-there-is-a-delay-between-pan-zoom-actions-response

### DIFF
--- a/AirCasting/Map/AirMapView.swift
+++ b/AirCasting/Map/AirMapView.swift
@@ -15,6 +15,7 @@ struct AirMapView: View {
     @StateObject var statsContainerViewModel: StatisticsContainerViewModel
     @StateObject var mapStatsDataSource: MapStatsDataSource
     @ObservedObject var session: SessionEntity
+    @Binding var showLoadingIndicator: Bool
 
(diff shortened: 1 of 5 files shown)


AirCasting/Map/AirMapView.swift
```diff
@@ -15,6 +15,7 @@ struct AirMapView: View {
     @StateObject var statsContainerViewModel: StatisticsContainerViewModel
     @StateObject var mapStatsDataSource: MapStatsDataSource
     @ObservedObject var session: SessionEntity
+    @Binding var showLoadingIndicator: Bool
```
pawelgil, line 18: Let's not change it, but for the record - I think that showing a loading indicator should not be implemented in the `AirMapView` itself. Switching between loading indicator and map view should be performed higher in the view hierarchy because loading state not an intrinsic part of the map view itself, it's a higher level concept.

pawelgil (approved):
Some optional improvements suggested

PR #221, opened 2021-09-13
Create cellular session

smialko:
Creating cellular AB session. 

https://trello.com/c/p07gVkJC/181-creating-fixed-cellular-session

### DIFF
--- a/AirCasting/CreateSessionViews/ConfirmCreatingSessionView.swift
+++ b/AirCasting/CreateSessionViews/ConfirmCreatingSessionView.swift
@@ -18,13 +18,16 @@ struct ConfirmCreatingSessionView: View {
     }
 
     @State private var isPresentingAlert: Bool = false
-    @State private var didStartRecordingSession = false
     @EnvironmentObject var selectedSection: SelectSection
     @EnvironmentObject private var sessionContext: CreateSessionContext
     @EnvironmentObject private var locationTracker: LocationTracker
     @EnvironmentObject private var tabSelection: TabBarSelection
+    @EnvironmentObject var persistenceController: PersistenceController
+    @EnvironmentObject var bluetoothManager: BluetoothManager
+    @EnvironmentObject var userAuthenticationSession: UserAuthenticationSession
+    @EnvironmentObject private var microphoneManager: MicrophoneManager
     @Binding var creatingSessionFlowContinues: Bool
-    let sessionCreator: SessionCreator
+    let baseURL: BaseURLProvider
     var sessionName: String
     private var sessionType: String { (sessionContext.sessionType ?? .fixed).description.lowercased() }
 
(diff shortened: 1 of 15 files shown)


AirCasting/CreateSessionViews/ConfirmCreatingSessionView.swift
```diff
@@ -18,13 +18,16 @@ struct ConfirmCreatingSessionView: View {
...
     @State private var isPresentingAlert: Bool = false
-    @State private var didStartRecordingSession = false
     @EnvironmentObject var selectedSection: SelectSection
     @EnvironmentObject private var sessionContext: CreateSessionContext
     @EnvironmentObject private var locationTracker: LocationTracker
     @EnvironmentObject private var tabSelection: TabBarSelection
+    @EnvironmentObject var persistenceController: PersistenceController
+    @EnvironmentObject var bluetoothManager: BluetoothManager
+    @EnvironmentObject var userAuthenticationSession: UserAuthenticationSession
+    @EnvironmentObject private var microphoneManager: MicrophoneManager
```
pawelgil, line 28: Something we need to address/discuss in the future:
problem with making all those `Manager` type classes an `EnvironmentObject` and importing directly in views is that it will reload the UI way too often - on each change of each of the managers. It will also take the possibility to decide when we want to reload the UI out of our hands. For example importing `LocationTracker` in a view where we just need the one latest location will still cause the UI to reload on each location fetch from `CLLocationManager`. It's pretty bad on its own, but If paired with another high-frequency `Manager` like `MicrophoneManager` those values can get really high and energy impact of the app skyrockets (maybe even UI cluttering). I feel like this `EnvironmentObject` thing is not supposed to be a dependency management tool, rather a way to share UI-related data transfer objects between views.

AirCasting/Models/AirBeamCellularSessionCreator.swift
```diff
@@ -0,0 +1,95 @@
...
+
+import Foundation
+import CoreLocation
+
+final class AirBeamCellularSessionCreator: SessionCreator {
+    enum AirBeamSessionCreatorError: Swift.Error {
+        case invalidCreateSessionContext(CreateSessionContext)
+    }
+    let userAuthenticationSession: UserAuthenticationSession
+    let measurementStreamStorage: MeasurementStreamStorage
```
pawelgil, line 12: i feel those two should be market `private` ?

AirCasting/Models/AirBeamCellularSessionCreator.swift
```diff
@@ -0,0 +1,95 @@
...
+    private let createSessionService: CreateSessionAPIService
+    
+    convenience init(measurementStreamStorage: MeasurementStreamStorage, userAuthenticationSession: UserAuthenticationSession, baseUrl: BaseURLProvider) {
+        self.init(measurementStreamStorage: measurementStreamStorage,
+                  createSessionService: CreateSessionAPIService(authorisationService: userAuthenticationSession, baseUrlProvider: baseUrl),
```
pawelgil, line 17: I don't quite like this introduction of low-level detail which is a `CreateSessionAPIService` here just for convenience reasons. I think we should remove this convenience init and only use interfaces. Moving purely clean-code reasons aside, this kind of import will cause longer compile times. This is because each time something inside the file containing `CreateSessionAPIService` changes it will also force this file to recompile (and all stuff that depends on it too, etc. etc.). I've seen iOS projects with 1h+ compile times. Not fun at all. This is why the [dependency inversion principle](https://stackify.com/dependency-inversion-principle/) is really important to follow

AirCasting/Models/AirBeamCellularSessionCreator.swift
```diff
@@ -0,0 +1,95 @@
...
+    
+    func createSession(_ sessionContext: CreateSessionContext, completion: @escaping (Result<Void, Error>) -> Void) {
+        guard let sessionType = sessionContext.sessionType,
+              let sessionUUID = sessionContext.sessionUUID else {
+            assertionFailure("invalidCreateSessionContext \(sessionContext)")
+            completion(.failure(AirBeamSessionCreatorError.invalidCreateSessionContext(sessionContext)))
+            return
+        }
+        // Save data to app's database
+        let session = Session(uuid: sessionUUID,
```
pawelgil, line 35: idk, maybe I'm missing it, but I think this is not needed. It won't save to database by itself (this is just a DTO struct after all), and it's not used later in the func.

AirCasting/Models/AirBeamCellularSessionCreator.swift
```diff
@@ -0,0 +1,95 @@
...
+                                                    notes: [],
+                                                    version: 0,
+                                                    streams: [:],
+                                                    latitude: sessionContext.startingLocation?.latitude,
+                                                    longitude: sessionContext.startingLocation?.longitude)
+        
+        createSessionService.createEmptyFixedWifiSession(input: .init(session: params,
+                                                                      compression: true),
+                                                         completion: { [measurementStreamStorage, userAuthenticationSession] result in
+                                                            DispatchQueue.main.async {
```
pawelgil, line 74: I'm not a fan of dispatching to main thread in objects other than views. Couple of main reasons:
* If you dispatch completion to main it will suddenly get much much harder to stub and test
* It's really a UIKit requirement, why would some low-level "creator" type object know about it after all? ([single responsibility principle](https://stackify.com/solid-design-principles/))
* The more we just casually dispatch to main instead of keeping stuff on background threads the quicker we'll run into UI lagging/cluttering issues. Also energy impact will get pretty high (it's already high for our app)

Please take a look if this dispatch can be moved out of here, preferably to a View. If it cannot be done easily - a warning to do so in the future would be nice

pawelgil (changes requested):

jankrzempek (approved):
I won't add anything more 😅

AirCasting/Models/AirBeamCellularSessionCreator.swift
smialko: I investigated a bit and I think it cannot be done easily. I'm adding the warning there, and the task in Trello

pawelgil: bummer, this main thread abuse is starting to come out ;) we should put it high on the list of things to do after mvp relase

pawelgil (approved):

PR #226, opened 2021-09-15
Onboarding changes differrences

jankrzempek:
https://trello.com/c/0HwaqLLe/252-differentiate-empty-onboarding-screens-for-mobile-fixed

### DIFF
--- a/AirCasting/CreateSessionViews/ChooseSessionType/ChooseSessionTypeView.swift
+++ b/AirCasting/CreateSessionViews/ChooseSessionType/ChooseSessionTypeView.swift
@@ -14,8 +14,8 @@ struct ChooseSessionTypeView: View {
     @State private var isPowerABLinkActive = false
     @State private var isMobileLinkActive = false
     @State private var didTapFixedSession = false
-    @EnvironmentObject private var tabSelection: TabBarSelection
     var viewModel: ChooseSessionTypeViewModel
+    @EnvironmentObject private var tabSelection: TabBarSelection
     
     var body: some View {
         NavigationView {
@@ -80,8 +80,8 @@ struct ChooseSessionTypeView: View {
                     didTapFixedSession = false
                 }
             }
-            .onAppear() {
-                if tabSelection.selection == .createSession && tabSelection.mobileProcceding {
+            .onAppear {
+                if tabSelection.selection == .createSession && tabSelection.mobileProceeding {
                     isMobileLinkActive = true
                 }
             }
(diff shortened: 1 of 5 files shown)


AirCasting/Dashboard/EmptyDashboardButtonView.swift
```diff
@@ -9,7 +9,12 @@ struct EmptyDashboardButtonView: View {
     var isFixed: Bool
     var body: some View {
         Button(action: {
-            tabSelection.selection = .createSession
+            if isFixed {
+                tabSelection.selection = .createSession
+            } else {
+                tabSelection.mobileProceeding = true
+                tabSelection.selection = .createSession
+            }
```
pawelgil, line 17: Can be shorter and without unnecessary conditional:

```Swift
tabSelection.selection = .createSelection
tabSelection.mobileProceeding = isFixed
```

AirCasting/MainTabBarView.swift
```diff
@@ -84,6 +84,7 @@ private extension MainTabBarView {
 
 class TabBarSelection: ObservableObject {
     @Published var selection = Tab.dashboard
+    @Published var mobileProceeding = false
```
pawelgil, line 87: what it stands for actually? I can't figure it out. Name is unclear or it doesn't belong here. `TabBarSelection` should only be about tab bar selection ;)

After reading the trello I'm quite convinced this is not the way it should be handled. Consider combining the value of `TabBarSelection` value with the `AirSectionPickerView.SelectedSection` to recognize which flow the app should take when create is pressed instead of this flag. Do it in appropriate layer. And don't be afraid of just creating new objects if you feel like something doesn't fit anywhere.

AirCasting/CreateSessionViews/ChooseSessionType/ChooseSessionTypeView.swift
```diff
@@ -79,6 +80,11 @@ struct ChooseSessionTypeView: View {
                     didTapFixedSession = false
                 }
             }
+            .onAppear {
+                if tabSelection.selection == .createSession && tabSelection.mobileProceeding {
```
pawelgil, line 84: related to my comment in `TabBarSelection`  - I have no idea what it accomplishes without looking at the Trello card. I feel like this is the easiest, but also the messiest way of modeling this. Consider changing it to something more robust and self explanatory

pawelgil (changes requested):
Let's change direction with this one.

Also - please try to improve commit messages as well as PR titles.

AirCasting/MainTabBarView.swift
jankrzempek: Yeah, just realise we have the ability to use something like this

jankrzempek: Cool. thank You!

PR #227, opened 2021-09-15
Chart fix

jankrzempek:
Michael added: 

-shrink chart dots
-y-axis label should display the "Unit symbol/abbreviation" not the "Short type of measurement", e.g. for RH the label should read "1 hr avg - %"


smialko (approved):
🥷🏻

PR #228, opened 2021-09-15
fix streams of AirBeam2/1

jankrzempek:
https://trello.com/c/kT13LHxW/247-change-streams-order

### DIFF
--- a/AirCasting/Models/SessionExtension.swift
+++ b/AirCasting/Models/SessionExtension.swift
@@ -19,7 +19,9 @@ extension SessionEntity {
     
     var pm1Stream: MeasurementStreamEntity? {
         let match = measurementStreams?.first(where: { (stream) -> Bool in
-            (stream as? MeasurementStreamEntity)?.sensorName == "AirBeam3-PM1"
+            (stream as? MeasurementStreamEntity)?.sensorName == "AirBeam3-PM1" ||
+                (stream as? MeasurementStreamEntity)?.sensorName == "AirBeam2-PM1" ||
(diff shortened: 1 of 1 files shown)


AirCasting/Models/SessionExtension.swift
```diff
@@ -19,7 +19,9 @@ extension SessionEntity {
     
     var pm1Stream: MeasurementStreamEntity? {
         let match = measurementStreams?.first(where: { (stream) -> Bool in
-            (stream as? MeasurementStreamEntity)?.sensorName == "AirBeam3-PM1"
+            (stream as? MeasurementStreamEntity)?.sensorName == "AirBeam3-PM1" ||
+                (stream as? MeasurementStreamEntity)?.sensorName == "AirBeam2-PM1" ||
+                    (stream as? MeasurementStreamEntity)?.sensorName == "AirBeam1-PM1"
```
smialko, line 24: Why do we need those AirBeams if we can't connect with them?

smialko (commented):
🙃

AirCasting/Models/SessionExtension.swift
jankrzempek: Because then PM1 stream from AirBeam2 or 1 is not recognised and stream goes the way, where PM1 is last and it should't be

smialko: Well, that's not what I meant but nvm, now I get it - we might have sessions recorded with AB1 and AB2 on Android.

smialko (approved):

PR #229, opened 2021-09-15
sensor name fix

jankrzempek:
https://trello.com/c/EGFDCPSp/282-all-tabs-cards-missing-sensor-package-name-wrong-sensor-package-name

For now (still) it needs to be handled somehow to show the right sensor name

### DIFF
--- a/AirCasting/SessionViews/SessionHeaderView.swift
+++ b/AirCasting/SessionViews/SessionHeaderView.swift
@@ -79,26 +79,28 @@ private extension SessionHeaderView {
     }
     
     var sensorType: some View {
-        var sensorNames = [String]()
+        var stream = [String]()
         var text = ""
-        for session in session.allStreams! {
+        
+        session.allStreams!.forEach { session in
             if var name = session.sensorPackageName {
-                if name.contains(":") {
-                    name = name.components(separatedBy: ":").first!
-                } else {
-                    name = name.components(separatedBy: "-").first!
-                }
+                componentsSeparation(name: &name)
                 (name == "Builtin") ? (name = "Phone mic") : (name = name)
-                !sensorNames.contains(name) ? sensorNames.append(name) : nil
+                !stream.contains(name) ? stream.append(name) : nil
             }
         }
-        for sensor in sensorNames {
-            text.append(sensor)
-            if sensor != sensorNames.last {
-                text.append(", ")
-            }
-        }
+        text = stream.joined(separator: ", ")
         return Text("\(session.type!.description) : \(text)")
+    }
+    
+    func componentsSeparation(name: inout String) {
+        // separation is used to nicely handle the case where sensor could be
+        // AirBeam2-xxxx or AirBeam2:xxx
+        if name.contains(":") {
+            name = name.components(separatedBy: ":").first!
+        } else {
+            name = name.components(separatedBy: "-").first!
+        }
     }
     
     var actionsMenuMobile: some View {
--- a/AirCasting/SessionViews/SessionHeaderView.swift
+++ b/AirCasting/SessionViews/SessionHeaderView.swift
@@ -72,24 +72,35 @@ private extension SessionHeaderView {
             //  |_________|     |-------------------|
             // so the idea at leat for now is this below
             #warning("Fix - Handle session.deviceType (for now it is always nill)")
-            switch session.type?.description {
-            case "Fixed":
-                Text("\(session.type!.description) : AirBeam3")
-                    .font(Font.moderate(size: 13, weight: .regular))
-            case "Mobile":
-                if session.allStreams!.count > 1 {
-                    Text("\(session.type!.description): AirBeam3")
-                        .font(Font.moderate(size: 13, weight: .regular))
-                } else {
-                    Text("\(session.type!.description): Phone Mic")
-                        .font(Font.moderate(size: 13, weight: .regular))
-                }
-            default:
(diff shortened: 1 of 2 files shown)


AirCasting/SessionViews/SessionHeaderView.swift
```diff
@@ -72,24 +72,35 @@ private extension SessionHeaderView {
...
+        switch session.type.description {
             case "Fixed":
-                Text("\(session.type!.description) : AirBeam3")
-                    .font(Font.moderate(size: 13, weight: .regular))
+                return airBeamText
             case "Mobile":
-                if session.allStreams!.count > 1 {
-                    Text("\(session.type!.description): AirBeam3")
-                        .font(Font.moderate(size: 13, weight: .regular))
+                if session.allStreams!.count == 1 {
```
smialko, line 86: I'm not a fan of this - we can have a session with one stream which isn't mic session - we might have session with removed 4/5 streams created on an android device. So it's not edgy edge case :D 
What would you say about creating sth like this in SessionEntity extension: 
`var isMic: Bool { deviceType =  .MIC }` ? :D 
Might be useful in the future, too :D

AirCasting/SessionViews/SessionHeaderView.swift
```diff
@@ -72,24 +72,35 @@ private extension SessionHeaderView {
...
             default:
-                Text(session.deviceType?.description ?? "")
-                    .font(Font.moderate(size: 13, weight: .regular))
-            }
+                return Text(session.deviceType?.description ?? "")
+        }
+    }
+    
+    var airBeamText: Text {
+        if session.pm1Stream?.sensorName?.contains(Strings.SessionHeaderView.airBeam1) != nil {
```
smialko, line 97: that's kind of weird expression ;D 
What would you say about adding device types here and then using them in this expression?  

```
public enum DeviceType: Int, CustomStringConvertible {
    case MIC = 0
    case AIRBEAM3 = 1

    public var description: String {
        switch self {
        case .MIC: return "Device's Microphone"
        case .AIRBEAM3: return "AirBeam 3"
        }
    }
}
```

smialko: correct me if I'm wrong but I think you'd had to check the integers for AB1 and AB2 in android code

smialko (changes requested):
🦆

AirCasting/SessionViews/SessionHeaderView.swift
smialko: PS we already have extension with isMobile, isFixed, etc, please add it there ;)

AirCasting/SessionViews/SessionHeaderView.swift
pawelgil, line 97: this is really convoluted and a bit weird (like - why are you using pm1Stream here? -> it's getting a bit clear after really focusing on reading this piece of code, but the fact that this question is asked is already a big tell that something is maybe under-designed).

Another pitfall - imagine there is an AB4 introduce like a year from now. Will you remember to go here and add another `else if`? This should really be more universal than this or at least be future proof in some way.

pawelgil (changes requested):

AirCasting/SessionViews/SessionHeaderView.swift
```diff
@@ -72,24 +72,33 @@ private extension SessionHeaderView {
...
                 }
-            default:
-                Text(session.deviceType?.description ?? "")
-                    .font(Font.moderate(size: 13, weight: .regular))
+                (name == "Builtin") ? (name = "Phone mic") : (name = name)
+                !sensorNames.contains(name) ? sensorNames.append(name) : nil
             }
         }
-        .foregroundColor(.darkBlue)
+        for sensor in sensorNames {
```
pawelgil, line 95: no need for this loop, just use `sensorNames.joined(separator: ", ")`

pawelgil (changes requested):

AirCasting/SessionViews/SessionHeaderView.swift
jankrzempek: WOW, Cool!

AirCasting/SessionViews/SessionHeaderView.swift
```diff
@@ -72,24 +72,33 @@ private extension SessionHeaderView {
...
+            sensorType
+                .font(Font.moderate(size: 13, weight: .regular))
+        }
+        .foregroundColor(.darkBlue)
+    }
+    
+    var sensorType: some View {
+        var sensorNames = [String]()
+        var text = ""
+        for session in session.allStreams! {
```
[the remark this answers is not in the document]
jankrzempek: Of course, will be done then

AirCasting/SessionViews/SessionHeaderView.swift
```diff
@@ -79,28 +79,30 @@ private extension SessionHeaderView {
...
-        for session in session.allStreams! {
+        
+        session.allStreams!.forEach { session in
             if var name = session.sensorPackageName {
-                if name.contains(":") {
-                    name = name.components(separatedBy: ":").first!
-                } else {
-                    name = name.components(separatedBy: "-").first!
-                }
+                componentsSeparation(name: &name)
```
pawelgil, line 87: for the future: please use verbs to name functions. In this case it would be `separateComponents(name: &name)`

pawelgil (approved):
Much better! Nice work

PR #243, opened 2021-09-22
Added ability to cut out UI updates from storage (Core Data stack)

pawelgil:
## What was done?
Core data stack changed to a parent/child model with a `single source of truth` pattern:
1. Only one, "main" `NSManagedObjectContext` is responsible for writing to the store (and it's a background one)
2. The UI Context as well as edit contexts are children of the main context
3. Changes from the main context to the UI context are being propagated only when the new `uiSuspended` flag is not set

## Technical notes
* `FixedSessionFaker` left in the code - it might be handy (it's faking a running fixed session with random measurements)
### New CD stack structure
new_coredata

### Old CD stack structure
old_coredata

### DIFF
--- a/AirCasting/AirCastingApp.swift
+++ b/AirCasting/AirCastingApp.swift
@@ -64,12 +64,8 @@ struct AirCastingApp: App {
             case .active:
                 persistenceController.uiSuspended = false
                 appBecameActive.send()
-            case .background:
+            case .background, .inactive:
                 persistenceController.uiSuspended = true
-                break
-            case .inactive:
-                persistenceController.uiSuspended = true
-                break
             @unknown default:
                 fatalError()
             }
--- a/AirCasting/AirCastingApp.swift
+++ b/AirCasting/AirCastingApp.swift
@@ -28,7 +28,6 @@ struct AirCastingApp: App {
         FirebaseApp.configure()
         #endif
         AppBootstrap(firstRunInfoProvider: lifeTimeEventsProvider, deauthorizable: authorization).bootstrap()
-        
         let synchronizationContextProvider = SessionSynchronizationService(client: URLSession.shared, authorization: authorization, responseValidator: DefaultHTTPResponseValidator())
         let downloadService = SessionDownloadService(client: URLSession.shared, authorization: authorization, responseValidator: DefaultHTTPResponseValidator())
         let uploadService = SessionUploadService(client: URLSession.shared, authorization: authorization, responseValidator: DefaultHTTPResponseValidator())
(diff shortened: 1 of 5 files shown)


smialko (approved):
Seems legit

AirCasting/AirCastingApp.swift
```diff
@@ -62,10 +62,13 @@ struct AirCastingApp: App {
...
             switch newScenePhase {
             case .active:
+                persistenceController.uiSuspended = false
                 appBecameActive.send()
             case .background:
+                persistenceController.uiSuspended = true
                 break
             case .inactive:
+                persistenceController.uiSuspended = true
                 break
```
jankrzempek, line 72: Don't You want to merge those two into one, as they have the same condition?

FixedSessionFaker.swift
```diff
@@ -0,0 +1,73 @@
...
+class FixedSessionFaker {
+    init(context: NSManagedObjectContext) {
+        let req: NSFetchRequest<SessionEntity> = SessionEntity.fetchRequest()
+        req.predicate = NSPredicate(format: "name LIKE %@", "Its complicated")
+        var session: SessionEntity!
+        if let already = try! context.fetch(req).first {
+            print("A")
+            session = already
+        } else {
+            print("B")
```
jankrzempek, line 16: Have you intentionally left those prints?

FixedSessionFaker.swift
```diff
@@ -0,0 +1,73 @@
...
+            session = SessionEntity(context: context)
+            session.type = .fixed
+            session.uuid = SessionUUID()
+            session.name = "Its complicated"
```
jankrzempek, line 20: Don't know if we want to use Strings to keep this somewhere or not 😄

jankrzempek (approved):
👌🏻

AirCasting/CoreData/PersistenceController.swift
```diff
@@ -6,11 +6,28 @@
...
+    private(set) lazy var viewContext: NSManagedObjectContext = {
+        let ctx = NSManagedObjectContext(concurrencyType: .mainQueueConcurrencyType)
+        ctx.parent = mainContext
+        ctx.automaticallyMergesChangesFromParent = true
+        return ctx
+    }()
+    
     private let container: NSPersistentContainer
+    
+    private lazy var mainContext: NSManagedObjectContext = container.newBackgroundContext()
```
smialko, line 30: This name is misleading, it's not a main context ;)

FixedSessionFaker.swift
smialko: As far as I remember we created Strings to contain texts that are displayed for users, this one's not :)

smialko (approved):
😎

smialko (approved):
You may wonder why approving, then requesting changes, and then approving again. Well, it turns out that the thing that I thought wasn't working on your branch isn't working on develop :D

AirCasting/CoreData/PersistenceController.swift
pawelgil: well, it is :) it's not main-thread context, but it indeed is our main context - the one that holds all the data and from which we spawn children. But I get your point - it might be a bit confusing beacause of the threading nature of core data. Will change it to something better

FixedSessionFaker.swift
pawelgil: Yes. This is only a debug-tool for programmers, it's not a subject for `Strings`

FixedSessionFaker.swift
pawelgil: no, good catch 💯

AirCasting/AirCastingApp.swift
pawelgil: good catch!

PR #244, opened 2021-09-22
time formatting in sessions header

jankrzempek:
https://trello.com/c/RmN7eePA/74-time-displayed-on-top-of-the-session-is-incorrect

For now, after a lot of testing, I think it should be working well. Please do not look at the graph, for now, I will focus on this on the other ticket, not to messed it up, and make it in small steps.
I have added there a code that results in logging out from the app without a crash. I know it is not perfect at all, but after thinking it could be easier for us and a better UX for the user. Please @pawelgil consider it, I am going to leave there a warning (It was working for me every time I logged out).
PS. Thank You @pawelgil for pair programming 😄


pawelgil (approved):
Seems a bit odd that it does the trick since abbreviation and identifier can be used interchangeably in case of a UTC, but 🤷‍♂️

PR #245, opened 2021-09-23
Refactor to fix app crashing after creating ABmobile session & syncing measurments

smialko:
Long story short I added an `accessStorage` which ensures thread-safety by dispatching all calls to the queue owned by the NSManagedObjectContext. This way app won't crash on creating a mobile AB session, the graph is being drawn, measurement can be synced, etc.

### DIFF
--- a/AirCasting/CoreData/MeasurementStreamStorage.swift
+++ b/AirCasting/CoreData/MeasurementStreamStorage.swift
@@ -7,6 +7,10 @@ import Foundation
 import Combine
 
 protocol MeasurementStreamStorage {
+    func accessStorage(_ task: @escaping(HiddenCoreDataMeasurementStreamStorage) -> Void)
+}
+
+protocol MeasurementStreamStorageContextUpdate {
     func addMeasurement(_ measurement: Measurement, toStreamWithID id: MeasurementStreamLocalID) throws
     func saveThresholdFor(sensorName: String, thresholdVeryHigh: Int32, thresholdHigh: Int32, thresholdMedium: Int32, thresholdLow: Int32, thresholdVeryLow: Int32) throws
     func createMeasurementStream(_ stream: MeasurementStream, for sessionUUID: SessionUUID) throws -> MeasurementStreamLocalID
@@ -19,7 +23,7 @@ protocol MeasurementStreamStorage {
     func save() throws
 }
 
-extension MeasurementStreamStorage {
+extension HiddenCoreDataMeasurementStreamStorage {
     func addMeasurementValue(_ value: Double, at location: CLLocationCoordinate2D? = nil, toStreamWithID id: MeasurementStreamLocalID) throws {
         try addMeasurement(Measurement(time: Date(), value: value, location: location), toStreamWithID: id)
         try save()
@@ -27,11 +31,35 @@ extension MeasurementStreamStorage {
 }
 
 final class CoreDataMeasurementStreamStorage: MeasurementStreamStorage {
+
+    private let persistenceController: PersistenceController
+    private lazy var updateSessionParamsService = UpdateSessionParamsService()
+    private let context: NSManagedObjectContext
+    let hiddenStorage: HiddenCoreDataMeasurementStreamStorage
+    
+    init(persistenceController: PersistenceController) {
+        self.persistenceController = persistenceController
+        self.hiddenStorage = HiddenCoreDataMeasurementStreamStorage(persistenceController: persistenceController)
+        self.context = persistenceController.editContext()
+    }
+    
+    /// All actions performed on CoreDataMeasurementStreamStorage must be performed
+    /// within a block passed to this methood.
+    /// This ensures thread-safety by dispatching all calls to the queue owned by the NSManagedObjectContext.
+    func accessStorage(_ task: @escaping(HiddenCoreDataMeasurementStreamStorage) -> Void) {
+        context.perform {
+            task(self.hiddenStorage)
+            try? self.hiddenStorage.save()
+        }
+    }
+}
+
+final class HiddenCoreDataMeasurementStreamStorage: MeasurementStreamStorageContextUpdate {
     enum Error: Swift.Error {
         case missingMeasurementStream
         case missingSensorName
     }
-
+    
     private let persistenceController: PersistenceController
     private lazy var updateSessionParamsService = UpdateSessionParamsService()
     private let context: NSManagedObjectContext
@@ -42,17 +70,10 @@ final class CoreDataMeasurementStreamStorage: MeasurementStreamStorage {
     }
     
     func save() throws {
-        context.perform {
-            do {
-                try self.context.save()
-            } catch {
-                Log.info("Couldn't save the context: \(self.context)")
-            }
-        }
+        try self.context.save()
     }
-    
+
     func addMeasurement(_ measurement: Measurement, toStreamWithID id: MeasurementStreamLocalID) throws {
-        
         let stream = try context.existingObject(with: id.id) as! MeasurementStreamEntity
 
         let newMeasurement = MeasurementEntity(context: context)
@@ -148,18 +169,16 @@ final class CoreDataMeasurementStreamStorage: MeasurementStreamStorage {
     }
     
     func updateSessionFollowing(_ sessionFollowing: SessionFollowing, for sessionUUID: SessionUUID) {
-        context.performAndWait {
-            do {
-                let sessionEntity = try context.existingSession(uuid: sessionUUID)
-                if sessionFollowing.rawValue == 1 {
-                    sessionEntity.followedAt = Date()
-                } else {
-                    sessionEntity.followedAt = nil
-                }
-                try context.save()
-            } catch {
-                Log.info("Error when saving changes in session")
+        do {
+            let sessionEntity = try context.existingSession(uuid: sessionUUID)
+            if sessionFollowing == SessionFollowing.following {
+                sessionEntity.followedAt = Date()
+            } else {
+                sessionEntity.followedAt = nil
             }
+            try context.save()
+        } catch {
+            Log.info("Error when saving changes in session")
         }
     }
 
@@ -173,6 +192,10 @@ final class CoreDataMeasurementStreamStorage: MeasurementStreamStorage {
 #if DEBUG
 /// Only to be used for swiftui previews
 final class PreviewMeasurementStreamStorage: MeasurementStreamStorage {
+    func accessStorage(_ task: @escaping (HiddenCoreDataMeasurementStreamStorage) -> Void) {
+        print("accessing storage")
+    }
+    
     func save() throws {
         print("Faking saving ")
     }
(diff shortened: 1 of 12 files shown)


AirCasting/CoreData/MeasurementStreamStorage.swift
```diff
@@ -7,6 +7,7 @@ import Foundation
 import Combine
 
 protocol MeasurementStreamStorage {
+    func accessStorage(_ task: @escaping(MeasurementStreamStorage) -> Void)
```
pawelgil, line 10: some PRs ago (when `save()` was introduced) I proposed a solution that would ensure secure access during compile time instead of runtime, maybe we can revisit

AirCasting/CoreData/MeasurementStreamStorage.swift
```diff
@@ -41,18 +42,26 @@ final class CoreDataMeasurementStreamStorage: MeasurementStreamStorage {
...
+    /// within a block passed to this methood.
+    /// This ensures thread-safety by dispathing all calls to the queue owned by the NSManagedObjectContext.
+    func accessStorage(_ task: @escaping(MeasurementStreamStorage) -> Void) {
         context.perform {
-            do {
-                try self.context.save()
-            } catch {
-                Log.info("Couldn't save the context: \(self.context)")
-            }
+            self.isSafelyAccessed = true
```
pawelgil, line 52: this is quite clever 🧠 But I don't see why can't we just do `context.perform { }` inside each method instead

AirCasting/CreateSessionViews/ConfirmCreatingSessionView.swift
```diff
@@ -16,66 +16,63 @@ struct ConfirmCreatingSessionView: View {
             isPresentingAlert = error != nil
         }
     }
-
```
pawelgil, line 19: The only modifications in this file are formatting ones, we should avoid that (due to conflicts)

AirCasting/MicrophoneSession/MicrophoneManager.swift
```diff
@@ -78,7 +86,11 @@ extension MicrophoneManager: AVAudioRecorderDelegate {
     func audioRecorderBeginInterruption(_ recorder: AVAudioRecorder) {
         Log.warning("audio recorder interruption began")
         levelTimer?.invalidate()
-        try! measurementStreamStorage.updateSessionStatus(.DISCONNECTED, for: session!.uuid)
+        measurementStreamStorage.accessStorage { [self] storage in
```
pawelgil, line 89: strong capturing `[self]` is redundant (seems like `self` is not used inside too)

AirCasting/MicrophoneSession/MicrophoneManager.swift
```diff
@@ -100,15 +112,22 @@ private extension MicrophoneManager {
         let power = recorder.averagePower(forChannel: 0)
         let decibels = Double(power + 90.0)
         let location = obtainCurrentLocation()
-        Log.debug("New mic measurement \(decibels) at \(String(describing: location))")
-        try measurementStreamStorage.addMeasurementValue(decibels, at: location, toStreamWithID: measurementStreamLocalID!)
+        
+        measurementStreamStorage.accessStorage { [self] storage in
```
pawelgil, line 116: strong capturing `[self]` is redundant (seems like `self` is not used inside too)

AirCasting/MicrophoneSession/MicrophoneManager.swift
```diff
@@ -121,7 +140,15 @@ private extension MicrophoneManager {
                                        thresholdMedium: 70,
                                        thresholdLow: 60,
                                        thresholdVeryLow: 20)
-        return try measurementStreamStorage.createSessionAndMeasurementStream(session, stream)
+        
+        measurementStreamStorage.accessStorage { [self] storage in
```
pawelgil, line 144: strong capturing `[self]` is redundant (seems like `self` is not used inside too)

AirCasting/MobilePeripheralSessionManager.swift
```diff
@@ -17,7 +17,13 @@ class MobilePeripheralSessionManager {
     }
     
     func startRecording(session: Session, peripheral: CBPeripheral) throws {
```
pawelgil, line 19: this function doesn't throw anymore

AirCasting/MobilePeripheralSessionManager.swift
```diff
@@ -17,7 +17,13 @@ class MobilePeripheralSessionManager {
...
-        try measurementStreamStorage.createSession(session)
+        measurementStreamStorage.accessStorage { storage in
+            do {
+                try storage.createSession(session)
+            } catch {
+                Log.info("\(error)")
+            }
+        }
         locationProvider.requestLocation()
         activeMobileSession = MobileSession(peripheral: peripheral, session: session)
```
pawelgil, line 28: in case of an error in creating session - will this be a problem (nil session passed in here)?

AirCasting/MicrophoneSession/MicrophoneManager.swift
```diff
@@ -100,15 +112,22 @@ private extension MicrophoneManager {
...
+            }
+        }
     }
 
     @objc func timerTick() {
         try! sampleMeasurement()
     }
 
-    func createMeasurementStream(for session: Session) throws -> MeasurementStreamLocalID {
+    func createMeasurementStream(for session: Session, completion: @escaping(Result<MeasurementStreamLocalID, Error>) -> Void) throws {
```
pawelgil, line 130: does it really throw tho?

AirCasting/MicrophoneSession/MicrophoneManager.swift
```diff
@@ -100,15 +112,22 @@ private extension MicrophoneManager {
         let power = recorder.averagePower(forChannel: 0)
         let decibels = Double(power + 90.0)
         let location = obtainCurrentLocation()
-        Log.debug("New mic measurement \(decibels) at \(String(describing: location))")
-        try measurementStreamStorage.addMeasurementValue(decibels, at: location, toStreamWithID: measurementStreamLocalID!)
+
```
pawelgil, line 115: I think this function doesn't throw anymore and is still marked as throws

AirCasting/SessionViews/SessionCartViewModel/DefaultSyncingMeasurementsViewModel.swift
```diff
@@ -49,19 +49,37 @@ final class DefaultSyncingMeasurementsViewModel: SyncingMeasurementsViewModel {
...
-                            Log.info("failed to get existing streamID for synced measurements from session \(String(describing: self.session.name))")
-                            return }
-                        try? measurementStreamStorage.addMeasurement(Measurement(time: measurement.time, value: measurement.value, location: location), toStreamWithID: streamID)
+                        
+                        let sessionId = self.session.uuid!
+                        let sensorName = stream.sensorName
+                        let sessionName = self.session.name
+                        measurementStreamStorage.accessStorage { storage in
+                            do {
+                                guard let streamID = try? storage.existingMeasurementStream(sessionId, name: sensorName) else {
```
pawelgil, line 58: we can omit `guard` and use a regular `try` since we're catching anyway

pawelgil (approved):
Seems ok, some cosmetics and suggestions left

AirCasting/CoreData/MeasurementStreamStorage.swift
```diff
@@ -71,6 +80,7 @@ final class CoreDataMeasurementStreamStorage: MeasurementStreamStorage {
     }
     
     func saveThresholdFor(sensorName: String, thresholdVeryHigh: Int32, thresholdHigh: Int32, thresholdMedium: Int32, thresholdLow: Int32, thresholdVeryLow: Int32) throws {
+        assert(isSafelyAccessed, "Func saveThresholdFor() should be acccess within accessStorage()")
```
jankrzempek, line 83: Don't You want to use Strings to keep those phrase and others below? if now - ok 😄

jankrzempek (approved):
👌🏻👌🏻

AirCasting/CoreData/MeasurementStreamStorage.swift
smialko: That's what I originally intended, but I eventually decided this is a better way.
Having `perform` in every method makes them all asynchronous and so we can no longer throw from them. Instead, I'd need to add `completionBlocks` to all of them. Other callers usually call a few methods of `MeasurementStreamStorage` in sequence and thus we'd end up having pyramids of doom everywhere.🤷🏻‍♀️

AirCasting/CoreData/MeasurementStreamStorage.swift
smialko: I thought Strings are texts that we display to users, aren't they? Now I'm not sure ;D

AirCasting/MicrophoneSession/MicrophoneManager.swift
smialko: It seems like it but when I remove it I get en error "Reference to property 'session' in closure requires explicit use of 'self' to make capture semantics explicit". That's why it's there :)

smialko: I can also have it here, not sure if it makes a difference: 
`                try! storage.updateSessionStatus(.DISCONNECTED, for: self.session!.uuid)
`

AirCasting/MicrophoneSession/MicrophoneManager.swift
smialko: same as above :)

AirCasting/MicrophoneSession/MicrophoneManager.swift
smialko: Removed

AirCasting/MobilePeripheralSessionManager.swift
smialko: Hmm, not sure. Maybe I can wrap this in if let, like this: 

```
        if let session = Session(uuid: sessionUUID,
                                 type: sessionType,
                                 name: sessionContext.sessionName,
                                 deviceType: sessionContext.deviceType,
                                 location: startingLocation,
                                 startTime: Date(),
                                 contribute: contribute) {
            do {
                try microphoneManager.startRecording(session: session)
                completion(.success(()))
            } catch {
                assertionFailure("Can't start recording microphone session: \(error)")
                completion(.failure(error))
            }
        }
```

smialko: True, what would you say about sth like this: 
```
        measurementStreamStorage.accessStorage { [weak self] storage in
            do {
                try storage.createSession(session)
                DispatchQueue.main.async {
                    self?.locationProvider.requestLocation()
                    self?.activeMobileSession = MobileSession(peripheral: peripheral, session: session)
                }
            } catch {
                Log.info("\(error)")
            }
        }
```

AirCasting/SessionViews/SessionCartViewModel/DefaultSyncingMeasurementsViewModel.swift
smialko: We can't we need to pass nonoptional `streamID` to `addMeasurementValue`. Without guard `streamID` is optional. 
I can use `if let` instead of `guard let` 🤷🏻‍♀️

PR #248, opened 2021-09-24
Popping navigation stack when in mapview and going background

pawelgil:
When user leaves the app while in map view it will pop to prevent major CPU usage and app kill. **This is a bandaid solution**


jankrzempek (approved):

smialko:
@pawelgil I'm merging your code bc we want to test both of your branches at once 💪🏻

PR #252, opened 2021-09-25
Disabling mobile session state

jankrzempek:
https://trello.com/c/2FO8t6Om/305-bug-during-creating-a-mobile-session

### DIFF
--- a/AirCasting/CreateSessionViews/SelectDeviceView.swift
+++ b/AirCasting/CreateSessionViews/SelectDeviceView.swift
@@ -20,6 +20,7 @@ struct SelectDeviceView: View {
     @EnvironmentObject private var microphoneManager: MicrophoneManager
     @Binding var creatingSessionFlowContinues : Bool
     @State private var showAlert = false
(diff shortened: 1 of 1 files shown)


AirCasting/CreateSessionViews/SelectDeviceView.swift
```diff
@@ -20,6 +20,7 @@ struct SelectDeviceView: View {
     @EnvironmentObject private var microphoneManager: MicrophoneManager
     @Binding var creatingSessionFlowContinues : Bool
     @State private var showAlert = false
+    @EnvironmentObject private var emptyDashboardButtonTapped: EmptyDashboardButtonTapped
```
pawelgil, line 23: `EmptyDashboardButtonTapped` is not a good solution in the first place. It's never a good idea to add more state than absolutely necessary. And it's basically a global state.

I'm wondering why didn't you use `SelectSection` (which is already an `EnvironmentObject` *and* also already present in the `ChooseSessionTypeView`) ?

jankrzempek: I am going to add a warning here and pushed it for now, because `SelectSection ` won't handle that for now

PR #253, opened 2021-09-26
MeasurementUpdatingService is now being run on background thread instead of main

pawelgil:
... which was causing the app to freeze like crazy 🥶🧊🥶🧊🥶🧊


jankrzempek (approved):
👌🏻

smialko:
@pawelgil I'm merging your code bc we want to test both of your branches at once 💪🏻

PR #261, opened 2021-09-27
added spinner to airBeam mobile

jankrzempek:
https://trello.com/c/Mb4Qkbho/306-mobile-active-spinner-not-shown-at-end-of-wizard

### DIFF
--- a/AirCasting/SessionViews/ABMeasurementsView.swift
+++ b/AirCasting/SessionViews/ABMeasurementsView.swift
@@ -85,7 +85,7 @@ struct _ABMeasurementsView: View {
                         measurementsTitle
                             .font(Font.moderate(size: 12))
                         streamNames
-                        if session.type == .mobile {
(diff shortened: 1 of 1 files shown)


AirCasting/SessionViews/ABMeasurementsView.swift
```diff
@@ -85,9 +85,16 @@ struct _ABMeasurementsView: View {
                         measurementsTitle
                             .font(Font.moderate(size: 12))
                         streamNames
-                        if !isCollapsed && measurementsViewModel.showLoadingIndicator {
-                            ProgressView()
-                                .progressViewStyle(CircularProgressViewStyle())
+                        if session.type == .mobile && session.deviceType == .AIRBEAM3 {
```
pawelgil, line 88: all this should be done in a viewmodel since this change is not layout related. It's a view logic and the viewmodel should be responsible for a view logic, not the view itself. I think this file should not be changed at all in order to implement required changes.

PR #262, opened 2021-09-27
Handle right time graph

jankrzempek:
https://trello.com/c/DhHadEvW/304-graph-incorrect-time

### DIFF
--- a/AirCasting/Graph/AirCastingGraph.swift
+++ b/AirCasting/Graph/AirCastingGraph.swift
@@ -51,13 +51,13 @@ class AirCastingGraph: UIView {
         lineChartView.legend.enabled = false
         lineChartView.scaleYEnabled = false
         
-        lineChartView.xAxis.drawLabelsEnabled = true
+        lineChartView.xAxis.drawLabelsEnabled = false
         lineChartView.xAxis.labelCount = 2
         lineChartView.extraBottomOffset = 2
         
         lineChartView.highlightPerTapEnabled = false
         lineChartView.highlightPerDragEnabled = false
-        
+     
 //        lineChartView.xAxisRenderer = TimeAxisRenderer(viewPortHandler: lineChartView.viewPortHandler,
 //                                                       xAxis: lineChartView.xAxis,
 //                                                       transformer: lineChartView.getTransformer(forAxis: .left))
@@ -110,7 +110,7 @@ class AirCastingGraph: UIView {
         if !didMoveOrScaleGraph && isAutozoomEnabled {
             zoomoutToThirtyMinutes(dataSet: dataSet)
         }
-        callDateRangeChangeObserver()
+//        callDateRangeChangeObserver()
     }
     
     private func callDateRangeChangeObserver() {
@@ -149,12 +149,12 @@ extension AirCastingGraph: ChartViewDelegate {
     
     // Callbacks when the chart is scaled / zoomed via pinch zoom gesture.
     @objc func chartScaled(_ chartView: ChartViewBase, scaleX: CGFloat, scaleY: CGFloat) {
-        callDateRangeChangeObserver()
+//        callDateRangeChangeObserver()
         didMoveOrScaleGraph = true
     }
     // Callbacks when the chart is moved / translated via drag gesture.
     @objc func chartTranslated(_ chartView: ChartViewBase, dX: CGFloat, dY: CGFloat) {
-        callDateRangeChangeObserver()
+//        callDateRangeChangeObserver()
         didMoveOrScaleGraph = true
     }
 }
--- a/AirCasting/Graph/AirCastingGraph.swift
+++ b/AirCasting/Graph/AirCastingGraph.swift
@@ -57,7 +57,7 @@ class AirCastingGraph: UIView {
         
         lineChartView.highlightPerTapEnabled = false
         lineChartView.highlightPerDragEnabled = false
-     
+        #warning("It will be commented for the possibility of future use then - Think about it please")
 //        lineChartView.xAxisRenderer = TimeAxisRenderer(viewPortHandler: lineChartView.viewPortHandler,
 //                                                       xAxis: lineChartView.xAxis,
 //                                                       transformer: lineChartView.getTransformer(forAxis: .left))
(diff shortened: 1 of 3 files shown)


AirCasting/Graph/AirCastingGraph.swift
```diff
@@ -110,7 +110,7 @@ class AirCastingGraph: UIView {
         if !didMoveOrScaleGraph && isAutozoomEnabled {
             zoomoutToThirtyMinutes(dataSet: dataSet)
         }
-        callDateRangeChangeObserver()
+//        callDateRangeChangeObserver()
```
pawelgil, line 113: commenting those out will probably cause statistics container to stop working

AirCasting/Graph/AirCastingGraph.swift
```diff
@@ -51,16 +51,16 @@ class AirCastingGraph: UIView {
...
-        lineChartView.extraBottomOffset = 25
+        lineChartView.extraBottomOffset = 2
         
         lineChartView.highlightPerTapEnabled = false
         lineChartView.highlightPerDragEnabled = false
-        
-        lineChartView.xAxisRenderer = TimeAxisRenderer(viewPortHandler: lineChartView.viewPortHandler,
-                                                       xAxis: lineChartView.xAxis,
-                                                       transformer: lineChartView.getTransformer(forAxis: .left))
+        #warning("It will be commented for the possibility of future use then - Think about it please")
```
pawelgil, line 60: can we just remove it instead? We have git for coming back

AirCasting/Graph/GraphView.swift
```diff
@@ -36,35 +36,92 @@ struct GraphView<StatsViewModelType>: View where StatsViewModelType: StatisticsC
...
+    
+    var startTimeText: some View {
+        startTime()
+    }
+    
+    var endTimeText: some View {
+        endTime()
+    }
+    
+    func procceding(session: SessionEntity) -> Bool {
```
pawelgil, line 87: typo and I'd suggest making it `isProceeding`

AirCasting/Graph/GraphView.swift
```diff
@@ -36,35 +36,92 @@ struct GraphView<StatsViewModelType>: View where StatsViewModelType: StatisticsC
...
+        var showGraph = false
+        guard session.allStreams != nil else { return false }
+        (session.allStreams!.forEach({ stream in
```
pawelgil, line 90: most of the `(` and `)` are not needed here

AirCasting/Graph/GraphView.swift
```diff
@@ -36,35 +36,92 @@ struct GraphView<StatsViewModelType>: View where StatsViewModelType: StatisticsC
...
+            (stream.allMeasurements != []) ? (showGraph = true) : (showGraph = false)
```
pawelgil, line 91: `isEmpty` instead, also can be simplified to just `showGraph = !stream.allMeasurements.isEmpty`. 

But the main point - are you sure it will do the thing you want it to be doing? For me it doesn't make much sense to iterate over all streams when each one will override this `showGraph` value. It's basically equivalent to just doing something like that:
`showGraph = !(session.allStreams.last?.isEmpty ?? true)`

AirCasting/Graph/GraphView.swift
```diff
@@ -36,35 +36,92 @@ struct GraphView<StatsViewModelType>: View where StatsViewModelType: StatisticsC
...
+        }))
+        if showGraph {
```
pawelgil, line 93: `return showGraph`

AirCasting/Graph/GraphView.swift
```diff
@@ -36,35 +36,92 @@ struct GraphView<StatsViewModelType>: View where StatsViewModelType: StatisticsC
...
+                            .padding(.bottom)
+                    }
             }
             Spacer()
         }
         .navigationBarTitleDisplayMode(.inline)
     }
+    
+    var startTimeText: some View {
+        startTime()
```
pawelgil, line 80: if it only calls this func let's move the body of it here instead

AirCasting/Graph/GraphView.swift
```diff
@@ -36,35 +36,92 @@ struct GraphView<StatsViewModelType>: View where StatsViewModelType: StatisticsC
...
+    }
+    
+    var endTimeText: some View {
+        endTime()
```
pawelgil, line 84: if it only calls this func let's move the body of it here instead

AirCasting/Graph/GraphView.swift
```diff
@@ -36,35 +36,92 @@ struct GraphView<StatsViewModelType>: View where StatsViewModelType: StatisticsC
...
+        }))
+        if showGraph {
+            return true
+        } else {
+            return false
+        }
+    }
+    
+    func startTime() -> Text {
+            let formatter = DateFormatter()
```
pawelgil, line 101: creating a new `DateFormatter` is a CPU-heavy task, plase use other solution so you don't create a new formatter on each view reload

AirCasting/Graph/GraphView.swift
```diff
@@ -36,35 +36,92 @@ struct GraphView<StatsViewModelType>: View where StatsViewModelType: StatisticsC
...
+            formatter.locale = Locale(identifier: "en_US")
+        if !(session.isMobile && session.isActive) {
```
pawelgil, line 103: 1. This feels magical, I know why it's here but if someone sees it for the first time they will have absolutely no clue why it's like that and will probably not be able to figure it out (so we're harming the flexibility to change the app, which is a core value of a software product). Please at least add a comment
2. Formatting seems off

AirCasting/Graph/GraphView.swift
```diff
@@ -36,35 +36,92 @@ struct GraphView<StatsViewModelType>: View where StatsViewModelType: StatisticsC
...
+            formatter.timeZone =  TimeZone.init(abbreviation: "UTC")
+        }
+            formatter.dateFormat = "HH:mm"
+            guard let start = session.startTime else { return Text("") }
+     
+            let string = formatter.string(from: start)
+            return Text(string)
+        }
+    
+    func endTime() -> Text {
```
pawelgil, line 113: in this case repeated parts of `startTime()` should be extracted and reused instead of copy/pasted

pawelgil (commented):
I know it's merged already, but I think some stuff needs to be addressed

AirCasting/Graph/GraphView.swift
jankrzempek: Doing this mostly like it is, with some changes from you results in not crashing when entering graph before measurements are loaded so I will change a bit

PR #263, opened 2021-09-27
Remove padding to center measurements names with dots and values

anna1901:
Removing the padding on text in singleMeasurementView fixes an issue with measurement values and dots not being aligned with the name.


jankrzempek (approved):
Cool!

PR #264, opened 2021-09-27
Remove sensor type and device name from session header in map and graph view

anna1901:


jankrzempek (approved):

AirCasting/SessionViews/SessionHeaderView.swift
```diff
@@ -10,6 +10,7 @@ import SwiftUI
 struct SessionHeaderView: View {
     let action: () -> Void
     let isExpandButtonNeeded: Bool
+    var isSensorTypeNeeded: Bool = true
```
pawelgil, line 13: I think this can be a `let`, and there is probably no need for a default value

PR #270, opened 2021-09-28
Syncing measurements for dormant sessions - fixes

smialko:
Tbh I thought this will be just a first step of fixing the 0-values on dormant sessions but it wasn't. Also, removing save from `addMeasurementValues` improved the performance of saving measurements.

### DIFF
--- a/AirCasting/CoreData/MeasurementStreamStorage.swift
+++ b/AirCasting/CoreData/MeasurementStreamStorage.swift
@@ -26,7 +26,6 @@ protocol MeasurementStreamStorageContextUpdate {
 extension HiddenCoreDataMeasurementStreamStorage {
     func addMeasurementValue(_ value: Double, at location: CLLocationCoordinate2D? = nil, toStreamWithID id: MeasurementStreamLocalID) throws {
         try addMeasurement(Measurement(time: Date(), value: value, location: location), toStreamWithID: id)
-        try save()
     }
 }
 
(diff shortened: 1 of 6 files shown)


jankrzempek (approved):
👌🏻

AirCasting/CoreData/MeasurementStreamStorage.swift
```diff
@@ -26,7 +26,6 @@ protocol MeasurementStreamStorageContextUpdate {
 extension HiddenCoreDataMeasurementStreamStorage {
     func addMeasurementValue(_ value: Double, at location: CLLocationCoordinate2D? = nil, toStreamWithID id: MeasurementStreamLocalID) throws {
         try addMeasurement(Measurement(time: Date(), value: value, location: location), toStreamWithID: id)
-        try save()
```
anna1901, line 29: When does it gets saved now when `addMeasurementValue` is called in `SyncingMeasurementsViewModel` ?

smialko: Each method from `HiddenCoreDataMeasurementStreamStorage` including this one must be called within `accessStorage` method:
```
    func accessStorage(_ task: @escaping(HiddenCoreDataMeasurementStreamStorage) -> Void) {
        context.perform {
            task(self.hiddenStorage)
            try? self.hiddenStorage.save()
        }
```
This way we don't have to remember to call `save()` :)

anna1901: Awesome! Does it mean we could remove save() from `MobilePeripheralSessionManager` and `MicrophoneManager` as well?

smialko: Good catch! 🥷🏻

smialko:
@pawelgil  I'm merging it for Michael to test but please review.

pawelgil (commented):
LGTM

PR #275, opened 2021-09-29
Round time interval to seconds to fix the chart

anna1901:
Rounding beginning and end of each time interval seems to fix the problem with "dancing chart" (varying number of measurements included in the time interval).

### DIFF
--- a/AirCasting/Extensions/Date+Miliseconds.swift
+++ b/AirCasting/Extensions/Date+Miliseconds.swift
@@ -24,4 +24,10 @@ extension Date {
         dateFormatter.dateFormat = "yyyy-MM-dd HH:mm:ss"
         return dateFormatter.date(from: stringDate)!
     }
+
+    var roundedToSecond: Date {
+        let date = self
+        let diff = 1000000000 - Calendar.current.component(.nanosecond, from: date)
+        return Calendar.current.date(byAdding: .nanosecond, value: diff, to: date)!
+    }
 }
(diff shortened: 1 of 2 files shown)


AirCasting/SessionViews/ChartEntriesCreator.swift
```diff
@@ -40,8 +40,17 @@ final class ChartEntriesCreator {
...
 }
+
+extension Date {
+    var roundedToSecond: Date {
+        let date = self
+        let diff = 1000000000 - Calendar.current.component(.nanosecond, from: date)
+        return Calendar.current.date(byAdding: .nanosecond, value: diff, to: date)!
+    }
+}
+
```
jankrzempek, line 56: Could you move this to the extensions folder, please?

AirCasting/Extensions/Date+Miliseconds.swift
```diff
@@ -24,4 +24,10 @@ extension Date {
         dateFormatter.dateFormat = "yyyy-MM-dd HH:mm:ss"
         return dateFormatter.date(from: stringDate)!
     }
+
+    var roundedToSecond: Date {
```
pawelgil, line 28: After thinking about this for a bit I think if we just need to round seconds we can do it in one line:
```Swift
Date(timeIntervalSinceReferenceDate: timeIntervalSinceReferenceDate.round())
```
(idk if naming is correct, was typing from memory)

pawelgil (approved):

jankrzempek (approved):

PR #278, opened 2021-10-01
making container smaller and numbers formatting

jankrzempek:
https://trello.com/c/fbWq2j1G/275-stats-container-formatting

What was done here:
The number goes to the center
The container is now moved to the corner and the background is a bit smaller

### DIFF
--- a/AirCasting/SessionViews/StatisticsContainerView.swift
+++ b/AirCasting/SessionViews/StatisticsContainerView.swift
@@ -26,10 +26,11 @@ struct StatisticsContainerView<ViewModelType>: View where ViewModelType: Statist
         }
         .font(Font.muli(size: 12))
         .foregroundColor(.aircastingGray)
-        .frame(width: 260, height: 85)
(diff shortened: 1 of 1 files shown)


AirCasting/SessionViews/StatisticsContainerView.swift
```diff
@@ -26,10 +26,11 @@ struct StatisticsContainerView<ViewModelType>: View where ViewModelType: Statist
         }
         .font(Font.muli(size: 12))
         .foregroundColor(.aircastingGray)
-        .frame(width: 260, height: 85)
+        .frame(width: 220, height: 80)
         .background(Color.white)
         .cornerRadius(8)
-        .padding()
+        .padding(.leading, 10)
+        .padding(.top, 10)
```
pawelgil, line 33: padding takes a set as first arg so you can do:

```Swift
.padding([.leading, .top], 10)
```

pawelgil (approved):

PR #279, opened 2021-10-01
Setup timer for refreshing Chart View

anna1901:
We are adding a chartViewModel which is responsible for calculating averages for measurements in given time interval (1 minute or 1 hour).
We are doing that by setting up a timer which calculates averages each minute (or hour).
There is also a second timer which calculates a time for a full minute (or hour) since the session started in case the chart view was initialised when the session was already in progress.


AirCasting/SessionViews/ChartViewModel.swift
```diff
@@ -0,0 +1,72 @@
...
+    
+    private func generateEntries() {
+        guard var intervalEnd = stream.lastMeasurementTime else {
+            return
+        }
+        var intervalStart = intervalEnd - timeUnit
+        
+        var entries = [ChartDataEntry]()
+        
+        for i in (0...8).reversed() {
```
jankrzempek, line 54: Could you please use here 'index' or 'number' instead of 'i'? I think it would be nicer to look at and could tell us more

AirCasting/SessionViews/ChartViewModel.swift
```diff
@@ -0,0 +1,72 @@
...
+            intervalStart = intervalEnd - timeUnit
+        }
+        
+        self.entries = entries
+    }
+    
+    private func averagedValue(_ intervalStart: Date, _ intervalEnd: Date) -> Double? {
+        let measurements = stream.getMeasurementsFromTimeRange(intervalStart.roundedToSecond, intervalEnd.roundedToSecond)
+        let values = measurements.map { $0.value}
+        return values.isEmpty ? nil : round(values.reduce(0, +)/Double(values.count))
```
jankrzempek, line 70: What do you think of space here  `return values.isEmpty ? nil : round(values.reduce(0, +) / Double(values.count))` ?

AirCasting/SessionViews/ChartViewModel.swift
```diff
@@ -0,0 +1,72 @@
...
+import Foundation
+import Charts
+
+final class ChartViewModel: ObservableObject {
+    @Published var entries: [ChartDataEntry] = []
+    
+    private let stream: MeasurementStreamEntity
+
+    private var timeUnit: Double {
+        stream.session.type == .mobile ? 60 : 60*60
```
jankrzempek, line 13: Preference I think: maybe 60 * 60 or (60 * 60) here 😅?

jankrzempek (approved):
Nice 👌🏻

AirCasting/SessionViews/ChartViewModel.swift
```diff
@@ -0,0 +1,72 @@
...
+    
+    init(stream: MeasurementStreamEntity) {
+        self.stream = stream
+        generateEntries()
+        startTimers(stream.session)
+    }
+    
+    private func startTimers(_ session: SessionEntity) {
+        let sessionStartTime = session.startTime!
+        let secondsUntilFullMinute = timeUnit - Double(Int(Date().timeIntervalSince(sessionStartTime)) % Int(timeUnit))
```
mturnau, line 32: this name only makes sense for mobile sessions.
Also, it's hard to understand from the code what is happening here (why are we calculating this). Maybe some comment?..

AirCasting/SessionViews/ChartViewModel.swift
```diff
@@ -0,0 +1,96 @@
...
+    
+    func startMainTimer() {
+        mainTimer = Timer.scheduledTimer(withTimeInterval: timeUnit, repeats: true) { [weak self] timer in
+            self?.generateEntries()
+        }
+    }
+    
+    private func generateEntries() {
+        // Set up begning and end of the interval for the first average
+        //  - for fixed sessions we are taking the last full hour of the session, which has any measurements
```
mturnau, line 47: this is not accurate. For fixed sessions we should take the last 9 hours of the session. E.g. if session starts at 12:15 and now is 21:05, we'll have averages from:
12:15-12:59 (12)
13:00-13:59 (13)
...
20:00-20:59 (21)

AirCasting/SessionViews/ChartViewModel.swift
```diff
@@ -0,0 +1,96 @@
...
+import Foundation
+import Charts
+
+final class ChartViewModel: ObservableObject {
+    @Published var entries: [ChartDataEntry] = []
+    
+    private let stream: MeasurementStreamEntity
+
+    private var timeUnit: Double {
+        stream.session.type == .mobile ? 60 : 60 * 60
```
mturnau, line 13: I personally prefer to have any numeric constants named. We know 60 is a minute and 60*60 is an hour but it is not visible from the first sight. Also, should we ever have to deal with other intervals, it will be harder to alter

AirCasting/SessionViews/ChartViewModel.swift
anna1901: That's how it works (we are taking the last full hour). Maybe the wording in the comment is not accurate. Do you have an idea how to rephrase that? You can see in endTime function that we take `lastMeasurementTime.roundedDownToHour` as the end of the last interval (so in the example you hade it would be 21:05 rounded down to 21.

AirCasting/SessionViews/ChartViewModel.swift
```diff
@@ -0,0 +1,96 @@
...
+        } else {
+            let secondsSinceFullMinuteFromSessionStart = Double(Int(Date().timeIntervalSince(sessionStartTime)) % Int(timeUnit))
+            return lastMeasurementTime - secondsSinceFullMinuteFromSessionStart
+        }
+    }
+    
+    private func timeOfNextAverage() -> Double {
+        let sessionStartTime = stream.session.startTime!
+        
+        if stream.session.type == .fixed {
```
smialko, line 84: Maybe you could use `if session.isFixed` here? :-)

AirCasting/SessionViews/ChartViewModel.swift
```diff
@@ -0,0 +1,96 @@
...
+        }
+        
+        self.entries = entries
+    }
+    
+    private func endTime() -> Date? {
+        guard let lastMeasurementTime = stream.lastMeasurementTime else { return nil }
+        let sessionStartTime = stream.session.startTime!
+        
+        if stream.session.type == .fixed {
```
smialko, line 73: same as below, what do you think about `if session.isFixed`?

smialko (approved):

AirCasting/SessionViews/ChartViewModel.swift
mturnau: as for Janek's comment, I think i is generally used as an index var name so it's clear enough. However, I'd consider not hardcoding 8 here - make it a constant, so the chart is easily adjustable.

AirCasting/SessionViews/ChartViewModel.swift
```diff
@@ -0,0 +1,96 @@
...
+                entries.append(ChartDataEntry(x: Double(i), y: average))
+            }
+            intervalEnd = intervalStart
+            intervalStart = intervalEnd - timeUnit
+        }
+        
+        self.entries = entries
+    }
+    
+    private func endTime() -> Date? {
```
mturnau, line 69: I'd name it intervalEndTime() - endTime name may be confused with session endTime

AirCasting/SessionViews/ChartView.swift
```diff
@@ -52,26 +52,30 @@ class UI_PollutionChart: UIView {
...
-        if entries.isEmpty {
-            return()
-        } else {
-            entries.sort { (e1, e2) -> Bool in
-                e1.x < e2.x
-            }
+        var entries = chartCreator.entries
+        
+        entries.sort { (e1, e2) -> Bool in
+            e1.x < e2.x
```
pawelgil, line 78: This is a pretty common thing to do (sort over some property of an object), how about we make it nice and clean?

I've come up with a simple set of extensions that will allow us to do just:
```Swift
entries.sort(by: \.x, areInIncreasingOrder: <)
```
which I think is pretty neat. Just use my code:
```Swift
extension Sequence {
    func sorted<T: Comparable>(by keyPath: KeyPath<Element, T>, areInIncreasingOrder: (T, T) throws -> Bool) rethrows -> [Element] {
        try sorted {
            try areInIncreasingOrder($0[keyPath: keyPath], $1[keyPath: keyPath])
        }
    }
}

extension Array {
    mutating func sort<T: Comparable>(by keyPath: KeyPath<Element, T>, areInIncreasingOrder: (T, T) throws -> Bool) rethrows {
        self = try sorted(by: keyPath, areInIncreasingOrder: areInIncreasingOrder)
    }
}
```

AirCasting/SessionViews/ChartViewModel.swift
pawelgil: Since we're dealing with those across the whole codebase maybe we can make an extension on TimeInterval with those constants instead of just local ones?
```Swift
extension TimeInterval {
    static let minute: TimeInterval = 60
    static let hour: TimeInterval = minute * 60
    static let day: TimeInterval = hour * 24
    //etc, as needed
}
```

anna1901: I like this idea

AirCasting/SessionViews/ChartView.swift
anna1901: I personally think it makes it a bit more confusing, so I'd leave it as is.

pawelgil: Consider creating something on top of this general extensions like:
```Swift
entries.sort(by: \.x, ascending: true/false)
```

or even with default value it can get to this point:
```Swift
entries.sort(by: \.x)
```
I think it's much better

Also - leaving readability and code clarity aside - note that by using keypaths you'd also eliminate a whole class of typo errors, for example this one:
```Swift
cars.sort {
   $0.engineCapacity < $0.engineCylinders
}
```
won't be possible.

But I'm not forcing it that hard - if you feel it's not right - just leave it :)

AirCasting/SessionViews/ChartView.swift
```diff
@@ -52,26 +52,30 @@ class UI_PollutionChart: UIView {
 }
 
 struct ChartView: UIViewRepresentable {
-    
-    @ObservedObject var stream: MeasurementStreamEntity
-    var thresholds: [SensorThreshold]
+    private let stream: MeasurementStreamEntity
+    private let thresholds: [SensorThreshold]
 
+    @StateObject var chartCreator: ChartViewModel
```
smialko, line 58: We shouldn't be using @StateObject like this, it's a source of truth and should be created in one place. 

1. SwiftBySundell: 
"Although it’s technically possible to only use StateObject from now on — I still recommend using ObservedObject when observing external objects, and to **only use StateObject when dealing with objects that are owned by a view itself**".  
https://www.swiftbysundell.com/articles/swiftui-state-management-guide/

2. Apple's developer comment on a similar thing: 
A @StateObject is only meant for instantiating an object at some point in your view hierarchy, as you've done with presenterStateWorking. This attribute tells SwiftUI to create the object only once, regardless of how many times it recreates the enclosing view. **It isn't meant for passing an object through the interface**, as you've done with presenterStateWorking(I think he meant presenterStateNotWorking). For that, use @ObservedObject, as you have with presenterObserved."
https://developer.apple.com/forums/thread/650618

Adding this quotes because we already had this conversation with Paweł and I wanted to support my words by some experts opinions ;D

smialko: I know I already reviewed&accepted this PR and didn't add this comment at that time. I missed this line of code and I've just remembered you talked about @StateObject and wanted to take a look at how it's done :)

anna1901: This chartViewModel is very tightly associated with chartView (i.e, it's creating an array of [ChartDataEntries]. The logic is much better when we are creating it inside the view (and it belongs just to this view and is not passed through interface, which would be a problem according to quote 2). We can say that this class is owned by the view itself. In the current setup it is created only in one place. So I'm not convinced by those quotes that we should change that. However, we can discuss it further. Let's see what @pawelgil thinks

smialko: https://www.google.com/url?sa=i&url=https%3A%2F%2Ftenor.com%2Fview%2Fthumbs-up-nice-friends-joey-tribbiani-matt-le-blanc-gif-15167765&psig=AOvVaw3NOWy7bQ0sIh0iApcCLUqh&ust=1633595830759000&source=images&cd=vfe&ved=0CAsQjRxqFwoTCPig8eywtfMCFQAAAAAdAAAAABBE

smialko: image

AirCasting/SessionViews/ChartView.swift
anna1901: I tried implementing it, but it didn't work. However, this: `areInIncreasingOrder: <` is a bit confusing to me, but I guess when I want it ti be in the reversed order (descreasing) I should do it like: `areInIncreasingOrder: >`? Anyway, I pushed it to a different [branch](https://github.com/HabitatMap/AirCastingiOS/tree/new-sorting) so you can check what I did wrong. For now, I'm reverting it to the previous version.

PR #290, opened 2021-10-07
fix measurement time to proper one 

jankrzempek:
For now, there was a difference in the graph between android and iOS. The reason was the measurement time was equal always to = Date() and not the measurement. time as it should be. Now the graph is looking cool I think

https://trello.com/c/2VEszn7M/329-mobile-dormant-graph-inaccurate-missing-data

### DIFF
--- a/AirCasting/CoreData/MeasurementStreamStorage.swift
+++ b/AirCasting/CoreData/MeasurementStreamStorage.swift
@@ -24,8 +24,8 @@ protocol MeasurementStreamStorageContextUpdate {
 }
 
 extension HiddenCoreDataMeasurementStreamStorage {
-    func addMeasurementValue(_ value: Double, at location: CLLocationCoordinate2D? = nil, toStreamWithID id: MeasurementStreamLocalID) throws {
-        try addMeasurement(Measurement(time: Date(), value: value, location: location), toStreamWithID: id)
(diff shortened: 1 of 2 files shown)


AirCasting/CoreData/MeasurementStreamStorage.swift
```diff
@@ -24,8 +24,8 @@ protocol MeasurementStreamStorageContextUpdate {
 }
 
 extension HiddenCoreDataMeasurementStreamStorage {
-    func addMeasurementValue(_ value: Double, at location: CLLocationCoordinate2D? = nil, toStreamWithID id: MeasurementStreamLocalID) throws {
-        try addMeasurement(Measurement(time: Date(), value: value, location: location), toStreamWithID: id)
+    func addMeasurementValue(_ value: Double, at location: CLLocationCoordinate2D? = nil, toStreamWithID id: MeasurementStreamLocalID, on time: Date? = Date()) throws {
+        try addMeasurement(Measurement(time: time!, value: value, location: location), toStreamWithID: id)
```
pawelgil, line 28: wouldn't it be better to not make the argument an optional in the first place?

anna1901: I agree. Is there a reason for making it optional?

jankrzempek: There is no, you are right

PR #291, opened 2021-10-07
Speed up fetching of measurements

anna1901:
Previously we were getting streamID from database for each measurement. Now we are getting it once per stream.


AirCasting/SessionViews/SessionCartViewModel/DefaultSyncingMeasurementsViewModel.swift
```diff
@@ -49,25 +48,31 @@ final class DefaultSyncingMeasurementsViewModel: SyncingMeasurementsViewModel {
...
+                            stream.measurements.forEach { measurement in
+                                let location: CLLocationCoordinate2D? = {
+                                    guard let latitude = measurement.latitude,
+                                          let longitude = measurement.longitude else { return CLLocationCoordinate2D(latitude: 200, longitude: 200) }
+                                    return CLLocationCoordinate2D(latitude: latitude, longitude: longitude)
+                                }()
+                                do {
+                                    guard let streamID = streamID else {
+                                        return
+                                    }
```
jankrzempek, line 63: can it be in one line please `else { return }`

PR #298, opened 2021-10-11
Fix - midnight lines on graph

jankrzempek:
In this ticket, I've added a day to the current day when looking for midnight points. It is a small hack that will allow us to consider each day when looking for midnight. It is needed especially on the existing, recording sessions.

https://trello.com/c/qHtJtfC6/273-all-tabs-graph-show-midnight-lines

### DIFF
--- a/AirCasting/Graph/Graph.swift
+++ b/AirCasting/Graph/Graph.swift
@@ -79,22 +79,25 @@ struct Graph: UIViewRepresentable {
     }
     
     func getMidnightsPoints(startingDate: Date) -> [Double] {
-        let now = Date()
+        let day = Date()
+        // [SMALL HACK] - By adding a day to the current Date we are ensuring
+        // that session which is currently recording will be able to show midnight line
+        let now = Calendar.current.date(byAdding: .day, value: 1, to: day)!
         var midnight = Calendar.current.date(bySettingHour: 0, minute: 0, second: 0, of: now)!
         let components = Calendar.current.dateComponents([.hour, .minute, .second], from: midnight)
         let firstMeasurementDate = startingDate
         var midnightPoints: [Double] = []
         
         while (midnight > firstMeasurementDate) {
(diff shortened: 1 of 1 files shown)


AirCasting/Graph/Graph.swift
```diff
@@ -79,22 +79,25 @@ struct Graph: UIViewRepresentable {
     }
     
     func getMidnightsPoints(startingDate: Date) -> [Double] {
-        let now = Date()
+        let day = Date()
+        // [SMALL HACK] - By adding a day to the current Date we are ensuring
+        // that session which is currently recording will be able to show midnight line
+        let now = Calendar.current.date(byAdding: .day, value: 1, to: day)!
```
smialko, line 85: I think this shouldn't be named "now" - if I understand it correctly it's more like `nextDay`?

anna1901, line 85: I think the name is misleading. I propose `nextDay`

AirCasting/Graph/Graph.swift
```diff
@@ -79,22 +79,25 @@ struct Graph: UIViewRepresentable {
     }
     
     func getMidnightsPoints(startingDate: Date) -> [Double] {
-        let now = Date()
+        let day = Date()
```
anna1901, line 82: Should this always be `Date()`? When we are looking at the session which finished eg. in 2019 we are calculating all of the midnights since the first measurements, and that can be a lot. Do you think it would be possible to pass the lastMeasurementDate to this function?

AirCasting/Graph/Graph.swift
jankrzempek: yeah, definitely I will change it

AirCasting/Graph/Graph.swift
jankrzempek: Cool, haven't thought about it but yeees. I will try if it is gonna work - if will I change a way of making it

PR #300, opened 2021-10-11
Proper timezones 

mturnau:
https://trello.com/c/Q90tAWd1/337-time-issues

The purpose of this change is to always save all of the times (session startTime, session endTime and measurement time) in local time as UTC. This is how it is done on other API clients and it makes  easiest to display times well on all clients.
Example: it's 7am in New York and a user starts recording an AB session. Both session start/end times and all measurements times will be saved as 6:00 UTC. If a user in Poland starts session at ecatly same time, this session times will be saved as 13:00 UTC. It is not necessarily intuitive but proved to work well.

Times on Graph

- [x] Mobile mic active
- [x] Mobile ab active
- [x] Mobile mic dormant from iOS
- [x] Mobile ab dormant from iOS
- [ ] Mobile mic dormant from android
- [ ] Mobile ab dormant from android
- [x] Fixed new (before any measurements come)
- [x] Fixed with measurements
- [ ] Fixed from android
- [x] Measurements mic on backend (test on server)
- [x] Measurements ab on backend (test on server)
- [x] Measurements mic -> Web
- [x] Measurements ab -> Web

Times on Chart

- [x] Mobile mic active
- [x] Mobile ab active
- [x] Mobile mic dormant from iOS
- [x] Mobile ab dormant from iOS
- [ ] Mobile mic dormant from android
- [ ] Mobile ab dormant from android
- [x] Following new (before any measurements come)
- [x] Following with measurements
- [ ] Following from android

### DIFF
--- a/AirCasting/CoreData/MeasurementStreamStorage.swift
+++ b/AirCasting/CoreData/MeasurementStreamStorage.swift
@@ -187,7 +187,7 @@ final class HiddenCoreDataMeasurementStreamStorage: MeasurementStreamStorageCont
         do {
             let sessionEntity = try context.existingSession(uuid: sessionUUID)
             if sessionFollowing == SessionFollowing.following {
-                sessionEntity.followedAt = Date()
+                sessionEntity.followedAt = Date().currentUTCTimeZoneDate
             } else {
                 sessionEntity.followedAt = nil
             }
--- a/AirCasting/MobilePeripheralSessionManager.swift
+++ b/AirCasting/MobilePeripheralSessionManager.swift
@@ -55,7 +55,7 @@ class MobilePeripheralSessionManager {
             measurementStreamStorage.accessStorage { storage in
                 do {
                     try storage.updateSessionStatus(.FINISHED, for: session.uuid)
-                    try storage.updateSessionEndtime(Date().currentUTCTimeZoneDate, for: session.uuid)
+                    try storage.updateSessionEndtime(Date(), for: session.uuid)
                 } catch {
                     Log.error("Unable to change session status to finished because of an error: \(error)")
                 }
--- a/AirCasting/SessionStopping/MicrophoneSessionStopper.swift
+++ b/AirCasting/SessionStopping/MicrophoneSessionStopper.swift
@@ -19,7 +19,7 @@ class MicrophoneSessionStopper: SessionStoppable {
         try microphoneManager.stopRecording()
         measurementStreamStorage.accessStorage { [self] storage in
             do {
-                try storage.updateSessionEndtime(Date().currentUTCTimeZoneDate, for: self.uuid)
+                try storage.updateSessionEndtime(Date(), for: self.uuid)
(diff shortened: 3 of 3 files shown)


AirCasting/ABConnector/HexMessagesBuilder.swift
```diff
@@ -13,7 +13,12 @@ struct HexMessagesBuilder {
         func getTimezoneOffsetInHours() -> Int {
             let calendar = Calendar.current
             let timeZone = calendar.timeZone
-            let hoursOffset = Int(timeZone.secondsFromGMT() / 3600)
+            var offset = timeZone.secondsFromGMT()
+            // We need to send *standard* offset, ignoring DST because AirBeam is calculating DST anyway
+            if (timeZone.isDaylightSavingTime()) {
+                offset = offset - Int(timeZone.daylightSavingTimeOffset())
+            }
+            let hoursOffset = Int(offset / 3600)
```
pawelgil, line 21: I'm not sure getting timezone offset in integer hours is the correct way. Not all timezones are nicely rounded to full hour, for example Indian standard is +5:30 or Nepal with its +5:45.

AirCasting/ABConnector/HexMessagesBuilder.swift
```diff
@@ -13,7 +13,12 @@ struct HexMessagesBuilder {
         func getTimezoneOffsetInHours() -> Int {
             let calendar = Calendar.current
             let timeZone = calendar.timeZone
-            let hoursOffset = Int(timeZone.secondsFromGMT() / 3600)
+            var offset = timeZone.secondsFromGMT()
+            // We need to send *standard* offset, ignoring DST because AirBeam is calculating DST anyway
+            if (timeZone.isDaylightSavingTime()) {
+                offset = offset - Int(timeZone.daylightSavingTimeOffset())
```
pawelgil, line 19: `-=`

AirCasting/ABConnector/HexMessagesBuilder.swift
```diff
@@ -13,7 +13,12 @@ struct HexMessagesBuilder {
         func getTimezoneOffsetInHours() -> Int {
             let calendar = Calendar.current
             let timeZone = calendar.timeZone
-            let hoursOffset = Int(timeZone.secondsFromGMT() / 3600)
+            var offset = timeZone.secondsFromGMT()
+            // We need to send *standard* offset, ignoring DST because AirBeam is calculating DST anyway
+            if (timeZone.isDaylightSavingTime()) {
```
pawelgil, line 18: no `(`  `)` needed i Swift

AirCasting/ABConnector/HexMessagesBuilder.swift
mturnau: I agree the way we configure AirBeam and handle TZ is daft. However this is what AirBeam expects (hour offset as Int) and there is not so much we can do about it. I guess there are no AB users in weird timezones...yet? ;) I'll poit it out to Michael though

PR #302, opened 2021-10-12
Fix time and label below the chart

anna1901:


AirCasting/SessionViews/ChartViewModel.swift
```diff
@@ -76,11 +77,11 @@ final class ChartViewModel: ObservableObject {
             if let average = average {
                 entries.append(ChartDataEntry(x: Double(i), y: average))
             }
+            chartStartTime = intervalEnd
```
pawelgil, line 80: `chartStartTime` is a `@Published` property - are you sure we want to update it that often?

PR #303, opened 2021-10-12
Start timers for chart for new sessions and refresh chart on appear

anna1901:
This PR fixes issue with:
- chart not appearing for followed session (only for the first followed session) until the stream was selected
- chart not refreshing for mobile AB sessions

### DIFF
--- a/AirCasting/SessionViews/ChartViewModel.swift
+++ b/AirCasting/SessionViews/ChartViewModel.swift
@@ -34,7 +34,7 @@ final class ChartViewModel: ObservableObject {
         self.session = session
         self.chartStartTime = session.endTime
         self.chartEndTime = session.endTime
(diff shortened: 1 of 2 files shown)


AirCasting/SessionViews/ChartViewModel.swift
```diff
@@ -34,7 +34,7 @@ final class ChartViewModel: ObservableObject {
         self.session = session
         self.chartStartTime = session.endTime
         self.chartEndTime = session.endTime
-        if session.isActive || session.isFollowed {
+        if session.isActive || session.isFollowed || session.status == .NEW {
```
pawelgil, line 37: a bit weird that `isActive` doesn't pick up `.NEW` sessions

PR #308, opened 2021-10-13
Change map and graph views to full screen covers

anna1901:
When map and graph are displayed as full screen covers we fix two issues:
* the bottom menu bar is not visible: https://trello.com/c/LwLqY99w
* we can use custom back button: https://trello.com/c/cVR4mUtq

### DIFF
--- a/AirCasting/SessionViews/SessionCartView.swift
+++ b/AirCasting/SessionViews/SessionCartView.swift
@@ -62,10 +62,58 @@ struct SessionCartView: View {
     }
     
     var body: some View {
+        if #available(iOS 15, *) {
+            sessionCard
+            .fullScreenCover(isPresented: $isGraphButtonActive) {
+                GraphView(session: session,
+                          thresholds: thresholds,
+                          selectedStream: $selectedStream,
+                          statsContainerViewModel: graphStatsViewModel,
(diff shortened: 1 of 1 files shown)


jankrzempek (approved):
Working super 🥳

AirCasting/Graph/GraphView.swift
```diff
@@ -18,12 +20,16 @@ struct GraphView<StatsViewModelType>: View where StatsViewModelType: StatisticsC
...
     var body: some View {
         VStack(alignment: .trailing) {
-            SessionHeaderView(action: {},
-                              isExpandButtonNeeded: false,
-                              isSensorTypeNeeded: false,
-                              isCollapsed: Binding.constant(false),
-                              session: session,
-                              sessionStopperFactory: sessionStoppableFactory).padding()
+            HStack {
+                backButton
```
smialko, line 24: I know it's now the same as Android's but it doesn't feel iOSy. 
Maybe we could ask Michael if he's ok with moving the `backButton` to the top? 
I just think it's really important to follow Apple's design principles so I'd like to try to convince Michael.

anna1901: Sure! :) Can we merge this and move it to the top in another PR?

PR #315, opened 2021-10-14
Shadow and card fix

jankrzempek:
https://trello.com/c/4RrL5LWV/333-dashboard-all-tabs-reduce-vertical-space-between-cards-so-it-matches-the-android-app

-reduce space between cards

-make it so cards move off screen when they meet the line below the tab titles up top and the line above the menu buttons on bottom. currently they move off screen before they should.

To make this happen in nice way, I have added shadows with help from @smialko - Thank U!

### DIFF
--- a/AirCasting/MainTabBarView.swift
+++ b/AirCasting/MainTabBarView.swift
@@ -41,7 +41,7 @@ struct MainTabBarView: View {
             UITabBar.appearance().backgroundColor = .systemBackground
             let appearance = UITabBarAppearance()
             appearance.backgroundImage = UIImage()
-            appearance.shadowImage = UIImage.MainTabBarShadow
+            appearance.shadowImage = UIImage.mainTabBarShadow
             UITabBar.appearance().standardAppearance = appearance
             measurementUpdatingService.start()
         }
--- a/AirCasting/Utils/ShadowImages.swift
+++ b/AirCasting/Utils/ShadowImages.swift
@@ -6,7 +6,7 @@ import UIKit
 import SwiftUI
 
 extension UIImage {
-    static let MainTabBarShadow = UIImage.gradientImageWithBounds(
+    static let mainTabBarShadow = UIImage.gradientImageWithBounds(
         bounds: CGRect(x: 0, y: 0, width: UIScreen.main.scale, height: 5),
         colors: [
             UIColor.clear.cgColor,


AirCasting/MainTabBarView.swift
```diff
@@ -38,6 +38,13 @@ struct MainTabBarView: View {
             settingsTab
         }
         .onAppear {
+            UITabBar.appearance().backgroundColor = .systemBackground
+
+            let appearance = UITabBarAppearance()
+            appearance.backgroundImage = UIImage()
+            appearance.shadowImage = UIImage.MainTabBarShadow
+
```
smialko, line 46: do we want this spaces? :D  (this and in the 42nd line?)

AirCasting/Dashboard/DashboardView.swift
```diff
@@ -30,16 +30,16 @@ struct DashboardView: View {
...
             // When user pops back to this view using navigation the `large` title is displayed incorrectly.
             // As a workaround I`ve put a 1px rectangle between ScrollView and top. It seems to be doing the trick.
             //
             // Bug report was filled with Apple
             PreventCollapseView()
             AirSectionPickerView(selection: self.$selectedSection.selectedSection)
-                .padding(.leading, 8)
-
+                .zIndex(2)
+
```
smialko, line 42: Is this space needed? :D

AirCasting/Utils/ShadowImages.swift
```diff
@@ -0,0 +1,28 @@
...
+    static func gradientImageWithBounds(bounds: CGRect, colors: [CGColor]) -> UIImage {
+        let gradientLayer = CAGradientLayer()
+        gradientLayer.frame = bounds
+        gradientLayer.colors = colors
+        
+        UIGraphicsBeginImageContext(gradientLayer.bounds.size)
+        gradientLayer.render(in: UIGraphicsGetCurrentContext()!)
+        let image = UIGraphicsGetImageFromCurrentImageContext()
+        UIGraphicsEndImageContext()
+        return image!
```
smialko, line 26: I'm not a fan of it, lots of force unwrapping, are we sure it won't crash? Especially as it's called in `onAppear` of our main view 😬

jankrzempek: Yeah, you are right. I am gonna to change it to if let, it was the fastest way during 'first stage' of code

pawelgil: this is a pretty 50/50 scenario here. The reason it's optional is that those old functions like `UIGraphicsGetCurrentContext()` won't throw on error, they will return `nil` instead. But in this case `nil` value is actually an indication of error and here it doesn't look like we'll be having any errors ;) both `UIGraphicsGetCurrentContext` and `UIGraphicsGetImageFromCurrentImageContext` will probably be returning correct values all the time. IDK if propagating this optionality further is a good idea TBH

PR #323, opened 2021-10-17
Fonts moved to 'Fonts'

jankrzempek:
https://trello.com/c/DeId1WLV/358-refactor


anna1901 (approved):
Nice! 🙌  Looks like so much work

pawelgil (commented):
Nice work!

I was thinking if maybe we should structure it differently than on per-view basis? There is *a lot* of repeating fonts now, like for example `Fonts.AirBeamOnboarding.title` and `Fonts.ProviacyOnboarding.sheetTitle` are exactly the same font. How about we structure fonts like a text processor does:

enum Fonts {
   static let title1 = Font.moderate(size:28)
   static let boldTitle1 = Font.moderate(size:28, weight: .bold)

   static let heading1 = Font.....
   etc...
}

PR #325, opened 2021-10-18
Add 9h+ averaging and change editContext to be lazy var

smialko:
1. Changed editContext to be a lazy var. It's quick&easy fix to a bug that we have which is initializing MeeasurementStreamStorage in more than 1 place - by doing that we create more than one `editContext`. Then we might be changing an object within two contexts and cause conflicts. 
This fix is quick&easy but it's risky, too. It may cause some crashes or performance issues. 😬 
The best but requiring much more work would be creating just one MeeasurementStreamStorage and passing it. 

2. Added 9h+ averaging. 
I was testing how much time it takes to perform averaging and save data to CoreData. 

After 30 minutes the averaging started and it took 0.2 s:
pół godziny

Then avery 5 second it lasted for ~0.15 s:  
co 5s

After an hour I started reaveraging and it took: 0.1 s 
po 1h

And then every minute ~0.13 s: 
co minute

Screenshots show only one example of each averaging. It may fluctuate a bit, I tried to take a screenshot of the modal value.

### DIFF
--- a/AirCasting/Utils/AveragingService.swift
+++ b/AirCasting/Utils/AveragingService.swift
@@ -93,6 +93,7 @@ final class AveragingService: NSObject, ObservableObject {
                 
                 let chunkElementsCount = AveragingWindow.secondThresholdWindow.rawValue / AveragingWindow.firstThresholdWindow.rawValue
                 averagedWithFirstThreshold.chunks(ofCount: chunkElementsCount).forEach { measuremensInChunk in
+                    guard measuremensInChunk.count == chunkElementsCount else { return }
                     let averaged = self.averagedMeasurementFrom(chunk: measuremensInChunk, window: averagingWindow, measurementCount: chunkElementsCount)
                     averagedMeasurements.append(contentsOf: averaged)
                 }


jankrzempek (approved):
👌🏻✈️

AirCasting/CoreData/MeasurementStreamStorage.swift
```diff
@@ -49,7 +49,6 @@ final class CoreDataMeasurementStreamStorage: MeasurementStreamStorage {
         context.perform {
             task(self.hiddenStorage)
             try? self.hiddenStorage.save()
-        }
```
anna1901, line 52: I think this was deleted by mistake. The app doesn't build without this

anna1901 (changes requested):

AirCasting/Utils/AveragingService.swift
```diff
@@ -81,39 +78,45 @@ final class AveragingService: NSObject, ObservableObject {
...
     }
     
-    private func perform(sessionEntity: SessionEntity, averagingWindow: AveragingWindow) {
-        if let uuid = sessionEntity.uuid {
-            measurementStreamStorage.accessStorage { storage in
-                
-                let session = try? storage.getExistingSession(with: uuid)
+    private func perform(uuid: SessionUUID, averagingWindow: AveragingWindow, windowDidChange: Bool) {
+        measurementStreamStorage.accessStorage { storage in
+            let session = try? storage.getExistingSession(with: uuid)
```
pawelgil, line 83: Calling this function with an uuid of a session that cannot be retrieved from DB seems like a scenario we want to at least take a warning log of. Maybe let's do it instead of `try?`

AirCasting/Utils/AveragingService.swift
```diff
@@ -81,39 +78,45 @@ final class AveragingService: NSObject, ObservableObject {
...
+                    $0.averagingWindow == averagingWindow.rawValue
+                }
                 
-                session?.allStreams?.forEach { stream in
-                    var averagedMeasurements: [MeasurementEntity] = (stream.allMeasurements ?? []).filter {
-                        $0.averagingWindow != AveragingWindow.zeroWindow.rawValue
-                    }
-                    let unaveragedMeasurements: [MeasurementEntity] = (stream.allMeasurements ?? []).filter {
-                        $0.averagingWindow == AveragingWindow.zeroWindow.rawValue
+                /// Step 1 - perform averaging again of elements in the previous threshold
```
pawelgil, line 89: how about making separate, smaller functions out of those steps and name them nicely so it's easier to follow the general flow of this algorithm?

AirCasting/Utils/AveragingService.swift
```diff
@@ -122,35 +125,60 @@ final class AveragingService: NSObject, ObservableObject {
        return values.map({ $0.value }).reduce(0.0, +) / Double(values.count)
     }
     
+    private func averagedMeasurementFrom(chunk: ArraySlice<MeasurementEntity>, window: AveragingWindow, measurementCount: Int) -> [MeasurementEntity] {
+        if chunk.count == measurementCount {
+            /// https://github.com/apple/swift-algorithms/blob/main/Guides/Chunked.md
+            /// For integer types, any remainder of the division is discarded so we need to add 1 to elementsInChunk/2 to get the middle value.
+            let middleIndex = chunk.index(chunk.startIndex, offsetBy: (measurementCount/2 + 1))
+            let middleMeasurement = chunk[middleIndex]
+            middleMeasurement.value = self.calculateAvg(from: chunk)
```
pawelgil, line 134: `self.` is not needed

AirCasting/Utils/AveragingService.swift
```diff
@@ -122,35 +125,60 @@ final class AveragingService: NSObject, ObservableObject {
...
+            middleMeasurement.averagingWindow = window.rawValue
+            return [middleMeasurement]
+        } else {
+            return Array(chunk)
```
pawelgil, line 138: [chunk]

AirCasting/Utils/AveragingService.swift
```diff
@@ -122,35 +125,60 @@ final class AveragingService: NSObject, ObservableObject {
        return values.map({ $0.value }).reduce(0.0, +) / Double(values.count)
     }
     
+    private func averagedMeasurementFrom(chunk: ArraySlice<MeasurementEntity>, window: AveragingWindow, measurementCount: Int) -> [MeasurementEntity] {
+        if chunk.count == measurementCount {
```
pawelgil, line 129:
```Swift
guard chunk.count == measurementCount else { return [chunk] }
...
```

AirCasting/Utils/AveragingService.swift
```diff
@@ -122,35 +125,60 @@ final class AveragingService: NSObject, ObservableObject {
        return values.map({ $0.value }).reduce(0.0, +) / Double(values.count)
     }
     
+    private func averagedMeasurementFrom(chunk: ArraySlice<MeasurementEntity>, window: AveragingWindow, measurementCount: Int) -> [MeasurementEntity] {
```
pawelgil, line 128: maybe we should guard against a scenario where chunk.count is lower than measurementCount passed in?

AirCasting/Utils/AveragingService.swift
```diff
@@ -122,35 +125,60 @@ final class AveragingService: NSObject, ObservableObject {
...
+                    }
+                    self.perform(uuid: uuid, averagingWindow: window, windowDidChange: windowDidChange)
                 }
             }
         timers[session.uuid] = timer
     }
+    
+    private func averagingWindowFor(session: SessionEntity) -> AveragingWindow {
+        let sessionDuration = abs(session.startTime!.timeIntervalSinceNow)
+        let shouldUseFirstThreshold: Bool = sessionDuration <= TimeInterval(TimeThreshold.secondThreshold.rawValue)
```
pawelgil, line 179: `: Bool` unneeded

AirCasting/Utils/AveragingService.swift
```diff
@@ -122,35 +125,60 @@ final class AveragingService: NSObject, ObservableObject {
        return values.map({ $0.value }).reduce(0.0, +) / Double(values.count)
     }
     
+    private func averagedMeasurementFrom(chunk: ArraySlice<MeasurementEntity>, window: AveragingWindow, measurementCount: Int) -> [MeasurementEntity] {
+        if chunk.count == measurementCount {
+            /// https://github.com/apple/swift-algorithms/blob/main/Guides/Chunked.md
+            /// For integer types, any remainder of the division is discarded so we need to add 1 to elementsInChunk/2 to get the middle value.
+            let middleIndex = chunk.index(chunk.startIndex, offsetBy: (measurementCount/2 + 1))
```
anna1901, line 132: I think this should be `measurementCount/2 - 1` if we want a middle measurement cause 12/2 + 1 = 7 and the middle is 5

AirCasting/Utils/AveragingService.swift
```diff
@@ -81,39 +78,45 @@ final class AveragingService: NSObject, ObservableObject {
...
+                }
                 
-                session?.allStreams?.forEach { stream in
-                    var averagedMeasurements: [MeasurementEntity] = (stream.allMeasurements ?? []).filter {
-                        $0.averagingWindow != AveragingWindow.zeroWindow.rawValue
-                    }
-                    let unaveragedMeasurements: [MeasurementEntity] = (stream.allMeasurements ?? []).filter {
-                        $0.averagingWindow == AveragingWindow.zeroWindow.rawValue
+                /// Step 1 - perform averaging again of elements in the previous threshold
+                if windowDidChange && averagingWindow == .secondThresholdWindow {
```
anna1901, line 90: Seems like we never enter this and the measurements from previous threshold still get wiped out

AirCasting/Utils/AveragingService.swift
pawelgil: It's out of scope here naturally, but I think we shouldn't mix averaging into the `MeasurementEntity` itself. It should be separated using a new CD entity or even a new model. It's always better to have raw data separated from processed data both on DB level and on code level (fewer conflict points for both + easier maintenance).

AirCasting/Utils/AveragingService.swift
```diff
@@ -81,39 +78,45 @@ final class AveragingService: NSObject, ObservableObject {
         }
     }
     
-    private func perform(sessionEntity: SessionEntity, averagingWindow: AveragingWindow) {
-        if let uuid = sessionEntity.uuid {
-            measurementStreamStorage.accessStorage { storage in
-                
-                let session = try? storage.getExistingSession(with: uuid)
+    private func perform(uuid: SessionUUID, averagingWindow: AveragingWindow, windowDidChange: Bool) {
+        measurementStreamStorage.accessStorage { storage in
```
pawelgil, line 82: I don't know... The `accessStorage` block here is doing some heavy stuff and since we moved to a single-context (`editContext` being `lazy var`) it will block all other editing while being performed. I'd recommend avoiding this

AirCasting/Utils/AveragingService.swift
```diff
@@ -122,35 +125,60 @@ final class AveragingService: NSObject, ObservableObject {
...
+                        self.startPeriodicAveraging(session: session, window: checkWindow)
+                    }
+                    self.perform(uuid: uuid, averagingWindow: window, windowDidChange: windowDidChange)
                 }
             }
         timers[session.uuid] = timer
     }
+    
+    private func averagingWindowFor(session: SessionEntity) -> AveragingWindow {
+        let sessionDuration = abs(session.startTime!.timeIntervalSinceNow)
```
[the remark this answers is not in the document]
anna1901: It was when I set firstThreshold to 20 and second to 120

AirCasting/Utils/AveragingService.swift
smialko: Actually, it should be just `measurementCount/2` for odd numbers like `firstThresholdWindow` and what you said for even numbers like in the case of `secondThresholdWindow`. Thanks!

smialko: @mturnau when you have an even number of measurements in the chunk which one do you consider the middle measurement?

AirCasting/Utils/AveragingService.swift
anna1901: Seems like this could work:
`let sessionDuration = Date().currentUTCTimeZoneDate.timeIntervalSince(session.startTime!)`

AirCasting/Utils/AveragingService.swift
```diff
@@ -122,35 +125,60 @@ final class AveragingService: NSObject, ObservableObject {
...
             .autoconnect()
             .sink { [weak self] _ in
+                guard let self = self else { return }
                 context.perform {
                     guard let session = try? context.existingSession(uuid: uuid) else { return }
-                    self?.perform(sessionEntity: session, averagingWindow: .firstThresholdWindow)
+                    let checkWindow = self.averagingWindowFor(session: session)
+                    let windowDidChange = checkWindow != window
+                    if windowDidChange {
+                        self.startPeriodicAveraging(session: session, window: checkWindow)
```
anna1901, line 169: If windowDidChange, we are calling startPeriodicAveraging for the second time window and then windowDidChange is false, so when we call `perform` for second time window, the windowDidChange value is already false. That's why we don't calculate averaged values for already averaged measurements and they get wiped out

AirCasting/Utils/AveragingService.swift
pawelgil: btw. note that this parameter is the offset, not an index. So it actually "starts" from 1. We can also create a simple extension for this so it can be reused and/or covered byt unit tests:

```Swift
extension RandomAccessCollection {
    var middleItems: (first: Element, second: Element?) {
        let firstIdx = index(startIndex, offsetBy: count / 2)
        guard count.isEven else {
            return (first: self[firstIdx], second: nil)
        }
        return (first: self[index(firstIdx, offsetBy: -1)], second: self[firstIdx])
    }
}

extension FixedWidthInteger {
    var isEven: Bool {
        self % 2 == 0
    }
}
```

output examples:
```Swift
// Odd, we have a single middle item
let odd = [0,1,2,3,4,5,6,7,8]
print(odd.middleItems) // (first: 4, second: nil)
// Even - we have 2 middle items
let even = [0,1,2,3,4,5]
print(even.middleItems) // (first: 2, second: Optional(3))
// Empty
let empty: [Int] = []
print(empty.middleItems) // nil
// Signgle item
let single = [42]
print(single.middleItems) // (first: 42, second: nil)
// Two items
let double = [42, 24]
print(double.middleItems) // (first: 42, second: Optional(24))
```

AirCasting/Utils/AveragingService.swift
smialko: You're right but it's fixable by a one-liner. The error was that I passed `window` instead of `checkWindow` to `perform`.

smialko: Good catch :D I can't believe I missed it 🤔

AirCasting/Utils/AveragingService.swift
smialko: I don't think it's a viable option right now. 
First thing is that when we finish the session we're sending measurements to the server and the server expects and uses the same model as `MeasurementEntity`. I think we should be consistent.
And second, a more trivial argument is that it would require heavy refactoring. We use `MeasurementEntity` in so many places and we would have to make sure we are able to use both `MeasurementEntity` and `AveragedMeasurement`. I don't see the point of doing that before the MVP release.

AirCasting/Utils/AveragingService.swift
smialko: I changed it, now I pass the session to this function and I fetch the session in startPeriodicAveraging.

AirCasting/Utils/AveragingService.swift
smialko: We don't want to have two middle items, we need just one to get all the measurement data apart from value. 
But I can make an extension for that anyways.

AirCasting/Utils/AveragingService.swift
```diff
@@ -122,35 +125,60 @@ final class AveragingService: NSObject, ObservableObject {
...
               let uuid = session.uuid else { return }
-
-        let timer = Timer.publish(every: 5.0, on: .main, in: .common)
+        
+        let timer = Timer.publish(every: TimeInterval(window.rawValue), on: .main, in: .common)
             .autoconnect()
             .sink { [weak self] _ in
+                guard let self = self else { return }
                 context.perform {
                     guard let session = try? context.existingSession(uuid: uuid) else { return }
```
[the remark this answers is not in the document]
smialko: Like I wrote above, I did some refactoring there. 
But to be honest I don't see moving getting a session from here to perform method would make performance better. We'll be doing it every 5s or 60s anyway. Could you elaborate on that?

AirCasting/Utils/AveragingService.swift
smialko: We don't want to have two middle elements, just one. Added extensions.

AirCasting/Utils/AveragingService.swift
smialko: Could you be more specific? Like giving tips on what should I do?

AirCasting/Utils/AveragingService.swift
```diff
@@ -122,35 +125,60 @@ final class AveragingService: NSObject, ObservableObject {
...
-                self?.measurementStreamStorage.accessStorage { storage in
-                    guard let session = try? storage.getExistingSession(with: uuid) else { return }
-                    self?.startPeriodicAveraging(session: session)
-                }
+                self?.startPeriodicAveraging(session: session, window: .firstThresholdWindow)
             }
         timers[uuid] = timer
     }
     
-    private func startPeriodicAveraging(session: SessionEntity) {
```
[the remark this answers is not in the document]
smialko: I did some refactoring and pass uuid to some methods now but I'm still passing sessions to `perform` the method. 
Why: 
First thing, I need to have a session in `startPeriodicAveraging` to be able to pass the data to `averagingWindowFor`. I think that retrieving `session` in both methods would be worse than passing session to `perform`. 

And second, I don't see how retrieving session in `perform` would make performance better? We would be fetching sessions every 5s or 60s anyways. Could you elaborate on that?

AirCasting/Utils/AveragingService.swift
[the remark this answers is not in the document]
smialko: To be fair - it's not only iOS. We had a chitchat with Hania in the kitchen once and she said that dealing with time is always a nightmare.😜

PR #326, opened 2021-10-18
change navigation to proper one

jankrzempek:
https://trello.com/c/posaRwe2/361-home-button-navigation

Always go to FOLLOWING tab

WHEN session is in mobile ACTIVE go to MOBILE ACTIVE

It was done by making a 'trick' -> adding a transparent button as it is not at all easy or don't know how possible to do with current tab selection. I am mentioning this because I considered many options, really 😅

### DIFF
--- a/AirCasting/Dashboard/DashboardView.swift
+++ b/AirCasting/Dashboard/DashboardView.swift
@@ -73,15 +73,6 @@ struct DashboardView: View {
             self.selectedSection.selectedSection = selectedSection
             try! coreDataHook.setup(selectedSection: self.selectedSection.selectedSection)
         }
-        .onAppear {
-            try! coreDataHook.setup(selectedSection: .mobileActive)
-            if sessions.contains(where: { $0.isActive }) {
-                selectedSection.selectedSection = .mobileActive
-            } else {
-                selectedSection.selectedSection = .following
-            }
-            try! coreDataHook.setup(selectedSection: selectedSection.selectedSection)
-        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


AirCasting/Dashboard/DashboardView.swift
```diff
@@ -74,6 +74,12 @@ struct DashboardView: View {
             try! coreDataHook.setup(selectedSection: self.selectedSection.selectedSection)
         }
         .onAppear {
+            try! coreDataHook.setup(selectedSection: .mobileActive)
```
pawelgil, line 77: why do this when you override in line `83` ?

AirCasting/Dashboard/DashboardView.swift
```diff
@@ -74,6 +74,12 @@ struct DashboardView: View {
             try! coreDataHook.setup(selectedSection: self.selectedSection.selectedSection)
         }
         .onAppear {
```
anna1901, line 76: Seems like it's still working when we delete this onAppear, because you've put the same in mainTapBar view

AirCasting/RootAppView.swift
```diff
@@ -17,6 +17,7 @@ struct RootAppView: View {
     @StateObject private var bluetoothManager = BluetoothManager(mobilePeripheralSessionManager: MobilePeripheralSessionManager(measurementStreamStorage: CoreDataMeasurementStreamStorage(persistenceController: PersistenceController.shared)))
     
     @StateObject private var userSettings = UserSettings()
+    @StateObject var selectedSection = SelectSection()
```
anna1901, line 20: Is there a reason for declaring it here instead of MainTabBarView?

jankrzempek: In some point of my mind process it was, but I will check if it is still the case 😄

AirCasting/Dashboard/DashboardView.swift
pawelgil: I still don't get it. Can you try with removing this line and check if it works as expected still?

PR #332, opened 2021-10-20
Fixed bug with stats container in mapview

pawelgil:
Fixed by trial and error. @jankrzempek please confirm that zooming works as expected


jankrzempek (approved):
I've tested it and it does! Sorry for making it in a wrong way :/

pawelgil:
The only way of reproducing the issue was by entering one of "Fixed" sessions. I believe this might be related to active AB sessions which I don't have access to. 
@anna1901 Can you please test following scenarios for me:
1. Do stats container work properly on ended fixed session?
2. Do stats container work properly on ongoing session using AB?

To confirm that this bug is present from the beginning plase test on both this PR branch and on a `532cfa3ff7b938d1ae3945810de605546f375bc5` revision

PR #339, opened 2021-10-21
view now properly scroll to following

jankrzempek:
https://trello.com/c/posaRwe2/361-home-button-navigation

### DIFF
--- a/AirCasting/Dashboard/AirSectionPickerView.swift
+++ b/AirCasting/Dashboard/AirSectionPickerView.swift
@@ -2,7 +2,6 @@
 //
 
 import SwiftUI
-import Combine
 
 struct AirSectionPickerView: View {
     
@@ -22,8 +21,10 @@ struct AirSectionPickerView: View {
                                 }
                                 selection = section
                             }
-                        }.onChange(of: selection, perform: {
-                            $0 == .following ? scrollReader.scrollTo(SelectedSection.following) : nil
+                        }.onChange(of: selection, perform: { section in
+                            withAnimation(.easeInOut(duration: 0.1)) {
+                                section == .following ? scrollReader.scrollTo(SelectedSection.following) : nil
+                            }
                         })
                         .buttonStyle(PickerButtonStyle(isSelected: section == selection))
                         .id(section)


AirCasting/Dashboard/AirSectionPickerView.swift
```diff
@@ -2,6 +2,7 @@
 //
 
 import SwiftUI
+import Combine
```
smialko, line 5: Why did you add Combine? You don't need it :)

AirCasting/Dashboard/AirSectionPickerView.swift
```diff
@@ -14,14 +15,16 @@ struct AirSectionPickerView: View {
                     ForEach(SelectedSection.allCases, id: \.self) { section in
                         Button(section.localizedString) {
                             withAnimation(.easeInOut(duration: 0.1)) {
-                                if section == .mobileDormant {
+                                if section == .mobileDormant || section == .fixed {
```
smialko, line 18: Why? :P We don't need that, do we?

AirCasting/Dashboard/AirSectionPickerView.swift
```diff
@@ -14,14 +15,16 @@ struct AirSectionPickerView: View {
...
                                     scrollReader.scrollTo(SelectedSection.fixed)
                                 } else if section == .mobileActive {
                                     scrollReader.scrollTo(SelectedSection.following)
                                 }
                                 selection = section
                             }
-                        }
+                        }.onChange(of: selection, perform: {
+                            $0 == .following ? scrollReader.scrollTo(SelectedSection.following) : nil
```
smialko, line 26: Could you add animation here? It pops to the left and it's kinda harsh :D

smialko (commented):
Good job ;D

AirCasting/Dashboard/AirSectionPickerView.swift
jankrzempek: Yeah, I've added it when I had the other conception 😅

AirCasting/Dashboard/AirSectionPickerView.swift
jankrzempek: You do, because without it for now: when you are at the leading edge (Following) and try to tap onto fixed but without moving the AirSection at all the 'Fixed' text will be cuted off

smialko: oh, okey, didn't try that edgy edge case :D 
Nice!

PR #340, opened 2021-10-21
Move configuring fixed session on AB to just after connection is made

smialko:
We have to configure AirBeam to a fixed session just after making a connection with it. 

I know, it may not be the most elegant solution, any suggestions are welcomed (but please remember, we'd like to release this app tomorrow. Suggestions with refactor of half of the app aren't welcomed 😜)

### DIFF
--- a/AirCasting/ABConnector/AirBeam3Configurator.swift
+++ b/AirCasting/ABConnector/AirBeam3Configurator.swift
@@ -106,7 +106,6 @@ struct AirBeam3Configurator {
 private extension AirBeam3Configurator {
     
     // MARK: Commands
-    
     private func sendUUIDRequest(uuid: SessionUUID) {
         let message = hexMessageBuilder.uuidMessage(uuid: uuid)
         sendConfigMessage(data: message)
--- a/AirCasting/ABConnector/ConnectingABViewModel.swift
+++ b/AirCasting/ABConnector/ConnectingABViewModel.swift
@@ -37,19 +37,16 @@ class AirbeamConnectionViewModelDefault: AirbeamConnectionViewModel, ObservableO
             self.isDeviceConnectedValue = success
             self.shouldDismissValue = !success
             
-            if success {
-                self.configureABToFixed()
-            }
+            guard success else { return }
+            self.configureABToFixed()
         }
     }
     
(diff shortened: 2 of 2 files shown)


AirCasting/ABConnector/ConnectingABViewModel.swift
```diff
@@ -19,16 +19,43 @@ class AirbeamConnectionViewModelDefault: AirbeamConnectionViewModel, ObservableO
...
+        self.userAuthenticationSession = userAuthenticationSession
+        self.sessionContext = sessionContext
     }
     
     func connectToAirBeam() {
         self.airBeamConnectionController.connectToAirBeam(peripheral: peripheral) { success in
             self.isDeviceConnectedValue = success
             self.shouldDismissValue = !success
+            
+            if success {
```
pawelgil, line 40: guard success else { return }

would be a bit more Swifty

AirCasting/ABConnector/ConnectingABViewModel.swift
```diff
@@ -19,16 +19,43 @@ class AirbeamConnectionViewModelDefault: AirbeamConnectionViewModel, ObservableO
...
+                self.configureABToFixed()
+            }
+        }
+    }
+    
+    private func configureABToFixed() {
+        DispatchQueue.main.asyncAfter(deadline: .now() + .seconds(2)) {
```
pawelgil, line 47: wondering if this really needs to be done on a main queue. I guess for now yes, because this class is using main queue in general, but overall I think we need to switch to a private serial queue at some point. Both here and in BluetoothManager which pretty careless about threads too (I think all of our backbone code needs a solid threading diagnosis)

pawelgil (approved):

AirCasting/ABConnector/AirBeam3Configurator.swift
```diff
@@ -105,6 +101,10 @@ private extension AirBeam3Configurator {
             sendAuthToken(authToken: token)
         }
     }
+}
+
+private extension AirBeam3Configurator {
+    
     // MARK: Commands
```
jankrzempek, line 109: Do we need those two spaces here 😄?

PR #383, opened 2021-10-29
Check if values count is 12

anna1901:
To avoid fatal error when values count != 12


jankrzempek (commented):
Not resolving an issue

jankrzempek (approved):
After our talk

PR #386, opened 2021-11-01
stay on map

jankrzempek:
https://trello.com/c/31DSosIn/397-mobile-active-map-stay-on-map

I know that this code implemented with .scenePhase was there for a reason, but do we really need it still? We have UI suspending feature that works as user locks the phone or put app in the background. I've made other solution to specifically prevent GoogleMap updates instead of going back to homescreen, but do we really need it now?

### DIFF
--- a/AirCasting/Map/GoogleMapView.swift
+++ b/AirCasting/Map/GoogleMapView.swift
@@ -58,7 +58,7 @@ struct GoogleMapView: UIViewRepresentable {
     }
     
     func updateUIView(_ uiView: GMSMapView, context: Context) {
-        if isUserInteracting {
+        guard isUserInteracting else { return }
             let thresholdWitness = ThresholdWitness(sensorThreshold: self.threshold)
             if pathPoints != context.coordinator.currentlyDisplayedPathPoints ||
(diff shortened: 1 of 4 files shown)


AirCasting/Map/GoogleMapView.swift
```diff
@@ -56,22 +58,24 @@ struct GoogleMapView: UIViewRepresentable {
...
-        placePickerDismissed ? uiView.moveCamera(cameraUpdate) : nil
-        // Update camera's starting point
-        guard context.coordinator.shouldAutoTrack else { return }
-            DispatchQueue.main.async {
-                uiView.moveCamera(cameraUpdate)
-                if uiView.camera.zoom > 16 {
-                    // The zoom is set automatically somehow which results sometimes in 'too close' map
-                    // This helps us to fix it and still manage to fit into the 'bigger picture' if needed because of the long session
-                    uiView.animate(toZoom: 16)
+        if isUserInteracting {
```
anna1901, line 61: guard?

anna1901 (approved):
Very smart solution. 🧠  Just add guard instead of if please

PR #391, opened 2021-11-02
Background audio interruptions problems fixed

pawelgil:
[Trello](https://trello.com/c/7M33g3AR/334-microphone-doesnt-resume-recording-when-other-app-was-using-a-microphone-and-then-we-came-back-to-the-app)

* `.duckOthers` option added to our audio session. This means any other AV session will be blocked by us (for example Music app won't be able to play music because of us).
* Moved `AVAudioRecorder` to a single-instance per `MicrophoneManager` instance. This is because it's not possible to create new recorders while in background. We need one and pause/resume it instead.
* Proper session activation/deactivation mechanism added with error handling (TODO)
* `AVSessionInterruptionObserver` implemented in place of the deprecated delegate we've been using.

### DIFF
--- a/AirCasting/MicrophoneSession/MicrophoneManager.swift
+++ b/AirCasting/MicrophoneSession/MicrophoneManager.swift
@@ -6,6 +6,7 @@ import Foundation
 import AVFoundation
 import CoreAudio
 import CoreLocation
+import UIKit
 
 final class MicrophoneManager: NSObject, ObservableObject {
     static private let recordSettings: [String: Any] = [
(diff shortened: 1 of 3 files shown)


jankrzempek (approved):
Tested and working great! 🚀

anna1901 (changes requested):
Interruptions handling works perfectly, but after finishing one microphone session I can't create another one.

AirCasting/MicrophoneSession/MicrophoneManager.swift
```diff
@@ -57,17 +62,17 @@ final class MicrophoneManager: NSObject, ObservableObject {
...
     }
     
     func stopRecording() throws {
+        guard isRecording, let recorder = recorder else { return }
         levelTimer?.invalidate()
         locationProvider.stopUpdatingLocation()
-        isRecording = false
-        recorder.stop()
-        recorder = nil
+        recorder.pause()
```
anna1901, line 75: I think you forgot to put `isRecording = false` here and that causes the issue

AirCasting/MicrophoneSession/MicrophoneManager.swift
```diff
@@ -19,33 +20,37 @@ final class MicrophoneManager: NSObject, ObservableObject {
...
             throw MicrophoneSessionError.permissionNotGranted
         }
         self.session = session
-
-        try audioSession.setCategory(AVAudioSession.Category.playAndRecord)
+        try setupAudioSession()
         try audioSession.setActive(true)
-        try recorder = AVAudioRecorder(url: URL(fileURLWithPath: "/dev/null", isDirectory: true), settings: MicrophoneManager.recordSettings)
         recorder.isMeteringEnabled = true
-        recorder.delegate = self
```
anna1901, line 48: This is not necessary?

AirCasting/MicrophoneSession/MicrophoneManager.swift
```diff
@@ -57,17 +62,17 @@ final class MicrophoneManager: NSObject, ObservableObject {
...
-        recorder.record()
-        sampleMeasurement()
-        levelTimer = Timer.scheduledTimer(timeInterval: 1, target: self, selector: #selector(timerTick), userInfo: nil, repeats: true)
+        if recorder.record() {
+            levelTimer = createTimer()
+            sampleMeasurement()
+        }
     }
     
     func stopRecording() throws {
```
anna1901, line 71: Can this function still throw an error?

AirCasting/MicrophoneSession/MicrophoneManager.swift
pawelgil: not anymore, delegate methods have been deprecated for some time and we use NotificationCenter notifs instead now

PR #394, opened 2021-11-03
Remove print which was doubled in log

anna1901:


jankrzempek (approved):
I really like the way you are cleaning up the code every day

PR #395, opened 2021-11-03
new graph simplify

jankrzempek:
Now it matches mostly android algorithm and modes to structure graph line so it should be closer.

https://trello.com/c/aaVUAALc/357-graph-simplified-line-looks-strange

### DIFF
--- a/AirCasting/Graph/AirCastingGraph.swift
+++ b/AirCasting/Graph/AirCastingGraph.swift
@@ -103,7 +103,7 @@ class AirCastingGraph: UIView {
         // Line styling
         dataSet.drawCirclesEnabled = false
         dataSet.setColor(UIColor(.white))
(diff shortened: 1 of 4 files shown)


AirCasting/Graph/AirCastingGraph.swift
```diff
@@ -103,7 +103,7 @@ class AirCastingGraph: UIView {
         // Line styling
         dataSet.drawCirclesEnabled = false
         dataSet.setColor(UIColor(.white))
-        dataSet.mode = .horizontalBezier
+        dataSet.mode = .cubicBezier
```
anna1901, line 106: Didn't Michael prefer the "horizontalBezier" at some point?

jankrzempek: I am not sure really, we will see then :)

PR #398, opened 2021-11-03
Setup statistics timer only for active and followed sessions

anna1901:


jankrzempek (approved):
Looks good, working as expected - cool!

PR #402, opened 2021-11-04
fix to graph simplifier 

jankrzempek:
https://trello.com/c/aaVUAALc/357-graph-simplified-line-looks-strange


AirCasting/Graph/Graph.swift
```diff
@@ -73,12 +73,16 @@ struct Graph: UIViewRepresentable {
         
         let counter: Int = entries.filter({ $0.x >= startTime && $0.x <= endTime }).count
         
-        let simplifiedPoints = AirCastingGraphSimplifier.simplify(points: entries,
-                                                             visibleElementsNumber: counter,
-                                                             thresholdLimit: simplifiedGraphEntryThreshold)
-        
-        uiView.updateWithEntries(entries: simplifiedPoints, isAutozoomEnabled: isAutozoomEnabled)
-        print(entries.count != simplifiedPoints.count ? "Simplified \(entries.count) to \(simplifiedPoints.count)" : "Not simplyfing")
+        if counter > simplifiedGraphEntryThreshold {
```
pawelgil, line 76: if/else blocks at the end for a function can be nicely simplified using `guard`

```Swift
guard counter > simplifiedGraphEntryThreshold else {
    return
}
let simplifiedPoints = AirCastingGraphSimplifier.simplify(points: entries, visibleElementsNumber: counter, thresholdLimit: simplifiedGraphEntryThreshold)
uiView.updateWithEntries(entries: simplifiedPoints, isAutozoomEnabled: isAutozoomEnabled)
print("Simplified \(entries.count) to \(simplifiedPoints.count)")
```

AirCasting/Graph/AirCastingGraphSimplifier.swift
```diff
@@ -2,16 +2,15 @@
 //
 
 import Foundation
-import SwiftSimplify
 import Charts
 
 final class AirCastingGraphSimplifier {
     
-    static func simplify<P: Point2DRepresentable>(points: [P], visibleElementsNumber: Int, thresholdLimit: Int) -> [P] {
+    static func simplify(points: [ChartDataEntry], visibleElementsNumber: Int, thresholdLimit: Int) -> [ChartDataEntry] {
```
pawelgil, line 9: is this de-generalization really needed? It doesn't look like it is changing anything except, well, de-generalizing ;)

jankrzempek: Yeah, you right. There is one thing - the  `Point2DRepresentable` was taken from SwiftSimplifier library which we have added. I thought that if we are going to use our own implementation now, we can get rid of the library in the future. That was the point of changing it, what do you think?

anna1901 (approved):
Line looks good now 🚀

AirCasting/Graph/AirCastingGraphSimplifier.swift
pawelgil: We could create our own protocol for that, but in this case I think it wouldn't matter. Cool then! 😀

pawelgil (approved):

PR #404, opened 2021-11-04
Reconnect only for mobile active session peripheral

anna1901:
This fixes the leftover bug and prevents the app from reconnecting with the AirBeam when it was set to a fixed session and not active mobile session.


jankrzempek (approved):
Awesomeness 🚀

PR #411, opened 2021-11-08
center map over current location

jankrzempek:
https://trello.com/c/QJGnphuq/398-mobile-active-map-when-first-entering-the-map-center-map-over-current-location


anna1901 (approved):
Nice! 🚀

PR #416, opened 2021-11-10
Wait for discovered characteristics when reconnecting with ab

anna1901:
* When reconnecting with device previously we were waiting 4 seconds before sending configuration. Now we are waiting for the discovered characteristic notification
* In the notification I added user info with a peripheral uuid just to double check that discovered characteristic notification is related to the right peripheral

### DIFF
--- a/AirCasting/ABConnector/BluetoothManager.swift
+++ b/AirCasting/ABConnector/BluetoothManager.swift
@@ -110,15 +110,19 @@ extension BluetoothManager: CBCentralManagerDelegate {
     func centralManager(_ central: CBCentralManager, didConnect peripheral: CBPeripheral) {
         // Here's code for getting data from AB.
         peripheral.delegate = self
-        peripheral.discoverServices(nil)
         if mobilePeripheralSessionManager.activeSessionInProgressWith(peripheral) {
-            DispatchQueue.main.asyncAfter(deadline: .now() + .seconds(4)) {
+            var characteristicsHandle: Any?
+            characteristicsHandle = NotificationCenter.default.addObserver(forName: .discoveredCharacteristic, object: nil, queue: .main) { notification in
+                guard notification.userInfo?["peripheral uuid"] as! UUID == peripheral.identifier else { return }
                 self.mobileSessionReconnected = true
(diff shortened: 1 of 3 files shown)


AirCasting/ABConnector/BluetoothManager.swift
```diff
@@ -110,15 +110,19 @@ extension BluetoothManager: CBCentralManagerDelegate {
     func centralManager(_ central: CBCentralManager, didConnect peripheral: CBPeripheral) {
         // Here's code for getting data from AB.
         peripheral.delegate = self
-        peripheral.discoverServices(nil)
         if mobilePeripheralSessionManager.activeSessionInProgressWith(peripheral) {
-            DispatchQueue.main.asyncAfter(deadline: .now() + .seconds(4)) {
+            var characteristicsHandle: Any?
+            characteristicsHandle = NotificationCenter.default.addObserver(forName: .discoveredCharacteristic, object: nil, queue: .main) { notification in
+                guard notification.userInfo?["peripheral uuid"] as! UUID == peripheral.identifier else { return }
```
pawelgil, line 116: we have a special place to store keys instead of typing a string (the same file as `.discoveredCharacteristic`, the `AirCastingNotificationKeys` enum). This prevents us from making a typo and wasting 3 days to find why stuff is not working ;)

AirCasting/ABConnector/BluetoothManager.swift
```diff
@@ -156,7 +160,7 @@ extension BluetoothManager: CBPeripheralDelegate {
                 }
             }
         }
-        hasSomeCharacteristics ? NotificationCenter.default.post(name: .discoveredCharacteristic, object: nil, userInfo: nil) : nil
+        hasSomeCharacteristics ? NotificationCenter.default.post(name: .discoveredCharacteristic, object: nil, userInfo: ["peripheral uuid" : peripheral.identifier]) : nil
```
pawelgil, line 163: and here

pawelgil (changes requested):
good find! Let's just move the key from string to const

jankrzempek (commented):
Nice one!

pawelgil (approved):

PR #417, opened 2021-11-10
Wizard wifi - improvement

jankrzempek:
https://trello.com/c/vcecEAYV/245-fixed-wizard-improve-wifi-password-input-ui


AirCasting/CreateSessionViews/CreateSessionDetails/CreateSessionDetailsView.swift
```diff
@@ -0,0 +1,141 @@
...
+import AirCastingStyling
+import SwiftUI
+
+struct CreateSessionDetailsView: View {
+    
+    @EnvironmentObject private var sessionContext: CreateSessionContext
+    @StateObject var viewModel: CreateSessionDetailsViewModel
+    @Binding var creatingSessionFlowContinues: Bool
+    
+    var body: some View {
```
pawelgil, line 16: nice one, but please split this into smaller, well named chunks

AirCasting/CreateSessionViews/CreateSessionDetails/CreateSessionDetailsView.swift
```diff
@@ -0,0 +1,141 @@
...
+        }
+        .onAppear { viewModel.onScreenEnter() }
+    }
+}
+
+private extension CreateSessionDetailsView {
+    
+    var navigation: some View {
+        ZStack {
+            NavigationLink(
```
pawelgil, line 60: let's move both of those NavigationLinks to separate `private vars` with nice names (locationPickerLink and createSesssionLink or smth)

AirCasting/CreateSessionViews/CreateSessionDetails/CreateSessionDetailsViewModel.swift
```diff
@@ -0,0 +1,81 @@
...
+    @Published var isIndoor = true
+    @Published var isWiFi = true
+    @Published var wifiPassword: String = ""
+    @Published var wifiSSID: String = ""
+    @Published var isConfirmCreatingSessionActive: Bool = false
+    @Published var isLocationSessionDetailsActive: Bool = false
+    @Published var showingAlert = false
+    @Published var isSSIDTextfieldDisplayed: Bool = false
+    
+    let baseURL: BaseURLProvider
```
pawelgil, line 19: private

AirCasting/CreateSessionViews/CreateSessionDetails/CreateSessionDetailsViewModel.swift
```diff
@@ -0,0 +1,81 @@
...
+    
+    func connectToOtherNetworkClick() {
+        isSSIDTextfieldDisplayed = true
+    }
+    
+    func showCompleteCredentials() -> Bool {
+        isWiFi && isSSIDTextfieldDisplayed
+    }
+    
+    func getWiFiSsid() -> String? {
```
pawelgil, line 69: can this be private?

AirCasting/CreateSessionViews/CreateSessionDetails/CreateSessionDetailsViewModel.swift
```diff
@@ -0,0 +1,81 @@
...
+    init(baseURL: BaseURLProvider) {
+        self.baseURL = baseURL
+    }
+    
+    func onScreenEnter() {
+        if let ssid = getWiFiSsid() { wifiSSID = ssid }
+        isSSIDTextfieldDisplayed = wifiSSID.isEmpty
+    }
+    
+    func onContinueClick(sessionContext: CreateSessionContext) -> CreateSessionContext {
```
pawelgil, line 30: this is a bit weird that this is modifying a context instead of handing it end-to-end. But I think this is how this flow works, so it's required. Maybe a comment about that would be nice so if someone doesn't know about that fact they won't be confused

AirCasting/CreateSessionViews/CreateSessionDetails/CreateSessionDetailsViewModel.swift
```diff
@@ -0,0 +1,81 @@
...
+        sessionContext.sessionName = sessionName
+        sessionContext.sessionTags = sessionTags
+        
+        if sessionContext.sessionType == SessionType.fixed {
```
pawelgil, line 34: let's make this a guard with early return instead:

```Swift
guard sessionContext.sessionType != .fixed {
    sessionContext.isIndoor = false
    isConfirmCreatingSessionActive = true
    return sessionContext
}
...
```

AirCasting/CreateSessionViews/CreateSessionDetails/CreateSessionDetailsViewModel.swift
pawelgil, line 30: after making changes from other comments in this function please try to split it up into 2 or 3 smaller ones if possible

AirCasting/CreateSessionViews/CreateSessionDetails/CreateSessionDetailsViewModel.swift
```diff
@@ -0,0 +1,81 @@
...
+    
+    func areCredentialsEmpty() -> Bool {
+        wifiSSID.isEmpty && wifiPassword.isEmpty
+    }
+    
+    func connectToOtherNetworkClick() {
+        isSSIDTextfieldDisplayed = true
+    }
+    
+    func showCompleteCredentials() -> Bool {
```
pawelgil, line 65: name is not good, it should be in asking form as it's a value-returning function. Something  like `shouldShowCompleteCredentials`

AirCasting/CreateSessionViews/CreateSessionDetails/CreateSessionDetailsViewModel.swift
```diff
@@ -0,0 +1,81 @@
...
+    
+    @Published var sessionName: String = ""
+    @Published var sessionTags: String = ""
+    @Published var isIndoor = true
+    @Published var isWiFi = true
+    @Published var wifiPassword: String = ""
+    @Published var wifiSSID: String = ""
+    @Published var isConfirmCreatingSessionActive: Bool = false
+    @Published var isLocationSessionDetailsActive: Bool = false
+    @Published var showingAlert = false
```
pawelgil, line 16: which one? Please enhance the name to be more descriptive

AirCasting/CreateSessionViews/CreateSessionDetails/CreateSessionDetailsViewModel.swift
```diff
@@ -0,0 +1,81 @@
...
+        isSSIDTextfieldDisplayed = true
+    }
+    
+    func showCompleteCredentials() -> Bool {
+        isWiFi && isSSIDTextfieldDisplayed
+    }
+    
+    func getWiFiSsid() -> String? {
+        var ssid: String?
+        if let interfaces = CNCopySupportedInterfaces() as NSArray? {
```
pawelgil, line 71: `guard` here plz

AirCasting/CreateSessionViews/CreateSessionDetails/CreateSessionDetailsView.swift
```diff
@@ -0,0 +1,141 @@
+//
+//  AddNameAndTagsView.swift
```
pawelgil, line 2: Name doesn't match the actual filename
(btw. this is the problem with those useless header comments in files, those need to be maintained and give zero value)

pawelgil (changes requested):
really good job! Some cosmetic changes are needed

AirCasting/CreateSessionViews/CreateSessionDetails/CreateSessionDetailsViewModel.swift
jankrzempek: Cannot be because I use it in the view

AirCasting/CreateSessionViews/CreateSessionDetails/CreateSessionDetailsViewModel.swift
jankrzempek: Yes!

AirCasting/CreateSessionViews/CreateSessionDetails/CreateSessionDetailsView.swift
```diff
@@ -0,0 +1,161 @@
+//  Created by Anna Olak on 24/02/2021.
+//
+import AirCastingStyling
+import SwiftUI
+
+struct CreateSessionDetailsView: View {
+    
+    @EnvironmentObject private var sessionContext: CreateSessionContext
+    @StateObject var viewModel: CreateSessionDetailsViewModel
```
smialko, line 9: As always, I don't like this. There was a discussion about that couple of times 😛 Apple doesn't recommend that: (see Apple's dev comment here --> https://developer.apple.com/forums/thread/650618)

Ania found a nice way to be sure that the viewModel won't be created a couple of times that with `init(wrappedValue: )`. 
Tbh, I'm not 100% sure if `init(wrappedValue: )` isn't a default init used here, you can test it. If yes, it should be ok, if not it's dangerous.

AirCasting/CreateSessionViews/CreateSessionDetails/CreateSessionDetailsViewModel.swift
pawelgil: Oh, now I see why. It's required by some further, nested views and because of that it needs to be exposed here. Sad. A comment would be nice explaining that

AirCasting/CreateSessionViews/CreateSessionDetails/CreateSessionDetailsView.swift
```diff
@@ -0,0 +1,161 @@
...
+        GeometryReader { geometry in
+            ScrollView(.vertical) {
+                    VStack(alignment: .leading, spacing: 25) {
+                        ProgressView(value: 0.75)
+                        titleLabel
+                        VStack(spacing: 20) {
+                            sessionNameField
+                            sessionTagsField
+                        }
+                        if sessionContext.sessionType == SessionType.fixed {
```
pawelgil, line 22: this whole body of `if` can be extracted to a private var called for example 

```Swift
private var fixedSessionDetails: some View
`

File: AirCasting/CreateSessionViews/CreateSessionDetails/CreateSessionDetailsView.swift [hunk 13]
```diff
@@ -0,0 +1,161 @@
...
+                            placementPicker
+                            transmissionTypePicker
+                            if viewModel.shouldShowCompleteCredentials() {
```
pawelgil, line 25: after extracting the containing `if` body, do the same extraction here with the `if` and `else if`. 

And observe how the code becomes easily top-down readable without the need to focus on low-level internals ;)

pawelgil (changes requested):

AirCasting/CreateSessionViews/CreateSessionDetails/CreateSessionDetailsView.swift
jankrzempek: I didn't know about that. As long as it is the view that is created every time we are going to create a new session I think it won't be a problem in that case. As I am seeing it right now is that it is created once and doesn't recreate when I move out of that view. Maybe I've not tested the way I should. Please tell me how to be sure and what do you think I should do now?

pawelgil (approved):

AirCasting/CreateSessionViews/CreateSessionDetails/CreateSessionDetailsView.swift
smialko: if you tested it and it's not created more than once while being on this view, it should be fine. :)
Anyway, I think it's good to know that it's not recommended :D

jankrzempek: For sure it is! Thanks for letting me know, I will try to take care in the future from now

anna1901: As I understand, view model (or any other source of truth for data) should be marked as @StateObject in the place where it's initialized, and then as @ObservedObject in the view to which it is passed (and which should change with a change of some published property). So for example, when we were working on the chart with @pawelgil, we've initialized ChartViewModel in SessionCardView and marked in as @StateObject and then we've passed it to ChartView, where it was an @ObservedObject. This way the whole chart view wasn't reinitialized with each change of ChartViewModel.

jankrzempek: Thank You for the whole, great explanation. I will do my best to make it great and as you said and merge

PR #420, opened 2021-11-12
Pull to refresh added

pawelgil:
## What was done?
* `RefreshControl` - a custom, pure SwiftUI component
* Pull to refresh implemented in dashboard for lists and empty views

Trello: https://trello.com/c/EwBBax6c

### DIFF
--- a/AirCasting/Dashboard/DashboardView.swift
+++ b/AirCasting/Dashboard/DashboardView.swift
@@ -7,6 +7,8 @@
 
 import CoreData
 import SwiftUI
+import AirCastingStyling
+import Combine
 
 struct DashboardView: View {
     #warning("This hook fires too often - on any stream measurement added/changed. Should only fire when list changes.")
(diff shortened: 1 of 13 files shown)


AirCasting/Dashboard/DashboardView.swift
```diff
@@ -7,26 +7,37 @@
...
     @EnvironmentObject var averaging: AveragingService
+    @State var isRefreshing: Bool = false
     
-    let measurementStreamStorage: MeasurementStreamStorage
-    let sessionStoppableFactory: SessionStoppableFactory
+    private let measurementStreamStorage: MeasurementStreamStorage
+    private let sessionStoppableFactory: SessionStoppableFactory
+    private let sessionSynchronizer: SessionSynchronizer
+    
+    private let dashboardCoordinateSpaceName = "dashboardCoordinateSpace"
```
jankrzempek, line 25: Do we want to move it into Strings?

AirCasting/Dashboard/RefreshControl.swift
```diff
@@ -0,0 +1,61 @@
...
+            }
+            ZStack(alignment: .center) {
+                if isRefreshing { ProgressView() } else { mimicPullToRefreshAnimation(with: geometry) }
+            }
+            .frame(width: geometry.size.width)
+        }
+        // paddings: for non-refreshing state we need negative top padding to hide
+        // the refresh control correctly under the top part of a parent view.
+        //
+        // Values here are handpicked using the eye-ball technique 👀
```
jankrzempek, line 25: 😁👌😁

jankrzempek (approved):
Working like a charm, I've refreshed my app many times now 👌

AirCasting/Dashboard/DashboardView.swift
pawelgil: We only put user-facing strings there (so we have them all in a single place), this one is not user facing, just a key

jankrzempek: Got it - thanks, so no doubts then 😄

AirCasting/Dashboard/DashboardView.swift
```diff
@@ -37,58 +48,97 @@ struct DashboardView: View {
...
+            if sessions.isEmpty { emptySessionsView } else { sessionListView }
         }
         .navigationBarTitle(NSLocalizedString(Strings.DashboardView.dashboardText, comment: ""))
         .onChange(of: selectedSection.selectedSection) { selectedSection in
             self.selectedSection.selectedSection = selectedSection
             try! coreDataHook.setup(selectedSection: self.selectedSection.selectedSection)
         }
+        .onChange(of: isRefreshing, perform: { newValue in
+            guard newValue == true else { return }
+            guard !sessionSynchronizer.syncInProgress.value else {
```
anna1901, line 61: I think, that maybe when we try to run syncing and there is already sync in progress (like here or on log out), we should check if there were any changes introduced already after the sync started so that we could schedule another sync with them. For example, there is a sync already in progress, but someone decides they want to delete one more session so they delete it and then pull to refresh. Then this deleted session would not be deleted from backed until another sync happens, isn't that right? Do you think that's something you could add in this PR? Maybe first version we could trigger another sync in closure below, but it's probably not an optimal solution.

anna1901 (approved):
Looks awesome!! 👏 🤩 🚀

AirCasting/Dashboard/DashboardView.swift
pawelgil: That is an excellent point and I think should be considered more widely since we're also triggering synchronization when user stops a session. So there is additional scenario:
* Sync is running because of coming back from background (for example)
* User stops a sesssion
* It won't get synchronized until next sync
And since we removed the 5-min timer synchronization it won't until user pulls to refresh. This is a problem and it's great you found it so quickly. I will resolve it on a higher level by adding `.required` option or smth like that to the synchronizer.

btw. @anna1901 @anna1901 @mturnau we're not triggering synchronization when deleting sessions, shouldn't we add it too?

PR #421, opened 2021-11-12
Fix for CoreData logout related crashes

pawelgil:


jankrzempek (approved):
I've tested it and seems to be working great. I've not encountered that bug many times before but still, it is working 🚀

anna1901 (approved):
Looks good and works well 🚀  🙂

PR #426, opened 2021-11-16
Reset thresholds to default

smialko:
I renamed HeatmapSettingsView because I think this name would be confusing after implementing heatmap.

### DIFF
--- a/AirCasting/CoreData/MeasurementStreamEntity.swift
+++ b/AirCasting/CoreData/MeasurementStreamEntity.swift
@@ -50,7 +50,7 @@ public class MeasurementStreamEntity: NSManagedObject, Identifiable {
         MeasurementStreamLocalID(id: objectID)
     }
     public var thresholds: [Int32] {
-     [thresholdVeryLow, thresholdLow , thresholdMedium ,thresholdHigh, thresholdVeryHigh ]
+     [thresholdVeryLow, thresholdLow, thresholdMedium, thresholdHigh, thresholdVeryHigh]
     }
 }
 
--- a/AirCasting/CoreData/MeasurementStreamEntity.swift
+++ b/AirCasting/CoreData/MeasurementStreamEntity.swift
@@ -49,6 +49,9 @@ public class MeasurementStreamEntity: NSManagedObject, Identifiable {
     public var localID: MeasurementStreamLocalID {
         MeasurementStreamLocalID(id: objectID)
     }
+    public var thresholds: [Int32] {
+     [thresholdVeryLow, thresholdLow , thresholdMedium ,thresholdHigh, thresholdVeryHigh ]
+    }
 }
 
 // MARK: Generated accessors for measurements
--- a/AirCasting/Graph/GraphView.swift
+++ b/AirCasting/Graph/GraphView.swift
@@ -42,33 +42,35 @@ struct GraphView<StatsViewModelType>: View where StatsViewModelType: StatisticsC
                 .padding(.horizontal)
            
             if isProceeding(session: session) {
-                    if let threshold = thresholds.threshold(for: selectedStream) {
+                if let threshold = thresholds.threshold(for: selectedStream) {
+                    if let selectedStream = selectedStream {
                         ZStack(alignment: .topLeading) {
-                            if let selectedStream = selectedStream {
-                                Graph(stream: selectedStream,
-                                      thresholds: threshold,
-                                      isAutozoomEnabled: session.type == .mobile).onDateRangeChange { [weak graphStatsDataSource, weak statsContainerViewModel] range in
-                                        graphStatsDataSource?.dateRange = range
-                                        statsContainerViewModel?.adjustForNewData()
-                                      }
-                                // Statistics container shouldn't be presented in mobile dormant tab
-                                if !(session.type == .mobile && session.isActive == false) {
-                                    StatisticsContainerView(statsContainerViewModel: statsContainerViewModel,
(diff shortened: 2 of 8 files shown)


AirCasting/Graph/GraphView.swift
```diff
@@ -42,33 +42,35 @@ struct GraphView<StatsViewModelType>: View where StatsViewModelType: StatisticsC
...
                                                         threshold: threshold)
-                                }
                             }
                         }
-                        NavigationLink(destination: HeatmapSettingsView(changedThresholdValues: threshold.rawThresholdsBinding)) {
+                        NavigationLink(destination: ThresholdsSettingsView(thresholdValues: threshold.rawThresholdsBinding,
+                                                                           initialThresholds: selectedStream.thresholds)) {
                             EditButtonView()
+                                .padding(.horizontal)
+                                .padding(.top)
```
jankrzempek, line 64: can we use `.paddding([.horizontal, .top])` please?

AirCasting/SessionViews/ThresholdSettings/ThresholdsSettingsView.swift
```diff
@@ -50,37 +53,22 @@ struct HeatmapSettingsView: View {
...
-            thresholdVeryLow = "\(Int(changedThresholdValues[0]))"
-            thresholdLow = "\(Int(changedThresholdValues[1]))"
-            thresholdMedium = "\(Int(changedThresholdValues[2]))"
-            thresholdHigh = "\(Int(changedThresholdValues[3]))"
-            thresholdVeryHigh = "\(Int(changedThresholdValues[4]))"
+            thresholdSettingsViewModel.thresholdVeryLow = "\(Int(thresholdValues[0]))"
+            thresholdSettingsViewModel.thresholdLow = "\(Int(thresholdValues[1]))"
+            thresholdSettingsViewModel.thresholdMedium = "\(Int(thresholdValues[2]))"
+            thresholdSettingsViewModel.thresholdHigh = "\(Int(thresholdValues[3]))"
+            thresholdSettingsViewModel.thresholdVeryHigh = "\(Int(thresholdValues[4]))"
```
jankrzempek, line 68: What would you say about sth like:

` .onAppear {
            thresholdSettingsViewModel.thresholdVeryLow = string(using: thresholdValues[0])
            thresholdSettingsViewModel.thresholdLow = string(using: thresholdValues[1])
            thresholdSettingsViewModel.thresholdMedium = string(using: thresholdValues[2])
            thresholdSettingsViewModel.thresholdHigh = string(using: thresholdValues[3])
            thresholdSettingsViewModel.thresholdVeryHigh = string(using: thresholdValues[4])
        }
    }
    
    func string(using thresholdValue: Float) -> String {
        return String(Int(thresholdValue))
    }`
    
    Just to consider: if not, then go below 😇

AirCasting/SessionViews/ThresholdSettings/ThresholdsSettingsViewModel.swift
```diff
@@ -0,0 +1,42 @@
...
+    }
+    
+    func resetToDefault() -> [Float] {
+        var newInitial: [Float] = []
+        for value in initialThresholds {
+            let newValue = Float(value)
+            newInitial.append(newValue)
+        }
+        return newInitial
+    }
```
jankrzempek, line 31: Suggestion time:

`func resetToDefault() -> [Float] {
        var newInitial: [Float] = []
        initialThresholds.forEach { value in
            newInitial.append(Float(value))
        }
        return newInitial
    }`

jankrzempek (changes requested):
Working like a charm! I left some notes, but nothing you cannot skip I think.
Only those spaces in the first comment, please 😇

AirCasting/SessionViews/ThresholdSettings/ThresholdsSettingsView.swift
smialko: Sure! tbh I didn't give it much thought - just changed this code to use the view model.

AirCasting/CoreData/MeasurementStreamEntity.swift
```diff
@@ -49,6 +49,9 @@ public class MeasurementStreamEntity: NSManagedObject, Identifiable {
     public var localID: MeasurementStreamLocalID {
         MeasurementStreamLocalID(id: objectID)
     }
+    public var thresholds: [Int32] {
+     [thresholdVeryLow, thresholdLow , thresholdMedium ,thresholdHigh, thresholdVeryHigh ]
+    }
```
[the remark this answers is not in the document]
smialko: you're right, this looks AWFUL :D

AirCasting/CoreData/MeasurementStreamEntity.swift
```diff
@@ -49,6 +49,9 @@ public class MeasurementStreamEntity: NSManagedObject, Identifiable {
     public var localID: MeasurementStreamLocalID {
         MeasurementStreamLocalID(id: objectID)
     }
+    public var thresholds: [Int32] {
+     [thresholdVeryLow, thresholdLow, thresholdMedium, thresholdHigh, thresholdVeryHigh]
```
anna1901, line 53: Perhaps this could be the thresholds struct and not an array? Using array in this case seems prone to error and we already know there can be only this limited set of 5 thresholds, so using a struct (or enum or dictionary) seems like a good fit. If that changes in the future, for example we will have 4 thresholds, it will be easier to edit without errors caused by `thresholds[4]`.

PR #429, opened 2021-11-17
Fit screen iphone se

jankrzempek:
https://trello.com/c/cu6jE4eo/382-scale-screen-so-it-fits-iphonese

To test, please use Ipod touch 7 simulator


AirCasting/SessionViews/ABMeasurementsView.swift
```diff
@@ -62,20 +62,17 @@ struct _ABMeasurementsView: View {
...
+                                                      threshold: threshold,
+                                                      selectedStream: $selectedStream,
+                                                      isCollapsed: $isCollapsed,
+                                                      measurementPresentationStyle: measurementPresentationStyle,
+                                                      isDormant: session.isDormant)
                             }
+                            Spacer()
                         }
-                        .padding(.horizontal, 8)
-                        .frame(maxWidth: .infinity, alignment: .leading)
```
smialko, line 78: Should AB measurements be on the left side of the session card? I think not but tbh I'm not sure. I think that removing the frame and padding moved them to the left. Also, they seem more condensed than previously. 
IMG_F07B0AE0675D-1

smialko: Uh, now I see the `Spacer()` - it's its fault the measurements are on the left side.

smialko (changes requested):
Great job! Just please fix the ABMeasurementsView :)

AirCasting/SessionViews/ABMeasurementsView.swift
smialko: For comparison I'm attaching a photo of how it looked before: 

image

pawelgil (approved):
Note: I only review the code assuming it's working as expected

AirCasting/SessionViews/ABMeasurementsView.swift
```diff
@@ -111,16 +109,17 @@ struct _ABMeasurementsView: View {
...
             Group {
+                streamsToShow.count != 1 ? Spacer() : nil
                 ForEach(streamsToShow, id : \.self) { stream in
                     SingleMeasurementView(stream: stream,
                                           threshold: nil,
                                           selectedStream: $selectedStream,
                                           isCollapsed: $isCollapsed,
                                           measurementPresentationStyle: .hideValues,
                                           isDormant: session.isDormant)
+                    Spacer()
```
smialko, line 120: I don't think using a Spacer() here is a good idea, right now the AB parameters are squeezed in the middle and there's too much space on the sides. Maybe try to use horizontal `.padding()` instead? :)

smialko: Here is just an example, don't hesitate to do it another way:

```
private var streamNames: some View {
        return HStack {
            Group {
                ForEach(streamsToShow, id : \.self) { stream in
                                          threshold: nil,
                                          selectedStream: $selectedStream,
                                          isCollapsed: $isCollapsed,
                                          measurementPresentationStyle: .hideValues,
                                          isDormant: session.isDormant)
                    streamsToShow.count == 1 ? Spacer() : nil
                }
            }
            .padding(.horizontal)
        }
    }
```

smialko (commented):
it's better now! 😄

PR #433, opened 2021-11-19
Create threshold struct

smialko:
Like Ania suggested - it's safer to have a Threshold struct than to use Array. (https://github.com/HabitatMap/AirCastingiOS/pull/426#discussion_r752078340)

### DIFF
--- a/AirCasting/Models/ThresholdsValue.swift
+++ b/AirCasting/Models/ThresholdsValue.swift
@@ -1,0 +1,33 @@
+// Created by Lunar on 19/11/2021.
+//
+
+import Foundation
+
+public struct ThresholdsValue {
+    let veryLow: Int32
+    let low: Int32
+    let medium: Int32
+    let high: Int32
+    let veryHigh: Int32
+    
+    var toArray: [Int32] {
+        [veryLow, low, medium, high, veryHigh]
+    }
+    
+    init(array: [Int32]) {
+        assert(array.count == 5)
+        veryLow = array[0]
+        low = array[1]
+        medium = array[2]
+        high = array[3]
+        veryHigh = array[4]
+    }
+    
+    public init(veryLow: Int32, low: Int32, medium: Int32, high: Int32, veryHigh: Int32) {
+        self.veryLow = veryLow
+        self.low = low
+        self.medium = medium
+        self.high = high
+        self.veryHigh = veryHigh
+    }
+}
--- a/AirCasting/SessionViews/ThresholdSettings/ThresholdsSettingsView.swift
+++ b/AirCasting/SessionViews/ThresholdSettings/ThresholdsSettingsView.swift
@@ -10,12 +10,12 @@ import AirCastingStyling
 
 struct ThresholdsSettingsView: View {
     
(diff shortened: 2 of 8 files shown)


AirCasting/SessionViews/ThresholdSettings/ThresholdsSettingsViewModel.swift
```diff
@@ -10,32 +10,29 @@ class ThresholdSettingsViewModel: ObservableObject {
...
-        var newThresholdValues: [Float] = []
+        var newThresholdValues: [Int32] = []
         for value in stringThresholdValues {
-            let convertedValue = convertToFloat(value: value)
+            let convertedValue = convertToInt(value)
             newThresholdValues.append(convertedValue)
         }
-        return newThresholdValues.sorted { $0 < $1 }
+        let sortedValue = newThresholdValues.sorted { $0 < $1 }
+        return ThresholdsValue(array: sortedValue)
```
anna1901, line 31: We are using an array here to sort the values, yes? Since we are creating the array here, and we are sure it has only 5 values, probably it is better to split it to separate values here and use the other initializer, and remove this initializer from ThresholdValue struct.

AirCasting/SessionViews/ThresholdSettings/ThresholdsSettingsView.swift
```diff
@@ -61,11 +61,11 @@ struct ThresholdsSettingsView: View {
             .buttonStyle(BorderlessButtonStyle())
         }
         .onAppear {
-            thresholdSettingsViewModel.thresholdVeryLow = string(thresholdValues[0])
-            thresholdSettingsViewModel.thresholdLow = string(thresholdValues[1])
-            thresholdSettingsViewModel.thresholdMedium = string(thresholdValues[2])
-            thresholdSettingsViewModel.thresholdHigh = string(thresholdValues[3])
-            thresholdSettingsViewModel.thresholdVeryHigh = string(thresholdValues[4])
+            thresholdSettingsViewModel.thresholdVeryLow = string(thresholdValues.veryLow)
```
anna1901, line 64: This looks much nicer, awesome 🙂

AirCasting/Models/ThresholdsValue.swift
```diff
@@ -0,0 +1,33 @@
...
+    let medium: Int32
+    let high: Int32
+    let veryHigh: Int32
+    
+    var toArray: [Int32] {
+        [veryLow, low, medium, high, veryHigh]
+    }
+    
+    init(array: [Int32]) {
+        assert(array.count == 5)
```
anna1901, line 18: Here we have to make an assumption that we will get an array with 5 elements. Probably it would be better to remove this initializer and split the array into appropriate values in the view model (as mentioned in another comment)

anna1901 (commented):
Nice! 🚀

jankrzempek (approved):
Nothing more than Ania mentioned. Cool work 🚀

pawelgil (commented):
@anna1901 comments are valid, added some thought from myself

PR #438, opened 2021-11-22
Sd card sync data download

anna1901:
* connecting to AB
* sending sd data download configuration mode
* adding mechanism for subscribing to characteristics to bluetooth manager
* saving data received form AB to mobile.csv and fixed.csv files


AirCasting/ABConnector/BluetoothManager.swift
```diff
@@ -68,6 +75,49 @@ class BluetoothManager: NSObject, ObservableObject {
...
+        characteristicsMappingLock.unlock()
+        return observer.identifier
+    }
+
+    /// Removes an entry from observing characteristic
+    /// - Parameter token: Opaque token received on subscription
+    /// - Returns: A `Bool` value indicating if a given token was successfuly removed. Only reason it can fail is double unregistration.
+    @discardableResult func unsubscribeCharacteristicObserver(_ token: AnyHashable) -> Bool {
+        guard let uuid = token as? UUID else { return false }
+//        characteristicsMappingLock.lock()
```
pawelgil, line 113: I think we should switch from `NSLock` to `NSRecursiveLock` and should be fine. Can you please test it out? Also there is a missed return without unlocking right in the below, so that's not good. Let's do:
`characteristicsMappingLock.lock(); defer { characteristicsMappingLock.unlock() }` (and remove the commented unlock from L118)

AirCasting/ABConnector/BluetoothManager.swift
```diff
@@ -158,32 +211,55 @@ extension BluetoothManager: CBPeripheralDelegate {
...
         hasSomeCharacteristics ? NotificationCenter.default.post(name: .discoveredCharacteristic, object: nil, userInfo: [AirCastingNotificationKeys.DiscoveredCharacteristic.peripheralUUID : peripheral.identifier]) : nil
     }
-    
+
     func peripheral(_ peripheral: CBPeripheral, didUpdateValueFor characteristic: CBCharacteristic, error: Error?) {
         guard let value = characteristic.value else {
             Log.warning("AirBeam sent measurement without value")
             return
         }
+//        characteristicsMappingLock.lock()
```
pawelgil, line 234: I think after making the lock recursive it should work just fine. But I think performance here might be hurt because we're unlocking only after triggering all observers. I think we should only obtain them in critical section and after that's done (by copying) we can trigger them outside. But maybe let's dig into this if we actually see some performance related issues

AirCasting/ABConnector/SDCardAirBeamServices.swift
```diff
@@ -0,0 +1,104 @@
+// Created by Lunar on 17/11/2021.
+//
+
+import CoreBluetooth
+import CoreMIDI
+
+enum SDCardSessionType: CaseIterable {
```
pawelgil, line 7: let's move this out to a separate file or to a controller file (it's higher level, universal concept, doesn't really belong only to the Services part imo)

AirCasting/ABConnector/SDCardAirBeamServices.swift
```diff
@@ -0,0 +1,104 @@
...
+            case .success(let data):
+                guard let data = data, let payload = String(data: data, encoding: .utf8) else {
+                    return
+                }
+                self.currentSessionType = self.currentSessionType.next
+                
+                Log.info(payload)
+                // Payload format is ` Some string: ${number_of_entries_expected} `
+                if payload == "SD_SYNC_FINISH" {
+                    completion(.success(()))
```
pawelgil, line 53: those 3 lines are repeated in here and in failure and are responsible for exactly the same logical thing, so a good idea would be to put them in a separate function and only call it here. For example
```Swift
finishSynchronization { completion(.success()) } // and .failure(error) for the failed case
```

AirCasting/ABConnector/SDCardAirBeamServices.swift
```diff
@@ -0,0 +1,104 @@
...
+        metadataCharacteristicObserver = bluetoothManager.subscribeToCharacteristic(DOWNLOAD_META_DATA_FROM_SD_CARD_CHARACTERISTIC_UUID) { result in
+            switch result {
+            case .success(let data):
+                guard let data = data, let payload = String(data: data, encoding: .utf8) else {
+                    return
+                }
+                self.currentSessionType = self.currentSessionType.next
+                
+                Log.info(payload)
+                // Payload format is ` Some string: ${number_of_entries_expected} `
```
pawelgil, line 51: this comment should be moved to the place it belongs

AirCasting/ABConnector/SDCardAirBeamServices.swift
```diff
@@ -0,0 +1,104 @@
...
+    
+    private var dataCharacteristicObserver: AnyHashable?
+    private var metadataCharacteristicObserver: AnyHashable?
+    
+    init(bluetoothManager: BluetoothManager) {
+        self.bluetoothManager = bluetoothManager
+    }
+    
+    func downloadData(from peripheral: CBPeripheral, progress: @escaping (SDCardDataChunk) -> Void, completion: @escaping (Result<Void, Error>) -> Void) {
+        metadataCharacteristicObserver = bluetoothManager.subscribeToCharacteristic(DOWNLOAD_META_DATA_FROM_SD_CARD_CHARACTERISTIC_UUID) { result in
```
pawelgil, line 42: let's split this function into smaller ones

AirCasting/ABConnector/SDSyncController.swift
```diff
@@ -0,0 +1,28 @@
...
+    
+    init(airbeamServices: SDCardAirBeamServices, fileWriter: SDSyncFileWriter) {
+        self.airbeamServices = airbeamServices
+        self.fileWriter = fileWriter
+    }
+    
+    func syncFromAirbeam(_ airbeamConnection: CBPeripheral, completion: @escaping (Bool) -> Void) {
+        airbeamServices.downloadData(from: airbeamConnection, progress: { [weak self] chunk in
+            // Filesystem write
+            self?.fileWriter.writeToFile(data: chunk.payload, sessionType: chunk.sessionType)
```
pawelgil, line 20: note that this writes are synchronous. I think we should schedule them to a different thread so that it won't block the bluetooth connection

pawelgil (commented):
I know this is more of a draft PR (maybe we should make it a draft for now?), but added some thoughts already

PR #448, opened 2021-11-26
FileLineReader introduced

pawelgil:
## What was done?

We needed a highly memory efficient way of reading files one by one in order to perform SD card synchronization. This PR provides an interface and an implementation for this.

I tested the performance and it looks like it's constant no matter the size of the file and is ~4KB


anna1901 (approved):
Works beautifully!

smialko (approved):
Looks :+1:

jankrzempek (approved):

PR #450, opened 2021-11-29
app loses GPS - background

jankrzempek:
https://trello.com/c/I2lEvikR/435-mobile-active-app-loses-gps-fix-when-in-background


anna1901 (approved):
Seems good

PR #456, opened 2021-11-29
[SD] added button for sync and block BT4

jankrzempek:
_In this ticket I am going to add 3 tickets:_ (I was thinking about it, and decided that this is the best solution)

**What was done/ can be tested:**
**1.** Button on "+" screen added, named "Sync Data from AirBeam"
**3.** Handled case with no internet -> Shows pop up to the user
**4.** Changes alerts structure a bit. The main reason for that is handling many alerts in one view - it is hard to make in another way.

https://trello.com/c/v9GZOVaf/316-lets-begin-always-show-sync-data-from-airbeam-3-option

https://trello.com/c/Hnp1969y/151-sd-sync-error-view-when-theres-no-internet-connection


AirCasting/CreateSessionViews/ChooseSessionType/ChooseSessionTypeView.swift
```diff
@@ -260,8 +288,24 @@ struct ChooseSessionTypeView: View {
...
+            Text(Strings.ChooseSessionTypeView.syncTitle)
+                .font(Fonts.boldHeading1)
+                .foregroundColor(.accentColor)
+            Text(Strings.ChooseSessionTypeView.syncDescription)
+                .font(Fonts.muliHeading3)
+                .foregroundColor(.aircastingGray)
+        }
+        .multilineTextAlignment(.leading)
+        .padding(15)
+        .frame(maxWidth: 147, maxHeight: 145)
```
anna1901, line 308: For some reason this button is wider than the rest fo the buttons

AirCasting/Helpers/DeviceHandler.swift
```diff
@@ -0,0 +1,65 @@
...
+    
+    static func notSupportedAlert() -> Alert {
+        Alert(
+            title: Text(Strings.DeviceHandler.alertTitle),
+            message: Text(Strings.DeviceHandler.alertMessage),
+            dismissButton: .default(Text(Strings.DeviceHandler.continueText))
+        )
+    }
+    
+    static func getDeviceNumber() -> Int {
```
anna1901, line 53: Is this function used anywhere?

AirCasting/SessionViews/StandaloneSessionCardView.swift
```diff
@@ -46,52 +46,67 @@ struct StandaloneSessionCardView: View {
...
+                             dismissButton: .default(alert.buttonTitle))
+            }
+        })
         .padding()
     }
     
     var finishAndSyncButton: some View {
         Button(Strings.StandaloneSessionCardView.finishAndSyncButtonLabel) {
-            showingFinishAndSyncAlert = true
+            if networkChecker.connectionAvailable {
```
anna1901, line 77: This seems to allow to run sync when there is no wifi but there is a cellular internet. Could you make sure if that's what we want to happen? I'm remember vaguely Michael didn't want to run sync on cellular data automatically, without user knowing, but we should double check.

anna1901 (changes requested):
Alerts work beautifully! I just added a couple of questions

AirCasting/Helpers/DeviceHandler.swift
```diff
@@ -0,0 +1,65 @@
...
+                                            .iPadAir2,
+                                            .iPadMini2,
+                                            .iPadMini3,
+                                            .iPadMini4,
+                                            .iPadPro9Inch,
+                                            .iPadPro10Inch,
+                                            .iPadPro12Inch,
+                                            .iPadPro12Inch2]
+    
+    static var disableButtonBasedOnBT: Bool {
```
pawelgil, line 41: Two things:
* What button? The name is unclear
* It's not great that UI stuff is here -> maybe let's change this to isBluetooth5Enabled and switch condition in View

AirCasting/Helpers/DeviceHandler.swift
```diff
@@ -0,0 +1,65 @@
...
+        DeviceHandler.BTBelow5Devices.contains(Device.current)
+    }
+    
+    static func notSupportedAlert() -> Alert {
```
pawelgil, line 45: It's not great that UI stuff is here -> maybe let's just move it to View?

If we have many reused alerts we can create a new struct like `AppAlerts` so it's clear and use it lie:
`AppAlerts.unsupportedBluetoothAlert()`, but I don't think we're reusing alerts ?

AirCasting/SessionViews/SessionHeaderView.swift
```diff
@@ -30,7 +30,22 @@ struct SessionHeaderView: View {
...
+        }
+        .alert(item: $alert, content: { alert in
+            if alert.id == .finishSessionAlert {
+                return Alert(title: alert.title,
+                             message: alert.message,
+                             primaryButton: .default(alert.buttonTitle, action: {
+                    finishSessionAlertAction(sessionStopper: sessionStopperFactory.getSessionStopper(for: session))
+                }),
+                             secondaryButton: .cancel())
+            } else {
```
pawelgil, line 42: else is not needed since you're returning inside `if` anyway

AirCasting/SessionViews/SessionHeaderView.swift
```diff
@@ -200,6 +208,14 @@ private extension SessionHeaderView {
         let string = formatter.string(from: start, to: end)
         return Text(string)
         }
+    
+    func finishSessionAlertAction(sessionStopper: SessionStoppable) {
```
pawelgil, line 212: private

AirCasting/SessionViews/StandaloneSessionCardView.swift
```diff
@@ -46,52 +46,67 @@ struct StandaloneSessionCardView: View {
...
+                    finishSessionAlertAction(sessionStopper: sessionStopperFactory.getSessionStopper(for: session))
+                }),
+                             secondaryButton: .cancel())
+            } else if alert.id == .finishSessionAndSyncAlert {
+                return Alert(title: alert.title,
+                             message: alert.message,
+                             primaryButton: .default(alert.buttonTitle, action: {
+                    finishSessionAndSyncAlertAction()
+                }), secondaryButton: .cancel())
+            } else {
```
pawelgil, line 66: this last `else` is not needed

AirCasting/Utils/InAppAlerts.swift
```diff
@@ -0,0 +1,60 @@
+// Created by Lunar on 01/12/2021.
+//
+
+import SwiftUI
+// This struct provides all nedded alerts in case of need. Using them from here, allows you
+// to place multiple instances of them (in one view) and every one of them will be called and show up nicely.
+// Key thing here - is the id.
+// Use them, have fun, ALERT the user 🛎
+struct AlertInfo: Identifiable {
+    enum AlertType {
```
pawelgil, line 10: Maybe instead of putting app-specific alert types here that would need to be handled inside Views (multiple if-else-fallback scenarios) we can further generalize it by adding `let buttons: [Button]` here and removing `buttonTitle` and the `AlertType`:
```Swift
struct AlertInfo: Identifiable {
    struct Button {
        enum ButtonType { case cancel, ok, destructive }
        let title: String
        let type: ButtonType
        let action: (() -> Void)?
    }
    ...
    let buttons: [Button]
    ...
}
```

and provide needed action blocks inside the static factory methods:

```Swift
static func finishAndSyncAlert(sessionName: String, syncAction: () -> Void) { ... }
```

we can than add an alert factory here too:

```Swift
extension AlertInfo {
   func makeAlert() -> Alert { ... }
}
```

so that inside view we only have to do this:
```Swift
.alert(item: $alert, content: { $0.makeAlert() }
```

pawelgil (changes requested):

AirCasting/SessionViews/StandaloneSessionCardView.swift
anna1901: Nevermind, I asked Michael and it is okay to allow users to sync over cellular data

.gitignore
```diff
@@ -14,6 +14,9 @@ xcuserdata/
 *.xcscmblueprint
 *.xccheckout
 
+.env
```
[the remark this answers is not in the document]
jankrzempek: I assume we have discussed it 😇

AirCasting/Helpers/DeviceHandler.swift
jankrzempek: Not yer, but I wrote it and it could be useful so I would like it to stay :)

AirCasting/Helpers/DeviceHandler.swift
```diff
@@ -0,0 +1,65 @@
...
+        Alert(
+            title: Text(Strings.DeviceHandler.alertTitle),
+            message: Text(Strings.DeviceHandler.alertMessage),
+            dismissButton: .default(Text(Strings.DeviceHandler.continueText))
+        )
+    }
+    
+    static func getDeviceNumber() -> Int {
+        if Device.current.isPhone {
```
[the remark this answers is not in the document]
jankrzempek: Sure!

AirCasting/CreateSessionViews/ChooseSessionType/ChooseSessionTypeView.swift
jankrzempek: Screenshot 2021-12-03 at 16 59 48

jankrzempek: It's the same I think and I checked

PR #449, opened 2021-11-27
Firebase config turned on

pawelgil:
## ‼️ Important testing note ‼️
In `DEBUG` builds the app is using `AllFeaturesOn` implementation, not Firebase!
If you want to test this in `DEBUG` build, please temporarily change contents of the `FeatureFlagsViewModel` to point to firebase instead.

## What was done?
* Required entitlement added
* Background modes added
* Firebase SDK updated
* FIrebase function added to the repo

PR #460, opened 2021-11-30
Possibility to override config in beta builds

pawelgil:
Beta (firebase) testers will be able to override configuration using settings menu:

Note: needs #449 to undraft


anna1901 (approved):
Awesome 🚀

PR #466, opened 2021-12-02
[SD] [sd-main-branch <--- this one] SD Sync - views

jankrzempek:
**What was done/ can be tested:**

- Added View flow to SD sync process.
- Added ViewModels to make it as clean as possible and ready for future logic or code expansion 💂‍♀️

Please test the whole View flow, as without AirBeam it is hard to go through BT screen

Ticket: https://trello.com/c/caxWvcY8/426-sync-sd-card-views


.gitignore
```diff
@@ -10,6 +10,9 @@
 ## User settings
 xcuserdata/
 
+.env
```
anna1901, line 13: This is included in your other PR. Is it added intentionally?

AirCasting/CreateSessionViews/SelectPeripheralView.swift
```diff
@@ -136,7 +136,7 @@ struct SelectPeripheralView: View {
                                        userAuthenticationSession: userAuthenticationSession,
                                        sessionContext: sessionContext,
                                        peripheral: selection)
-                destination = AnyView(SyncingABView(viewModel: viewModel, baseURL: urlProvider, creatingSessionFlowContinues: $creatingSessionFlowContinues))
+                destination = AnyView(SDSyncABView(viewModel: viewModel, creatingSessionFlowContinues: $creatingSessionFlowContinues))
```
anna1901, line 139: I named it SyncingABView because it is the view which is showed when the AB is syncing. SDSyncABView seems a bit less descriptive, maybe we could find a name which is a mix of those two? :D

AirCasting/SDSync/ViewsFlow/SDSyncRootView/SDSyncRootView.swift
```diff
@@ -0,0 +1,85 @@
...
+                messageLabel
+            }
+            Spacer()
+        }
+        .padding()
+        .background(navigationLink)
+        .onAppear() {
+            viewModel.onAppearExecute()
+            finishAndSyncButtonTapped.finishAndSyncButtonWasTapped = false
+            viewModel.startBackendSync()
```
anna1901, line 33: I don't understand the logic here. Why are we calling onAppearExecute, which is checking if there is a sync in progress, and if there is, it is calling startBackendSync when it ends and then we are calling the stantBackedSync again from here? As I understand, it could happen that we would call sync twice?

AirCasting/SDSync/ViewsFlow/SDSyncSuccess/SDSyncSuccessView.swift
```diff
@@ -0,0 +1,69 @@
+// Created by Lunar on 01/12/2021.
+//
+
+import AirCastingStyling
+import SwiftUI
+
+struct SDSyncSuccessView<VM: SDSyncSuccessViewModel>: View {
```
anna1901, line 7: It the view which shows that the whole SDSync was successful or that the backend sync was completed?

anna1901 (commented):
It goes through the views nicely 🙂  Great job, the views look great! 😊 👍 🚀   I just added a couple of questions

AirCasting/CreateSessionViews/ChooseSessionType/ChooseSessionTypeView.swift
```diff
@@ -37,7 +37,7 @@ struct ChooseSessionTypeView: View {
                 mainContent
                 .fullScreenCover(isPresented: $isPowerABLinkActive) {
                     CreatingSessionFlowRootView {
-                        PowerABView(creatingSessionFlowContinues: $isPowerABLinkActive, sdSyncContinues: .constant(false), urlProvider: viewModel.passURLProvider)
+                        PowerABView(creatingSessionFlowContinues: $isPowerABLinkActive, urlProvider: viewModel.passURLProvider)
```
anna1901, line 40: Since creatingSessionFlowContinues is used also in syncing flow maybe it should have more generic name? Like "viewsFlowContinues"? or "wizzardViewsContinue"? I'm not very happy with those names either, but keeping it as "creatingSessionFlowContinues" is a bit misleading probably

anna1901: However, we can think about the name some other time

AirCasting/SDSync/ViewsFlow/SDSyncingAB/SyncingABView.swift
```diff
@@ -90,12 +79,21 @@ struct SyncingABView<VM: SDSyncViewModel>: View {
         }
     }
     
-    var completeView: some View {
-        VStack {
-            Text("Complete")
-            Button("Finish") {
-                sdSyncContinues = false
+    var connectionTimeOutAlert: Alert {
```
anna1901, line 82: Do we need this?

anna1901: Ah, this should probably be a "Faild to sync" alert

anna1901 (approved):
Nice!! 🎆

AirCasting/SDSync/ViewsFlow/SDSyncRootView/SDSyncRootViewModel.swift
```diff
@@ -0,0 +1,68 @@
...
+    private func startBackendSync() {
+        sessionSynchronizer.triggerSynchronization() {
+            DispatchQueue.main.async {
+                self.backendSyncCompleted = true
+            }
+        }
+    }
+    
+    private func onCurrentSyncEnd(_ completion: @escaping () -> Void) {
+        guard sessionSynchronizer.syncInProgress.value else { completion(); return }
```
pawelgil, line 41: this pattern is used in so many places right now that maybe we can move it to extension or smth and replace all?

AirCasting/SDSync/ViewsFlow/SDSyncSuccess/BackendSyncCompletedView.swift
```diff
@@ -0,0 +1,69 @@
...
+    
+    var messageLabel: some View {
+        Text(Strings.SDSyncSuccessView.message)
+            .font(Fonts.regularHeading1)
+            .foregroundColor(.aircastingGray)
+    }
+    
+    var continueButton: some View {
+        Button {
+            viewModel.presentNextScreen = true
```
pawelgil, line 55: same as in other VM - this should be written to internally and only read in view (just like you did with `executeBackendSync()` and `backendSyncCompleted`. The only difference is that here implementation would be trivial one-liner)

AirCasting/SDSync/ViewsFlow/SDSyncSuccess/BackendSyncCompletedViewModel.swift
```diff
@@ -0,0 +1,19 @@
+// Created by Lunar on 02/12/2021.
+//
+
+import Foundation
+
+protocol BackendSyncCompletedViewModel: ObservableObject {
+    var presentNextScreen: Bool { get set }
+    var urlProvider: BaseURLProvider { get }
```
pawelgil, line 8: as in others - let's comment

AirCasting/SDSync/ViewsFlow/SDSyncingAB/SDSyncViewModel.swift
```diff
@@ -7,15 +7,20 @@ import CoreBluetooth
 protocol SDSyncViewModel: ObservableObject {
     var shouldDismiss: Published<Bool>.Publisher { get }
     var isSyncCompleted: Published<Bool>.Publisher { get }
+    var presentNextScreen: Bool { get set }
+    var presentAlert: Bool { get set }
```
pawelgil, line 11: unclear name

AirCasting/SDSync/ViewsFlow/SDSyncingAB/SyncingABView.swift
```diff
@@ -5,76 +5,65 @@
...
-                }
-            )
-        )
+            Spacer()
+        }
         .padding()
+        .background(navigationLink)
         .onReceive(viewModel.isSyncCompleted, perform: { isConnected in
-            showNextScreen = isConnected
+            viewModel.presentNextScreen = isConnected
```
pawelgil, line 35: same as with others - let's make it into a func on VM that does that write instead.

pawelgil (commented):
Overall very nice code. Especially `View`s are really small and neatly done. Some refinements are required, but really good job

AirCasting/SDSync/ViewsFlow/SDRestartAB/SDRestartABView.swift
```diff
@@ -0,0 +1,71 @@
+// Created by Lunar on 01/12/2021.
+//
+
+import AirCastingStyling
+import SwiftUI
+
+struct SDRestartABView<VM: SDRestartABViewModel>: View {
+    @StateObject var viewModel: VM
+    @Binding var creatingSessionFlowContinues: Bool
+    @EnvironmentObject private var sessionContext: CreateSessionContext
```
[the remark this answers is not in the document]
jankrzempek: **!! The below explanation is based on my knowledge and represents my probable lack of knowledge. I assume somehow it should be possible if you says it is possible - so if I am wrong could you please tell me how to do it?**

I am really not sure how to do this\ have some doubts. I like to move everything to VM as long as it is not @Binding or @EnviromentObject. I think that the whole purpose of this is to use it anywhere you want and if we will switch this @EnviromentObject to VM then we will need to use SwiftUI there (I think we don't want to) or pass this through the whole ViewFlow to make it work. That's the reason why it is here for now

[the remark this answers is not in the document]
pawelgil: Well, this whole environment object thing is a bit tricky - it's easy to start using it as dependency injection pattern which is defenitelly not good (but we have such code sadly, it needs to go at some point). Here it's resolving stuff that should really be handled by the VM - this line in particular: `sessionContext.deviceType = .AIRBEAM3`. It's 100% VM thing. I understand that it's difficult to move now... Maybe we can pass CreateSessionContext into VM in init? If not - let's skip this one.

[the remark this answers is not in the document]
anna1901: @jankrzempek I think we don't need to do anything with sessionContext in this view. We only need it in the creating session flow, so we can remove it from here altogether.

PR #476, opened 2021-12-05
Sd sync progress indication

pawelgil:
Added progress indication to SD synchronization API. 
Note that this has not been tested in any way, putting it here because interfaces have changed.

### DIFF
--- a/AirCasting/ABConnector/SDCardAirBeamServices.swift
+++ b/AirCasting/ABConnector/SDCardAirBeamServices.swift
@@ -99,7 +99,7 @@ class BluetoothSDCardAirBeamServices: SDCardAirBeamServices {
                     return
                 }
                 
-                let receivedMeasurementsNumber = receivedMeasurementsCount[sessionType]! < expectedMeasurementsCount  ? receivedMeasurementsCount[sessionType]! : expectedMeasurementsCount
+                let receivedMeasurementsNumber = receivedMeasurementsCount[sessionType]! < expectedMeasurementsCount ? receivedMeasurementsCount[sessionType]! : expectedMeasurementsCount
                 let progressFraction = SDCardProgress(received: receivedMeasurementsNumber, expected: expectedMeasurementsCount)
                 progress(SDCardDataChunk(payload: payload, sessionType: sessionType, progress: progressFraction))
 
--- a/AirCasting/ABConnector/SDCardAirBeamServices.swift
+++ b/AirCasting/ABConnector/SDCardAirBeamServices.swift
@@ -4,11 +4,6 @@
 import CoreBluetooth
 import CoreMIDI
 
-enum SDCardData {
-    case chunk(SDCardDataChunk)
-    case metadata(SDCardMetaData)
-}
-
 enum SDCardSyncError: Error {
     case cantDecodePayload
     case wrongOrderOfReceivedPayload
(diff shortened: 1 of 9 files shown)


AirCasting/ABConnector/SDCardAirBeamServices.swift
```diff
@@ -94,8 +87,9 @@ class BluetoothSDCardAirBeamServices: SDCardAirBeamServices {
                     self.finishSync { completion(.failure(SDCardSyncError.wrongOrderOfReceivedPayload)) }
                     return
                 }
-                progress(.chunk(SDCardDataChunk(payload: payload, sessionType: sessionType)))
-                
+                receivedMeasurementsCount[sessionType, default: 0] += 1
```
anna1901, line 90: I think this should be: += Constants.SDCardSync.numberOfMeasurementsInDataChunk

AirCasting/ABConnector/SDCardAirBeamServices.swift
```diff
@@ -94,8 +87,9 @@ class BluetoothSDCardAirBeamServices: SDCardAirBeamServices {
...
+                let progressFraction = Double(receivedMeasurementsCount[sessionType]!) / Double(expectedMeasurementsCount[sessionType].orOne)
```
anna1901, line 91: Hmmm... Shouldn't we check here if `expectedMeasurementsCount[sessionType] != 0 ` and if it is, then log an error? Cause if we didn't expect to receive measurements for specific session type, but we did, then we should be aware of this issue.

anna1901: Additionally, this is not exactly what we want to show in the view. We want to show eg. "1/10000" so basically what we have here but as a string and not the value of calculation, cause now it would look like: "mobile: 0.8794234832" etc.

anna1901 (changes requested):
Seems to be working very nicely, great job. 👏  I just have a few comments

anna1901 (approved):

AirCasting/ABConnector/SDCardAirBeamServices.swift
```diff
@@ -94,8 +92,16 @@ class BluetoothSDCardAirBeamServices: SDCardAirBeamServices {
...
                 }
-                progress(.chunk(SDCardDataChunk(payload: payload, sessionType: sessionType)))
+                receivedMeasurementsCount[sessionType, default: 0] += Constants.SDCardSync.numberOfMeasurementsInDataChunk
+                
+                guard let expectedMeasurementsCount = expectedMeasurementsCount[sessionType], expectedMeasurementsCount != 0 else {
+                    Log.error("[SD SYNC] Received data for session type which should have 0 measurements")
+                    return
+                }
                 
+                let receivedMeasurementsNumber = receivedMeasurementsCount[sessionType]! < expectedMeasurementsCount  ? receivedMeasurementsCount[sessionType]! : expectedMeasurementsCount
```
jankrzempek, line 102: Two spaces here` sCount  ? receivedM`, please delete one

AirCasting/SDSync/ViewsFlow/SDSyncingAB/SDSyncViewModel.swift
```diff
@@ -4,21 +4,33 @@
 import Combine
 import CoreBluetooth
 
+struct SDSyncProgressViewModel {
+    let sessionType: String
+    let progress: SDCardProgress
+    
+    func toString() -> String {
+        "Syncing " + sessionType + ": " + "\(progress.received)/\(progress.expected)"
```
jankrzempek, line 12: Do we want to put it into Strings?

AirCasting/SDSync/ViewsFlow/SDSyncingAB/SDSyncViewModel.swift
```diff
@@ -5,8 +5,12 @@ import Combine
 import CoreBluetooth
 
 struct SDSyncProgressViewModel {
-    let label: String
-    let value: String
+    let sessionType: String
+    let progress: SDCardProgress
+    
+    func toString() -> String {
```
pawelgil, line 11: won't be needed after changes from previous comment - this will be done inside `SDSyncViewModel`

PR #478, opened 2021-12-06
added sync after delete

jankrzempek:
https://trello.com/c/HR617RIB/134-mobile-dormant-fixed-delete-session

Michael wanted in the latest comment

### DIFF
--- a/AirCasting/SessionViews/SessionCartViewModel/DefaultDeleteSessionViewModel.swift
+++ b/AirCasting/SessionViews/SessionCartViewModel/DefaultDeleteSessionViewModel.swift
@@ -2,11 +2,13 @@
 //
 
 import Foundation
+import Combine
 
 class DefaultDeleteSessionViewModel: DeleteSessionViewModel {
     private let measurementStreamStorage: MeasurementStreamStorage
     private var session: SessionEntity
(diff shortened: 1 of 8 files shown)


AirCasting/SessionViews/SessionCartViewModel/DefaultDeleteSessionViewModel.swift
```diff
@@ -74,6 +77,16 @@ class DefaultDeleteSessionViewModel: DeleteSessionViewModel {
             } catch {
                 Log.info("Error when deleting sessions/streams")
             }
+            if sessionSynchronizer.syncInProgress.value {
```
anna1901, line 80: I think this could be extracted to a separate function. Additionally, I think if could be changed to guard, which indicates early return.

anna1901 (approved):
Just one suggestion, otherwise looks good to me 👍 🚀

AirCasting/SessionViews/SessionCartViewModel/DefaultDeleteSessionViewModel.swift
jankrzempek: Yeah, great catch - thanks! I've changed both then 😁

PR #480, opened 2021-12-06
Show "enter standalone mode" button only for active AB sessions and remove menu from standalone session card

anna1901:


AirCasting/SessionViews/SessionHeaderView.swift
```diff
@@ -11,6 +11,7 @@ struct SessionHeaderView: View {
     let action: () -> Void
     let isExpandButtonNeeded: Bool
     var isSensorTypeNeeded: Bool = true
+    var isMenuNeeded = true
```
jankrzempek, line 14: It is never used as I can see. Can we delete it?

jankrzempek (changes requested):
One comment here 😄

AirCasting/SessionViews/SessionHeaderView.swift
anna1901: It is necessary and used in line 34

jankrzempek: but it will always be true, won't it?

AirCasting/SessionViews/StandaloneSessionCardView.swift
```diff
@@ -33,6 +33,7 @@ struct StandaloneSessionCardView: View {
         SessionHeaderView(
             action: {},
             isExpandButtonNeeded: false,
+            isMenuNeeded: false,
```
anna1901, line 36: @jankrzempek here I'm setting it to false

jankrzempek: ooh, sorry then. Approve right away!

jankrzempek (approved):

PR #483, opened 2021-12-06
Sd clear - settings

jankrzempek:


AirCasting/CreateSessionViews/SelectPeripheralView.swift
```diff
@@ -95,7 +96,7 @@ struct SelectPeripheralView: View {
     
     var titleLabel: some View {
         var title: Text
-        if syncMode == true {
+        if syncMode == true || SDClearingRouteProcess {
             title = Text(Strings.SelectPeripheralView.titleSyncLabel)
```
anna1901, line 100: Should this be the same for clearing SD card? "Select the device you'd like to sync"?

AirCasting/CreateSessionViews/SelectPeripheralView.swift
```diff
@@ -95,7 +96,7 @@ struct SelectPeripheralView: View {
     
     var titleLabel: some View {
         var title: Text
-        if syncMode == true {
+        if syncMode == true || SDClearingRouteProcess {
```
anna1901, line 99: I realise it's mine, but could you please delete this "== true"?

AirCasting/CreateSessionViews/TurnOnBluetoothView.swift
```diff
@@ -11,10 +11,12 @@ import SwiftUI
 
 struct TurnOnBluetoothView: View {
     @State private var isPowerABLinkActive = false
+    @State private var SDRestartABLink = false
     @EnvironmentObject var settingsRedirection: DefaultSettingsRedirection
     @EnvironmentObject var bluetoothManager: BluetoothManager
     @Binding var creatingSessionFlowContinues: Bool
     @Binding var sdSyncContinues: Bool
+    var isSDClearProcess: Bool
```
anna1901, line 19: Maybe this could also default to "false"? This way we wouldn't have to set it to false in all of those views above.

AirCasting/SDSync/ViewsFlow/SDRestartAB/SDRestartABView.swift
```diff
@@ -54,15 +57,15 @@ extension SDRestartABView {
     
     var continueButton: some View {
         Button {
-            viewModel.presentNextScreen = true
+            viewModel.presentNextScreen.toggle()
```
anna1901, line 60: Why did you change "= true" to toggle?

AirCasting/Settings/SettingsView.swift
```diff
@@ -9,29 +9,58 @@ import SwiftUI
...
     }
     
     var body: some View {
         NavigationView {
             Form {
                 signOutSection
                 settingsSection
+                #if BETA || DEBUG
+                Section() {
+                    navigateToAppConfigurationButton
```
anna1901, line 40: Why is this in this PR? 🤔

AirCasting/Settings/SettingsView.swift
```diff
@@ -9,29 +9,58 @@ import SwiftUI
...
             }
             .listStyle(GroupedListStyle())
             .navigationBarTitle(Strings.Settings.title)
         }
+        .fullScreenCover(isPresented: $startSDClear) {
+            CreatingSessionFlowRootView {
+                SDRestartABView(viewModel: SDRestartABViewModelDefault(urlProvider: urlProvider, isSDClearProcess: SDClearingRouteProcess), creatingSessionFlowContinues: $startSDClear)
+            }
+        }
+        .fullScreenCover(isPresented: $LocationScreenGo) {
```
anna1901, line 53: This only works for ios15 unfortunately, so we need to check if it's ios15 or not and include that other implementation as well

AirCasting/SDSync/ViewsFlow/SDRestartAB/SDRestartABView.swift
jankrzempek: It seems better/cleaner in my opinion and work as expected :)

AirCasting/CreateSessionViews/SelectPeripheralView.swift
jankrzempek: Great catch, I will make a change then

AirCasting/CreateSessionViews/SelectPeripheralView.swift
jankrzempek: As long as syncMode is optional, unfortunately I cannot do it

AirCasting/SDSync/ViewsFlow/ClearingSDCard/ClearingSDCardViewModel.swift
```diff
@@ -1,18 +1,95 @@
...
+}
 
 protocol ClearingSDCardViewModel: ObservableObject {
+    var isClearingCompleted: Published<Bool>.Publisher { get }
+    var shouldDismiss: Published<Bool>.Publisher { get }
     var presentNextScreen: Bool { get set }
     var isSDClearProcess: Bool { get set }
+    func clearSDCard()
+    func getAlertTitle() -> String
+    func getAlertMessage() -> String
```
pawelgil, line 19: using our new system we can do `var alert: AlertInfo { get }` (if it's merged in, not sure).

If not, maybe we can omit `get` here and make it as `var`? We can use computed vars in implementation part

AirCasting/SDSync/ViewsFlow/ClearingSDCard/ClearingSDCardViewModel.swift
```diff
@@ -1,18 +1,95 @@
...
+    case noConnection
+    case noClearing
+}
 
 protocol ClearingSDCardViewModel: ObservableObject {
+    var isClearingCompleted: Published<Bool>.Publisher { get }
+    var shouldDismiss: Published<Bool>.Publisher { get }
     var presentNextScreen: Bool { get set }
     var isSDClearProcess: Bool { get set }
+    func clearSDCard()
```
pawelgil, line 17: like I wrote in Janek's PR - viewmodel should not be asked to do a certain thing by the view. It should only be informed what happend view-wise (like user tapped something, switched a switch, stuff like that). So for example here it should be phrased more like `clearSDCardButtonTapped()` or smth like that.

AirCasting/SDSync/ViewsFlow/ClearingSDCard/ClearingSDCardView.swift
```diff
@@ -68,6 +77,14 @@ extension ClearingSDCardView {
         }
     }
     
+    var alert: Alert {
```
pawelgil, line 80: can use new alert system instead (hope it's merged in)

AirCasting/SDSync/ViewsFlow/ClearingSDCard/ClearingSDCardView.swift
```diff
@@ -28,10 +30,17 @@ struct ClearingSDCardView<VM: ClearingSDCardViewModel>: View {
...
+        }
         .onAppear(perform: {
-            DispatchQueue.main.asyncAfter(deadline: .now() + .seconds(10)) {
-                viewModel.presentNextScreen.toggle()
-            }
+            viewModel.clearSDCard()
+        })
+        .onReceive(viewModel.isClearingCompleted, perform: { result in
+            viewModel.presentNextScreen = result
+        })
```
pawelgil, line 41: this `onReceive` is not needed. It is subscribed to viewmodel only to switch execution back into viewmodel. So it can be fully contained inside viewmodel implementation instead.

AirCasting/ABConnector/SDCardAirBeamServices.swift
```diff
@@ -104,7 +105,33 @@ class BluetoothSDCardAirBeamServices: SDCardAirBeamServices {
         }
     }
     
-    func finishSync(completion: () -> Void) {
+    func clearSDCard(of peripheral: CBPeripheral, completion: @escaping (Result<Void, Error>) -> Void) {
+        metadataCharacteristicObserver = bluetoothManager.subscribeToCharacteristic(DOWNLOAD_META_DATA_FROM_SD_CARD_CHARACTERISTIC_UUID) { result in
```
pawelgil, line 109: we are using the same `var` for `downloadData`. It can lead to problems (like if we invoke download and clear one after another the clear will overwrite the old token and it won't get unsubscribed ever). I think it's not intended?

AirCasting/SDSync/ViewsFlow/ClearingSDCard/ClearingSDCardViewModel.swift
```diff
@@ -1,18 +1,95 @@
...
+            return Strings.ClearingSDCardView.failedClearingAlertMessage
+        case .noConnection:
+            return Strings.AirBeamConnector.connectionTimeoutDescription
+        case .noClearing:
+            return Strings.ClearingSDCardView.failedClearingAlertMessage
+        }
+    }
+    
+    private func sendClearConfig() {
+        let configurator = AirBeam3Configurator(userAuthenticationSession: self.userAuthenticationSession,
```
pawelgil, line 91: I'd change it so that `BluetoothSDCardAirBeamServices` sends those configurations. Both for download and for erase. We want only super high level stuff here, not some obscure configuration sending. Would that be possible?

pawelgil (changes requested):

AirCasting/SDSync/ViewsFlow/ClearingSDCard/ClearingSDCardViewModel.swift
jankrzempek: As long as it is used in the .onAppear modifier I am really not sure @pawelgil. Does `clearSDCardButtonTapped()` still work here?

AirCasting/ABConnector/SDCardAirBeamServices.swift
anna1901: Very good catch, thank you!

pawelgil (approved):

PR #484, opened 2021-12-06
Fastlane and feature flags short desctiption added to README

pawelgil:


anna1901 (approved):
Nice! Maybe we could also include step by step informations about releasing like: "run `fastlane beta` in terminal.

jankrzempek (approved):
🚀

PR #489, opened 2021-12-08
Show finalizing view for sd sync

anna1901:
https://trello.com/c/JxCEN4rV


AirCasting/SDSync/ViewsFlow/SDSyncingAB/SyncingABView.swift
```diff
@@ -61,7 +65,14 @@ extension SyncingABView {
     }
     
     var titleLabel: some View {
-        Text(progressLabel ?? "")
+        VStack(alignment: .leading) {
+            if viewModel.isDownloadingFinished {
+                Text("Finalizing...")
+            } else {
+                progressTitle != nil ? Text("Syncing " + progressTitle!.lowercased()) : Text("Syncing...")
+                Text(progressCount ?? "")
```
jankrzempek, line 73: Strings please 😄

jankrzempek (commented):
Only those Strings please

AirCasting/ABConnector/SDSyncController.swift
```diff
@@ -32,7 +32,7 @@ class SDSyncController: ObservableObject {
         self.sessionSynchronizer = sessionSynchronizer
     }
     
-    func syncFromAirbeam(_ airbeamConnection: CBPeripheral, progress: @escaping (SDCardSyncProgess) -> Void, completion: @escaping (Bool) -> Void) {
+    func syncFromAirbeam(_ airbeamConnection: CBPeripheral, progress: @escaping (SDCardSyncProgess) -> Void, finalizing: @escaping () -> Void, completion: @escaping (Bool) -> Void) {
```
pawelgil, line 35: I think maybe switching from SDCardSyncProgress to something like:
```Swift
enum SDCardSyncStatus {
    case inProgress(SDCardSyncProgress)
    case finalizing
}
```
and having just a `progress` and `completion` blocks will be better? It looks a bit odd, I didn't quite get what is a difference between finalizing and completion at first

pawelgil (approved):
I left a comment with suggestion, but leaving it as approved because it's optional

PR #495, opened 2021-12-09
Upload synced fixed sessions 

smialko:


AirCasting/SDSync/FixedSessions/SDCardFixedSessionsSavingService.swift
```diff
@@ -0,0 +1,64 @@
...
+        csvSession.sessions.map { session -> [UploadFixedSessionAPIService.UploadFixedMeasurementsParams] in
+            let sessionStreams = csvSession.streamsWithMeasurements.filter { $0.key.sessionUUID == session.uuid }
+            
+            let perStreamParams = sessionStreams.map { sdStream, measurements -> UploadFixedSessionAPIService.UploadFixedMeasurementsParams in
+                let stream = CSVMeasurementStream.SUPPORTED_STREAMS[sdStream.header]!
+                let MILLISECONDS_IN_SECOND = 1000
+                let csvMeasurements = measurements.map {
+                    CSVMeasurement(longitude: $0.location?.longitude,
+                                   latitude: $0.location?.latitude,
+                                   milliseconds: 0,//Int($0.time.timeIntervalSince1970.remainder(dividingBy: TimeInterval(MILLISECONDS_IN_SECOND))),
```
smialko, line 38: Uncomment this 🙃

AirCasting/ABConnector/SDSyncController.swift
```diff
@@ -57,31 +61,61 @@ class SDSyncController: ObservableObject {
...
-                                }
-                                completion(true)
-                            case .failure(let error):
-                                Log.error("Failed to save sessions to database: \(error.localizedDescription)")
-                                completion(false)
-                            }
+                    self.checkFilesForCorruption(files, expectedMeasurementsCount: metadata.expectedMeasurementsCount) { result in
+                        switch result {
+                        case .success:
+                            self.handle(files: files, deviceID: sensorName)
```
anna1901, line 67: You forgot about calling completion and views never dismiss

AirCasting/ABConnector/SDSyncController.swift
```diff
@@ -57,31 +61,61 @@ class SDSyncController: ObservableObject {
...
             }
         })
     }
     
+    private func handle(files: [(URL, SDCardSessionType)], deviceID: String) {
+        let mobileFileURL = files.first(where: { $0.1 == SDCardSessionType.mobile })?.0
+        let fixedFileURL = files.first(where: { $0.1 == SDCardSessionType.fixed })?.0
+        
+        if let mobileFileURL = mobileFileURL {
+            process(mobileSessionFile: mobileFileURL, deviceID: deviceID) { _ in }
```
anna1901, line 86: Please add completion

AirCasting/ABConnector/SDSyncController.swift
```diff
@@ -57,31 +61,61 @@ class SDSyncController: ObservableObject {
...
+        }
+        if let fixedFileURL = fixedFileURL {
+            try? process(fixedSessionFile: fixedFileURL, deviceID: deviceID)
```
anna1901, line 89: Actually I think completion should be called after this is done.

AirCasting/SDSync/Models/CSVMeasurementStream.swift
```diff
@@ -0,0 +1,82 @@
...
+    let thresholdHigh: Int
+    let thresholdVeryHigh: Int
+    
+    private static let DEVICE_NAME = "AirBeam3"
+    private static let PM_MEASUREMENT_TYPE = "Particulate Matter"
+    private static let PM_MEASUREMENT_SHORT_TYPE = "PM"
+    private static let PM_UNIT_NAME = "microgram per cubic meter"
+    private static let PM_UNIT_SYMBOL = "µg/m³"
+    static let SUPPORTED_STREAMS: [SDCardCSVFileFactory.Header : CSVMeasurementStream] = [
+        SDCardCSVFileFactory.Header.f : CSVMeasurementStream(sensorName: "\(DEVICE_NAME)-F",
```
anna1901, line 24: This all is already implemented in MeasurementStream.swift

smialko: Not quite, those types are slightly different. One has `id`, other do not. Some properties are optional in MeasurementStreeam but aren't in CSVMeasurementStream.
And even if there were basically equal, I still think it's a good idea to distinguish them. Conceptually they represent different things. And our db scheme (`MeasurementStream`) might change while CSV format stays the same (`CSVMeasurementStream`).

AirCasting/SDSync/FixedSessions/SDCardFixedSessionsSavingService.swift
```diff
@@ -0,0 +1,64 @@
...
+                                                                                  threshold_very_high: stream.thresholdVeryHigh,
+                                                                                  threshold_high: stream.thresholdHigh,
+                                                                                  threshold_medium: stream.thresholdMedium,
+                                                                                  threshold_low: stream.thresholdLow,
+                                                                                  threshold_very_low: stream.thresholdVeryLow,
+                                                                                  measurements: csvMeasurements)
+            }
+            
+            Log.info("[SD Sync] Syncing \(perStreamParams.count) streams for session \(session.uuid)")
+
```
jankrzempek, line 60: Do we need those blank enters :) ?

AirCasting/SDSync/FixedSessions/UploadFixedSessionAPIService.swift
```diff
@@ -0,0 +1,95 @@
...
+    }
+    
+    fileprivate struct APIInput: Encodable {
+        let data: String
+        var compression: Bool = true
+    }
+    
+    struct APIOutput: Codable {
+        
+    }
```
jankrzempek, line 94: Is it used anywhere? It no can we delete it? If yes, can we make one-liner please?

AirCasting/SDSync/Models/CSVMeasurementStream.swift
```diff
@@ -0,0 +1,82 @@
...
+        SDCardCSVFileFactory.Header.pm2_5 : CSVMeasurementStream(sensorName: "\(DEVICE_NAME)-PM2.5",
+                                            measurementType: PM_MEASUREMENT_TYPE,
+                                            measurementShortType: PM_MEASUREMENT_SHORT_TYPE,
+                                            unitName: PM_UNIT_NAME,
+                                            unitSymbol: PM_UNIT_SYMBOL,
+                                            thresholdVeryLow: 0,
+                                            thresholdLow: 12,
+                                            thresholdMedium: 35,
+                                            thresholdHigh: 55,
+                                            thresholdVeryHigh: 150),
```
jankrzempek, line 66: No enter here, in above you have 😃

AirCasting/Utils/DataFormatters.swift
```diff
@@ -112,7 +112,8 @@ enum DateFormatters {
         static let fileParserFormatter: DateFormatter = {
             let dateFormatter = DateFormatter()
             dateFormatter.locale = Locale(identifier: "en_US_POSIX")
-            dateFormatter.dateFormat = "MM/dd/yyy'T'HH:mm:ss"
+            // yyy?
```
jankrzempek, line 115: What is the purpose of this?

jankrzempek (commented):
Few comments only, otherwise cooool 🤖

AirCasting/ABConnector/SDSyncController.swift
```diff
@@ -57,31 +61,65 @@ class SDSyncController: ObservableObject {
...
     }
     
+    private func handle(files: [(URL, SDCardSessionType)], deviceID: String, completion: @escaping (Bool) -> Void ) {
+        let mobileFileURL = files.first(where: { $0.1 == SDCardSessionType.mobile })?.0
+        let fixedFileURL = files.first(where: { $0.1 == SDCardSessionType.fixed })?.0
+        
+        ///TODO: There's a possibility that we have two files - one mobile and one fixed
+        ///then completion can be called more than once - after processing mobile is done and after processing fixed
+        /// Also we should handle situation when on processing is successfull and one is not.
+        #warning("TODO is important here!")
```
anna1901, line 88: Completion should be called only after both steps are done. We can't call completion when saving mobile measurements to db is successful, but uploading fixed measurements is not (for example because the internet is not working). If we did, then we would clear sd card even though we didn't send fixed measurements to db, and those measurements would be lost forever.

anna1901 (changes requested):

AirCasting/Utils/DataFormatters.swift
smialko: it's a leftover. There was yyy and I changed it to yyyy but earlier left a comment as a remainder to ask Ania about it :D

AirCasting/ABConnector/BluetoothManager.swift
```diff
@@ -8,7 +8,23 @@
 import Foundation
 import CoreBluetooth
 
-class BluetoothManager: NSObject, ObservableObject {
+protocol BluetoothCommunicator {
```
anna1901, line 11: Hmmm... I think this is already on develop. Could fix your branch so that we would have only your changes in this PR?

AirCasting/ABConnector/SDSyncController.swift
```diff
@@ -57,31 +61,65 @@ class SDSyncController: ObservableObject {
...
+        if let mobileFileURL = mobileFileURL {
+            process(mobileSessionFile: mobileFileURL, deviceID: deviceID, completion: completion)
+        }
+        if let fixedFileURL = fixedFileURL {
+            try? process(fixedSessionFile: fixedFileURL, deviceID: deviceID, completion: completion)
```
anna1901, line 93: What if it throws an error? We should call completion then

anna1901 (changes requested):

AirCasting/ABConnector/SDSyncController.swift
smialko: Done.

anna1901 (changes requested):
Uploading works great! However, when there are no measurements on SD Card, sync never ends. Please add completion

AirCasting/SDSync/Models/CSVMeasurementStream.swift
pawelgil: I agree with @smialko - each use-case should operate on its own data structures even if they are exactly the same as others. We should look at those from a conceptual standpoint - the code is only truly repeated when it is serving the same purpose, not when it has the same structure. 
However I think those default values like "Temperature" and threshold values especially should be placed in some constants or some kind of a template. I can see default threshold changing some day in the future and currently we have those all around the place and it would be difficult to change all at once

AirCasting/SDSync/Parsing/CSVSession.swift
```diff
@@ -0,0 +1,55 @@
...
+import CoreLocation
+
+class CSVSession {
+    
+    private(set) var sessions = Set<SDSession>()
+    private(set) var streamsWithMeasurements: [SDStream: [Measurement]] = [:]
+    
+    private let parser = SDCardMeasurementsParser()
+    
+    init(fileURL: URL, fileLineReader: FileLineReader) throws {
```
pawelgil, line 14: so this class is basically getting a file and parsing it into our structures `SDSession`, `SDStream`, `SDMeasurement`. So maybe it should be called `SDFileParser` and also for DI purposes be split into protocol and implementation. If we do so we can then inject `FileLineReader` into implementation using init like we do with other dependencies there and instead of doing all the paring in `init` we just create a func for this:
```Swift
protocol SDFileParser {
    func parse(file: URL) throws -> (sessions: [SDSession], streams: [SDStream: [Measurement]])
}

class DefaultSDFileParser: SDFileParser {
    ...
    init(fileReader: FileLineReader) { }
    ...
}
```

PR #497, opened 2021-12-09
Fix clearing sd card after sync

anna1901:
https://trello.com/c/S1gsTlsf


jankrzempek (approved):
Seems like a great code 😎

PR #507, opened 2021-12-13
Spacing between map/graph button and parameters

jankrzempek:
FOR ITERATION 1
https://trello.com/c/2wpv0AXi/473-dashboard-increase-space-between-parameter-values-graph-map-buttons

### DIFF
--- a/AirCasting/SessionViews/ABMeasurementsView.swift
+++ b/AirCasting/SessionViews/ABMeasurementsView.swift
@@ -67,7 +67,7 @@ struct _ABMeasurementsView: View {
                             Spacer()
                         }
                     }
-                }
+                }.padding(.bottom, 10)
             } else {
                 if session.isFollowed {
                     SessionLoadingView()
--- a/AirCasting/SessionViews/ABMeasurementsView.swift
+++ b/AirCasting/SessionViews/ABMeasurementsView.swift
@@ -67,7 +67,7 @@ struct _ABMeasurementsView: View {
                             Spacer()
                         }
                     }
-                }.padding(.bottom, 10)
+                }
             } else {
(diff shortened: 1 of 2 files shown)


AirCasting/SessionViews/ABMeasurementsView.swift
```diff
@@ -67,7 +67,7 @@ struct _ABMeasurementsView: View {
                             Spacer()
                         }
                     }
-                }
+                }.padding(.bottom, 10)
```
smialko, line 70: The task is "dashboard > increase space between parameter values & graph/map buttons" so is adding additional padding here not an overkill? I mean this will add space between ABMEasurementsView and all others while as far as I understand we should add it only fr fixed sessions between parameters and buttons?

smialko: So maybe you could add a padding in SessionCardView in this method? What do you think? 
```
func displayButtons(thresholds: [SensorThreshold]) -> some View {
        HStack() {
            if sessionCartViewModel.isFollowing && session.type == .fixed {
                unFollowButton
            } else if session.type == .fixed {
                followButton
            }
            Spacer()
            !session.isIndoor ? mapButton.padding(.trailing, 10) : nil
            graphButton
        }
        .buttonStyle(GrayButtonStyle())
    }

```

smialko: Adding comments to my comments XD But I see that dormant sessions also have parameters + buttons so we should take them into account, too. :D

smialko (commented):
Much more comments than changes in this PR, sorry, just thinking out loud :D

smialko (approved):
Perfect!

PR #521, opened 2021-12-15
Fastfile and gitignore adjusted for appstore api key use

pawelgil:
Note that you'll need to place AuthKey.p8 inside the fastlane directory and update your .env file to include:
APPSTORE_KEY_ID = ...
APPSTORE_ISSUER_ID = ...


anna1901 (approved):
Awesome 😎

PR #523, opened 2021-12-15
Fixed tests

pawelgil:
I had to remove one test because it was so enigmatic I just couldn't figure out what it's doing


smialko (approved):
👍

PR #525, opened 2021-12-16
Notes - views preparation

jankrzempek:
https://trello.com/c/dqYffVCd/483-mobile-active-add-a-note


AirCasting/Utils/Strings.swift
```diff
@@ -570,4 +570,21 @@ struct Strings {
     enum InAppAlerts {
         static let assertError: String = "Unsupported button count! For SwiftUI implementation max of 2 buttons is supported"
     }
+    
+    enum AddNoteView {
+        static let title: String = "Add a note"
+        static let description: String = "Your note will be timestamped and displayed on the AirCasting map"
+        static let placeholder: String = "Note"
+        static let continueButton: String = "Add a note"
+        static let cancelButton: String = "Cancel"
```
smialko, line 579: I'm not sure if that's a good idea because it's breaking a "rule" that we have separate enum for each view. 
But!
Maybe we could add an Enum for all Buttons? We would avoid having `cancel`, `continue` or other typical buttons repeated so many times. 

Just an idea to discuss :D

smialko (approved):
👏

AirCasting/Utils/Strings.swift
jankrzempek: Great idea, what would you say to make it in the new PR?

smialko: Sure! It's not about your code and we can do it as small refactor in the future. 
Anyways, I would ask Ania&Paweł for their opinion before doing it :)

anna1901 (approved):
Very nice!

AirCasting/Notes/AddNote/AddNoteView.swift
```diff
@@ -0,0 +1,73 @@
...
+        } label: {
+            Text(Strings.AddNoteView.continueButton)
+                .bold()
+        }
+        .buttonStyle(BlueButtonStyle())
+    }
+    
+    var cancelButton: some View {
+        Button {
+            addNoteModal.toggle()
```
pawelgil, line 59: I feel like this should too be handled via viewmodel

AirCasting/Notes/EditNote/EditNoteView.swift
```diff
@@ -0,0 +1,83 @@
...
+        } label: {
+            Text(Strings.EditNoteView.deleteButton)
+                .bold()
+        }
+        .buttonStyle(BlueButtonStyle())
+    }
+    
+    var cancelButton: some View {
+        Button {
+            editNoteModal.toggle()
```
pawelgil, line 69: I feel like this too should be handled by the viewmodel

AirCasting/SessionViews/SessionHeaderView.swift
```diff
@@ -198,6 +205,14 @@ private extension SessionHeaderView {
             Label(Strings.SessionHeaderView.deleteButton, systemImage: "xmark.circle")
         }
     }
+    
+    var actionsMenuNoteButton: some View {
+        Button {
+            showAddNoteModal.toggle()
+        } label: {
+            Label("Add a note", systemImage: "square.and.pencil")
```
pawelgil, line 213: `Strings`

AirCasting/Utils/Strings.swift
pawelgil: I think we should extract common strings to a `Common` enum. So we have all universal strings there like:

```Swift
enum Commons {
    static let cancel: String = "Cancel"
    static let ok: String = "Ok"
    // etc.
}
```

pawelgil (changes requested):
Great job, some small stuff requested

AirCasting/Utils/Strings.swift
jankrzempek: Wil do in new PR then, Thanks!

AirCasting/Notes/EditNote/EditNoteView.swift
jankrzempek: As long as in this and above I used Binding it is not easy for me to do. Is it worth doing in the case of Binding? If so I will dig deeper

pawelgil: It can be done using routing closures (so that whoever creates this scene needs to provide view routings for certain scenarios). I will create a PR to this PR so you can review and see how I'd do it.

pawelgil: Done. see #528

jankrzempek: Thank you @pawelgil. Learning it and looks awesome

pawelgil (approved):

PR #527, opened 2021-12-21
Share session via link

anna1901:
https://trello.com/c/cfc2Co6h


AirCasting/SessionViews/ShareSession/ShareSessionStreamOptionViewModel.swift
```diff
@@ -0,0 +1,20 @@
...
+    let isEnabled: Bool
+
+    mutating func toggleSelection() {
+        isSelected.toggle()
+    }
+
+    mutating func changeSelection(newSelected: Bool) {
+        isSelected = newSelected
+    }
+}
```
jankrzempek, line 20: I have no hard opinion here, and it would be best to ask @pawelgil. Assuming I know smth, I would like to know why it is better than the class? We are used to viewModels being class I think. Can this one be class too? I think it is a bit rule-breaking, as long as the struct is immutable we should leave it like this, but as I mentioned above I am not sure. I am going to read about it and increase my knowledge level

AirCasting/SessionViews/ShareSession/ShareSessionView.swift
```diff
@@ -6,13 +6,12 @@ import MessageUI
...
+    
+    // This email logic will be moved to view model in the next task
     @State var email: String = ""
-    @State var itemsForSharing: [String] = ["www.google.com"]
-    #warning("Implement working share sheet")
-    //To be sure that share sheet is working for now we are using "www.google.com" and then we will place some data there when making a logic to this
-    @State var showSheet = false
     @State var isShowingMailView = false
     @State var showingAlert = false
     @State var mailSendingResult: Result<MFMailComposeResult, Error>? = nil
```
jankrzempek, line 17: I know it was done by me, but I would like to ask you: Could you please move it to VM? If not I will do it in another task, let me know

AirCasting/SessionViews/ShareSession/ShareSessionView.swift
```diff
@@ -23,18 +22,24 @@ struct ShareView: View {
...
+                //                oKButton
                 cancelButton
             }
-        }.sheet(isPresented: $showSheet, content: {
-            ActivityViewController(itemsToShare: itemsForSharing)
-        }).padding()
+        }
+        .alert(item: $viewModel.alert, content: { $0.makeAlert() })
+        .sheet(isPresented: $viewModel.showSheet, content: {
+            ActivityViewController(itemsToShare: [viewModel.sharingLink as Any]) {activityType,completed,returnedItems,error in
```
jankrzempek, line 37: here please` { activityType,completed,returnedItems, error in` spaces

AirCasting/SessionViews/ShareSession/ShareSessionViewModel.swift
```diff
@@ -0,0 +1,113 @@
+// Created by Lunar on 16/12/2021.
+//
+
+import Foundation
+
+protocol ShareSessionViewModel: ObservableObject {
+    var streamOptions: [ShareSessionStreamOptionViewModel] {get set}
```
jankrzempek, line 7: `{ get set }` spaces please

AirCasting/SessionViews/ShareSession/ShareSessionViewModel.swift
```diff
@@ -0,0 +1,113 @@
...
+    var alert: AlertInfo? { get set }
+    var showSheet: Bool { get set }
+    var sharingLink: URL? { get set }
+    func didSelect(option: ShareSessionStreamOptionViewModel)
+    func shareLinkButtonGotPressed()
+}
+
+enum ShareSessionError: Error {
+    case noSessionURL
+}
```
jankrzempek, line 17: enum I think should be above protocol but not sure really @pawelgil could you please help here

AirCasting/SessionViews/ShareSession/ShareSessionViewModel.swift
```diff
@@ -0,0 +1,113 @@
...
+    func didSelect(option: ShareSessionStreamOptionViewModel) {
+        guard let index = streamOptions.firstIndex(where: { $0.id == option.id }) else {
+            assertionFailure("Unknown option index")
+            return
+        }
+        
+        if !streamOptions[index].isSelected {
+            for i in streamOptions.indices {
+                streamOptions[i].changeSelection(newSelected: false)
+            }
```
jankrzempek, line 52: Can we please rename i ? To smth like indice maybe

jankrzempek (commented):
Few comments left, few to discuss I think. I've tested and need to say that sheet is popping up all the time as expected - great thing 🚀

AirCasting/SessionViews/ShareSession/ShareSessionViewModel.swift
pawelgil: Also matter of taste. I usually put all types used by the protocol above it, yeah. As you cannot scope errors and structs inside protocol definition I use this kind-of rule of thumb: "If an enum/struct is something you'd put inside a protocol if it was possible, put it above it". IDK if this is clear - the reason is that you want a nice top to bottom readability in a file, so when you reach to the protocol you already know what errors mean and what return structs mean. This is also how it's done in majority of the projects afaik

AirCasting/SessionViews/ShareSession/ShareSessionView.swift
anna1901: I left it for the task with sharing file via email and I intend to do it (as mentioned in the comment). Is that okay?

AirCasting/CoreData/MeasurementStreamStorage.swift
```diff
@@ -103,12 +103,8 @@ final class HiddenCoreDataMeasurementStreamStorage: MeasurementStreamStorageCont
...
-            do {
-                toDelete.forEach { object in
-                    context.delete(object)
-                }
-            } catch {
-                Log.error("Error when deleting session")
+            toDelete.forEach { object in
+                context.delete(object)
             }
         }
```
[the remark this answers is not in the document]
anna1901: I changed it only to get rid of the warning that catch would never be reached because the call doesn't throw. I think it's from your ticket for deleting sessions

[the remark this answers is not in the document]
anna1901: But you are right, it does seem weird. You don't remember why was it done this way?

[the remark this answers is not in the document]
jankrzempek: Need to say that have no idea, but now I remember that it was for something so maybe we can leave it for now?

AirCasting/SessionViews/ShareSession/ShareSessionStreamOptionViewModel.swift
anna1901: Ah, you are right. It's just by mistake, I also think it should be a class. Thanks for catching that :)

AirCasting/SessionViews/ShareSession/ShareSessionView.swift
jankrzempek: Sure! Thank you a lot 🤖

AirCasting/CoreData/MeasurementStreamStorage.swift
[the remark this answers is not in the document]
anna1901: Sure, I'll just add a warning 😉 😄

AirCasting/SessionViews/ShareSession/ShareSessionStreamOptionViewModel.swift
anna1901: Hm, now I see that I took your `struct DeleteSessionOptionViewModel` and adapted it for sharing purposes. So I'm not sure if it should be a struct or a class. In your opinion, should I change `DeleteSessionOptionViewModel` to class as well?

pawelgil (approved):

PR #530, opened 2021-12-22
Strings struct - refactor 

jankrzempek:
https://trello.com/c/qobMnNDl/492-refactor-strings

### DIFF
--- a/AirCasting/Utils/Strings.swift
+++ b/AirCasting/Utils/Strings.swift
@@ -4,6 +4,14 @@
 import Foundation
 
 struct Strings {
+    enum Commons {
+        static let cancel: String = "Cancel"
+        static let ok: String = "OK"
+        static let myAccount: String = "My account"
+        static let `continue`: String = "Continue"
+        static let gotIt: String = "Got it!"
+    }
+    
     enum Settings {
         static let title: String = "Settings"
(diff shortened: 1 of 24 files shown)


anna1901 (approved):
Awesome job! I just noticed you left "Got it!" button in selectPeripheralView and ShareView, continueButton in OnboardingAirBeam and "OK" button in SignInView. I'm not sure if "My account" should be in commons, but it's probably a matter of taste.

AirCasting/Utils/Strings.swift
```diff
@@ -4,6 +4,14 @@
 import Foundation
 
 struct Strings {
+    enum Commons {
+        static let cancel: String = "Cancel"
+        static let ok: String = "OK"
+        static let myAccount: String = "My account"
+        static let `continue`: String = "Continue"
```
anna1901, line 11: why are there those `` around continue?

jankrzempek: Thank you for mentioning what I have left. This `` is used to make continue a let as the others because normally continue is a reserved keyword in Swift (Special thanks to @pawelgil for that knowledge!)

pawelgil (approved):
Nice refactor!

PR #531, opened 2021-12-22
Session name is blank

jankrzempek:
https://trello.com/c/rUzKLjUL/493-do-not-allow-empty-name

### DIFF
--- a/AirCasting/CreateSessionViews/CreateSessionDetails/CreateSessionDetailsViewModel.swift
+++ b/AirCasting/CreateSessionViews/CreateSessionDetails/CreateSessionDetailsViewModel.swift
@@ -15,7 +15,8 @@ class CreateSessionDetailsViewModel: ObservableObject {
     @Published var isLocationSessionDetailsActive: Bool = false
     @Published var showAlertAboutEmptyCredentials = false
     @Published var isSSIDTextfieldDisplayed: Bool = false
-    @Published var shouldShowError = false
+    var shouldShowError: Bool { sessionName.isEmpty && !yetNoInteraction }
+    private var yetNoInteraction = true
     
     let baseURL: BaseURLProvider
     // It cannot be private as long as it is going to be passed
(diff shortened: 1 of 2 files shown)


anna1901 (approved):
Great :)

AirCasting/CreateSessionViews/CreateSessionDetails/CreateSessionDetailsViewModel.swift
```diff
@@ -24,6 +25,10 @@ class CreateSessionDetailsViewModel: ObservableObject {
         self.baseURL = baseURL
     }
     
+    func onSessionNameChange() {
+        (sessionName != "") ? (shouldShowError = false) : (shouldShowError = true)
```
pawelgil, line 29: a bit overcomplicated, could've been:

```Swift
shouldShowError = (sessionName == "")
```

But there is no need for this method `onSessionNameChange` at all here. It can be done purely inside viewmodel, just change `shouldShowError` to a computed property instead:
``` Swift
class CreateSessionDetailsViewModel: ObservableObject {
    @Published var sessionName = ""
    var shouldShowError { sessionName.isEmpty }
}
```

Since `shouldShowError` is only read whenever `sessionName` changes we don't need to worry about `objectWillChange.send()` because it will get called each time `sessionName` changes thus forcing any `shouldShowError` reads too.

AirCasting/CreateSessionViews/CreateSessionDetails/CreateSessionDetailsViewModel.swift
```diff
@@ -32,6 +37,10 @@ class CreateSessionDetailsViewModel: ObservableObject {
     func onContinueClick(sessionContext: CreateSessionContext) -> CreateSessionContext {
         // sessionContext is needed becouse it is being modified in the session creation proccess
         // by 'modified' I mean - the data it ovverriden by the proper one (get from user) on every step
+        guard !sessionName.isEmpty else {
+            shouldShowError = true
+            return sessionContext
+        }
```
pawelgil, line 43: after the changes I think this should be changed:

```Swift
guard !sessionName.isEmpty else { return sessionContext }
```

pawelgil (changes requested):
I think we can make this a bit cleaner

AirCasting/CreateSessionViews/CreateSessionDetails/CreateSessionDetailsViewModel.swift
jankrzempek: Thank you, looks so much better!

pawelgil (approved):

PR #537, opened 2022-01-06
unpluggedView - image changed

jankrzempek:
https://trello.com/c/kRz9R8jM/464-sync-wizard-add-image-to-unplug-your-airbeam-screen


anna1901 (approved):
Awesome! 🚀

AirCasting/SDSync/ViewsFlow/SDSyncUnplugAB/UnplugABView.swift
```diff
@@ -33,7 +33,7 @@ struct UnplugABView<VM: UnplugABViewModel>: View {
 extension UnplugABView {
     
     var unplugImage: some View {
-        Image("airbeam")
+        Image("airbeam-unplugged")
```
anna1901, line 36: It looks like this image is a bit smaller than the one on the screen before and on the screen after. If that could be changed, that would be nice, but if not, I think it's still okay 😊

PR #538, opened 2022-01-07
Get settings for crowd map from user defaults at the start of the app

anna1901:
This just adds a small fix to UserSettings. Previously we were always setting contributingToCrowdMap to true at the app launch and now we will take them from userdefautls

### DIFF
--- a/AirCasting/AuthViews/UserSettings.swift
+++ b/AirCasting/AuthViews/UserSettings.swift
@@ -6,7 +6,7 @@ import UIKit
 
 class UserSettings: ObservableObject {
     private let userDefaults: UserDefaults
-    let crowdMapKey = Constants.UserDefaultsKeys.crowdMap
+    private let crowdMapKey = Constants.UserDefaultsKeys.crowdMap
     
     var contributingToCrowdMap: Bool {
         get {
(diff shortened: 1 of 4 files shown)


AirCasting/Extensions/UserDefaults.swift
```diff
@@ -0,0 +1,12 @@
+// Created by Lunar on 06/01/2022.
+//
+
+import Foundation
+
+extension UserDefaults {
+
+    func valueExists(forKey key: String) -> Bool {
```
pawelgil, line 8: neat 💯

AirCasting/AuthViews/UserSettings.swift
```diff
@@ -6,13 +6,14 @@ import UIKit
 
 class UserSettings: ObservableObject {
     private let userDefaults: UserDefaults
+    let crowdMapKey = Constants.UserDefaultsKeys.crowdMap
```
pawelgil, line 9: private

pawelgil (approved):

AirCasting/AuthViews/UserSettings.swift
anna1901: ah, when will I learn 🙃

PR #542, opened 2022-01-10
Session name empty - error

jankrzempek:
Show error only when (from Michael):
. show this message only if: the session name field is blank AND the user taps "Continue".


anna1901 (approved):
👌🏻

PR #544, opened 2022-01-10
Adds Resolver and moves some types to service locator

pawelgil:
Partial `Service Locator / Dependency Injecion` implementation for the AC app.

Using lightweight [Resolver](https://github.com/hmlongco/Resolver) library


pawelgil:
This branch is not supposed to be merged in just yet, converting back to draft

PR #546, opened 2022-01-12
Session creator aligement

jankrzempek:
https://trello.com/c/MQGkGo5P/498-session-creator-ui-alligment


AirCasting/CreateSessionViews/ConnectingABView.swift
```diff
@@ -19,22 +20,25 @@ struct ConnectingABView<VM: AirbeamConnectionViewModel>: View {
     
     
     var body: some View {
-        VStack(spacing: 50) {
-            ProgressView(value: 0.5)
+        VStack() {
+            ProgressView(value: 0.25)
```
anna1901, line 24: This value is actually decreasing progress bar. 🤔 Wasn't it good with 0.5?

anna1901 (changes requested):
Looks, very good. 😊 Please just change that progress bar value. Other than that, everything looks better. 😊

AirCasting/CreateSessionViews/ConnectingABView.swift
jankrzempek: Yeah, it was. Thank you! my bad :/

anna1901 (approved):

PR #547, opened 2022-01-12
Add notes on map and graph

jankrzempek:
https://trello.com/c/28joI4Ik/487-mobile-active-add-a-note-display-notes-on-graph-map


AirCasting/CreateSessionViews/ConfirmCreatingSessionView.swift
```diff
@@ -90,9 +90,9 @@ struct ConfirmCreatingSessionView: View {
                 .lineSpacing(9.0)
                 ZStack {
                     if sessionContext.sessionType == .mobile {
-                        GoogleMapView(pathPoints: [], isMyLocationEnabled: true, placePickerDismissed: Binding.constant(false), isUserInteracting: Binding.constant(true))
+                        GoogleMapView(pathPoints: [], isMyLocationEnabled: true, placePickerDismissed: Binding.constant(false), isUserInteracting: Binding.constant(true), mapNotes: .constant([]))
```
pawelgil, line 93: we need to make a task for resolving this issue. I think splitting GoogleMapView into:
* `GoogleMapView` -> Plain, simple mapview with some customization points (used here and in other places that don't need measurements, notes etc. plotted)
* `AirCastingMapView` -> The map with features added used for displaying sessions

AirCasting/Map/MapNotesViewModel.swift
```diff
@@ -0,0 +1,30 @@
...
+import CoreLocation
+import UIKit
+
+struct MapNote {
+    let id: Int
+    let location: CLLocationCoordinate2D
+    let markerImage: UIImage
+}
+
+class MapNotesViewModel: ObservableObject {
```
pawelgil, line 14: tapping a note should also be handlede here, but for that we'd need some form of a router class/block here so maybe let's skip it for now since those mapviews are really a mess and need refactoring

AirCasting/Notes/EmptyNotesHandler.swift
```diff
@@ -0,0 +1,34 @@
+// Created by Lunar on 05/01/2022.
+//
+
+import Foundation
+
+class EmptyNotesHandler: NotesHandler {
```
pawelgil, line 6: when we refactor AirMapView/GoogleMapView this can be removed

AirCasting/Notes/NotesHandler.swift
```diff
@@ -0,0 +1,80 @@
...
+    
+    func addNote(noteText: String) {
+        measurementStreamStorage.accessStorage { [self] storage in
+            do {
+                let currentNumber = try storage.getNotes(for: sessionUUID).map(\.number).sorted(by: < ).last
+                try storage.addNote(Note(date: Date(),
+                                         text: noteText,
+                                         lat: locationTracker.googleLocation.last?.location.latitude ?? 20.0,
+                                         long: locationTracker.googleLocation.last?.location.longitude ?? 20.0,
+                                         number: (currentNumber ?? -1) + 1),
```
pawelgil, line 33: gonna run into an Int overflow problem when user adds more than 9,223,372,036,854,775,807 notes 🤓🤓🤓

pawelgil (approved):
Nice one. Very cool that you came out with the correct design 100% by yourself first try!

AirCasting/Map/MapNotesViewModel.swift
jankrzempek: Should I make a task for a future: "make a router" or sth? Or we can just skip at all?

AirCasting/Notes/NotesHandler.swift
jankrzempek: 🤫😬 Sounds not fun at all, but I believe in the users a bit in this scenario

PR #551, opened 2022-01-14
Convert to celcius

anna1901:
* Add settings for converting temperature to celsius
* Implement conversion to celsius on UI level

https://trello.com/c/y2GtuRuw

### DIFF
--- a/AirCasting/SessionViews/SingleMeasurementView.swift
+++ b/AirCasting/SessionViews/SingleMeasurementView.swift
@@ -9,10 +9,17 @@ struct SingleMeasurementView: View {
     var threshold: SensorThreshold?
     @Binding var selectedStream: MeasurementStreamEntity?
     @Binding var isCollapsed: Bool
+    @EnvironmentObject var userSettings: UserSettings
     let measurementPresentationStyle: MeasurementPresentationStyle
     let isDormant: Bool
     var value: Double {
-        isDormant ? stream.averageValue : (stream.latestValue ?? 0)
+        let measurementValue = isDormant ? stream.averageValue : (stream.latestValue ?? 0)
+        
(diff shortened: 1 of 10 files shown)


anna1901:
[settings: celsius](https://trello.com/c/y2GtuRuw/496-settings-celsius)

AirCasting/SessionViews/SingleMeasurementView.swift
```diff
@@ -50,6 +57,8 @@ struct SingleMeasurementView: View {
         guard let streamName = stream.sensorName else { return "" }
         if streamName == Constants.SensorName.microphone {
             return "dB"
+        } else if stream.isTemperature {
+            return userSettings.convertToCelsius ? "C" : "F"
```
pawelgil, line 61: those units can be moved to `Strings`

pawelgil (approved):
This code is not wrong at all

but... 🤓

There is an iOS-SDK provided method for storing and converting measurements. We should seriously think about implementing it, especially because it just fits AirCasting perfectly. Please take a look here:
1) [General documentation for measurements](https://developer.apple.com/documentation/foundation/units_and_measurement)
2) [Temperature specific units](https://developer.apple.com/documentation/foundation/unittemperature)
3) [Formatting](https://developer.apple.com/documentation/foundation/measurementformatter)

So in our domain structures we'd use `Measurement` items, than convert those to strings in viewmodels or to JSON values when coming to/from the server. This secures correct units - Swift type system would ensure we're not ever confusing celcius values with fahrenheit or PM10 with PM2.5. Even NASA lost 200M$ spacecraft due to such error (https://www.simscale.com/blog/2017/12/nasa-mars-climate-orbiter-metric/) :)

jankrzempek (approved):

PR #552, opened 2022-01-17
[MAIN] Note upload and download - server side.

jankrzempek:
Download:
Fetching data from the server and encoding it trough 3 layers. After that, saving notes to CoreDatawhich are connected to the specified session.
Upload:
Fetching data from Core Data and coding it trough 3 layers. After that, sending it to server inside the session struct.

### DIFF
--- a/AirCasting/Graph/AirCastingGraph.swift
+++ b/AirCasting/Graph/AirCastingGraph.swift
@@ -179,7 +179,9 @@ extension AirCastingGraph {
         noteButtons.forEach { $0.button.removeFromSuperview() }
         noteButtons = notes.map { (createNewNoteButton(), $0, onTap) }
         noteButtons.forEach { lineChartView.addSubview($0.button) }
-        setNeedsLayout()
+        DispatchQueue.main.async { [self] in
+            setNeedsLayout()
+        }
     }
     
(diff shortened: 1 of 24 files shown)


AirCasting/Graph/AirCastingGraph.swift
```diff
@@ -179,7 +179,9 @@ extension AirCastingGraph {
         noteButtons.forEach { $0.button.removeFromSuperview() }
         noteButtons = notes.map { (createNewNoteButton(), $0, onTap) }
         noteButtons.forEach { lineChartView.addSubview($0.button) }
-        setNeedsLayout()
+        DispatchQueue.main.async { [self] in
```
pawelgil, line 182: if this `setupNotes` func is not called on main thread other lines here are also manipulating the UI, so it should be wrapped in the dispatch block too. But I don't think that's a good idea, the caller should be responsible for making sure it's main thread. Let's add an assert at the top and adjust the calling code:
```Swift
assert(Thread.isMainThread) // don't remember exact naming here, but something like that.
```

pawelgil (changes requested):
Looks pretty good, let's change this one dispatch tho.

pawelgil (approved):

PR #556, opened 2022-01-19
Delete noted - server side.

jankrzempek:
https://trello.com/c/NkiU9Wuu/485-mobile-active-add-a-note-delete-note

In the NotesHandler, deleting function was modified to send request to API to update session.

### DIFF
--- a/AirCasting/Notes/EditNote/EditNoteViewModel.swift
+++ b/AirCasting/Notes/EditNote/EditNoteViewModel.swift
@@ -33,7 +33,9 @@ class EditNoteViewModelDefault: EditNoteViewModel, ObservableObject {
     }
     
     func deleteTapped() {
-        notesHandler.deleteNote(note: note); exitRoute()
+        notesHandler.deleteNote(note: note, completion: {
+            self.exitRoute()
+        })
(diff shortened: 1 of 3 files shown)


AirCasting/Notes/EditNote/EditNoteViewModel.swift
```diff
@@ -33,7 +33,9 @@ class EditNoteViewModelDefault: EditNoteViewModel, ObservableObject {
     }
     
     func deleteTapped() {
-        notesHandler.deleteNote(note: note); exitRoute()
+        notesHandler.deleteNote(note: note, completion: {
+            self.exitRoute()
```
pawelgil, line 37: `[weak self]` would be better since the user might exit the screen before delete operation is finished

pawelgil (changes requested):

AirCasting/Notes/EditNote/EditNoteViewModel.swift
jankrzempek: sure! Changing

pawelgil (approved):

PR #557, opened 2022-01-19
TextView - place for note content

jankrzempek:
Added custom TextView for note.

Also fixed editing note when there are no tags or the session is active during editing.

https://trello.com/c/3l0uspen/509-note-text-field-improved-edit-note-fix

### DIFF
--- a/AirCasting/Notes/NotesTextView.swift
+++ b/AirCasting/Notes/NotesTextView.swift
@@ -2,64 +2,8 @@
 
 import SwiftUI
 
-struct NoteTextView: UIViewRepresentable {
-    
-    @Binding var text: String
-    var placeholder: String
-    var font = UIFont.preferredFont(forTextStyle: .body)
-    
-    func makeUIView(context: Context) -> UITextView {
-        let textView = UITextView()
-        textView.delegate = context.coordinator
-        
-        textView.font = font
-        textView.autocapitalizationType = .sentences
-        textView.isSelectable = true
-        textView.isUserInteractionEnabled = true
-        textView.backgroundColor = UIColor.aircastingGray.withAlphaComponent(0.05)
-        textView.layer.borderColor = UIColor.aircastingGray.withAlphaComponent(0.1).cgColor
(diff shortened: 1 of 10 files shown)


pawelgil (changes requested):
Let's not use 3rd party libraries in this case. It's a lot of risk and we can very easily achieve the goal by writing our own `UIViewRepresentable`. 90% of the time pulling 3rd party utility libraries is a bad idea. Some exceptions are:
* Really small libraries that we code reviewed and are 100% sure we could fork and maintain it if abandoned (like `Resolver` that I pulled recently. It's a small, 700-ish line library that I fully understand and would be able to maintain myself)
* Implementations that are not going to ever change (like crypto/maths stuff, for example `SwiftSimplify` or `ZipArchive` we're using)

UI Libraries in particular are considered the most risky and we shouldn't use them unless we have a *very* good reason (I think we shouldn't use `Charts` library, but since we're using it on Android too and it _kind of_ makes sense for baselining reasons I don't really push for removing it. But we are taking on a huge risk with it to be honest).

AirCasting/Notes/NotesTextView.swift
```diff
@@ -0,0 +1,68 @@
+// Created by Lunar on 20/01/2022.
+
+import SwiftUI
+
+struct NoteTextView: UIViewRepresentable {
```
pawelgil, line 5: This can be named more generically like `TextView`

AirCasting/Notes/NotesTextView.swift
```diff
@@ -0,0 +1,68 @@
...
+            }
+        }
+    }
+    
+    func makeCoordinator() -> Coordinator {
+        Coordinator(self)
+    }
+}
+
+func createNoteTextField(binding: Binding<String>) -> some View {
```
pawelgil, line 61: This should be moved out to some other place really if w make this view generic, notes-unrelated

pawelgil (approved):
Good job!!

PR #558, opened 2022-01-19
Core Data - model migration

jankrzempek:
NoteEntity was added to the Core Data model. It consists of five attributes and has a session relation.


pawelgil (approved):
Awesome!

PR #562, opened 2022-01-24
Notes - manage times zones.

jankrzempek:
For now, every not was timestamped with the current Date which could result in a time difference between graph and note. Now, every note is timestamped with Date.currentUTCTime. It matches our time zone convention now.


pawelgil (approved):

pawelgil:
This date thing would really use some more thought - how can we make it so that we don't forget about this? It looks like we'll be having exact same bug to fix each time we do anything with dates...

PR #567, opened 2022-01-25
Display celsius on chart and in statistics container

anna1901:


pawelgil (commented):
Such bool-flag approach is a bit troublesome, it causes a tech debit and should be avoided:
* It modifies existing core code (namely the controller, viewmodel etc. that are basically doing their jobs of calculating stats just fine as they are now).
* Adds complexity - the more bool permutations we add the messier the code gets. We can already see how it turns out in some of our views with massive if/else if/else if/... blocks.
* Closes viewmodel and controller for easy extension - before the change those were happily classes unaware of measurement types and all of our app logic. This module basically did its job of computing low/high/avg and providing info about how to display this data. It could without a change be used in another, completely different project (for example a racing game where we'd show lap times in it - high/low/avg? - Note that this is not the goal of open/closed, we won't ever use this in another project for real, but this is a good mental practice to think about modules this way because it validates the architecture). Now its tied to AirCasting and temperature. This is important because HabitatMap is really an open platform and it's really possible that some day we'll be asked to support custom streams with unknown types. This is already possible on the platform level.

I'll put up a PR to this branch with a polymorphic approach to this problem that will not touch current code at all and will still get the job done.

AirCasting/Graph/StatisticsContainer/Adapters/ConveringStatisticsDataSourceDecorator.swift
```diff
@@ -0,0 +1,38 @@
...
+class ConveringStatisticsDataSourceDecorator<D: MeasurementsStatisticsDataSource>: MeasurementsStatisticsDataSource, ObservableObject {
+    let dataSource: D
+    var stream: MeasurementStreamEntity?
+    private let settings: UserSettings
+    private var cancellables: [AnyCancellable] = []
+    
+    init(dataSource: D, stream: MeasurementStreamEntity?, settings: UserSettings) {
+        self.dataSource = dataSource
+        self.stream = stream
+        self.settings = settings
```
pawelgil, line 15: plz add `setupHooks()` here

AirCasting/SessionViews/ChartViewModel.swift
```diff
@@ -28,17 +28,19 @@ final class ChartViewModel: ObservableObject {
...
     private var backgroundNotificationHandle: Any?
+    private var useCelsius: Bool
     
     deinit {
         mainTimer?.invalidate()
         firstTimer?.invalidate()
     }
     
-    init(session: SessionEntity, persistence: PersistenceController) {
+    init(session: SessionEntity, persistence: PersistenceController, useCelsius: Bool) {
```
pawelgil, line 38: not a fan of passing a flag here too, but this would require a solid rewrite so maybe let's pass.

But... I think we'll have to start doing such refactorings along the way soon in order to clean the code up a bit. Let's wait for injection mechanism to be merged in first tho.

pawelgil (changes requested):

pawelgil (approved):

jankrzempek (approved):

AirCasting/SessionViews/SingleMeasurementViewModel.swift
```diff
@@ -0,0 +1,21 @@
+// Created by Lunar on 26/01/2022.
+//
+
+import Foundation
+import Combine
+
+class SingleMeasurementViewModel: ObservableObject {
+    let settings: UserSettings
```
pawelgil, line 8: well, this is not a correct approach to create viewmodel and still expose the original dependency thru it. I see 2 options here:
1) make this `UserSettings` dependency property private and expose some proper primitive value. Best would be to move the whole value here:
```Swift
@Published var measurementValue: Double
```
this way we can kill two birds with one stone - move some non-view logic from the `SingleMeasurementView` and hide this dependency.

2) Don't introduce this viewmodel just yet and instead properly implement `ObjervableObject` for `UserSettings` (currently this conformance is there only to be able to inject it via `@EnvironmentObject`, we can make it properly observable by calling `objectWillChange.send()` in each of the setters). This method would be easier to implement for sure, but will introduce some more tech debit instead of cleaning things up.

your call :)

pawelgil (changes requested):

AirCasting/SessionViews/SingleMeasurementViewModel.swift
anna1901: Okay, I'm choosing the second one, thanks 👌🏻

AirCasting/SessionViews/SingleMeasurementView.swift
```diff
@@ -5,7 +5,7 @@ import SwiftUI
 import AirCastingStyling
 
 struct SingleMeasurementView: View {
-    @ObservedObject var viewModel: SingleMeasurementViewModel
+//    @ObservedObject var viewModel: SingleMeasurementViewModel
```
pawelgil, line 8: this should be removed

AirCasting/AuthViews/UserSettings.swift
```diff
@@ -50,6 +50,7 @@ class UserSettings: ObservableObject {
         set {
             userDefaults.setValue(newValue, forKey: convertToCelsiusKey)
             Log.info("Changed convert to celcius setting to \(convertToCelsius ? "ON" : "OFF")")
+            objectWillChange.send()
```
pawelgil, line 53: maybe let's add this to each setter, not only the celcius one ;)

pawelgil (approved):

PR #568, opened 2022-01-25
Polymorphic approach to F<->C conversion in stats container

pawelgil:
# Note!
This is a draft code only to show how it can be done using polymorphism instead of conditionals. It needs:
* Fixing commented out preview part
* Testing if it works ;)

### DIFF
--- a/AirCasting/Graph/StatisticsContainer/Adapters/ConveringStatisticsDataSourceDecorator.swift
+++ b/AirCasting/Graph/StatisticsContainer/Adapters/ConveringStatisticsDataSourceDecorator.swift
@@ -1,0 +1,38 @@
+// Created by Lunar on 25/01/2022.
+//
+
+import Combine
+
+class ConveringStatisticsDataSourceDecorator<D: MeasurementsStatisticsDataSource>: MeasurementsStatisticsDataSource, ObservableObject {
+    let dataSource: D
+    var stream: MeasurementStreamEntity?
+    private let settings: UserSettings
+    private var cancellables: [AnyCancellable] = []
+    
+    init(dataSource: D, stream: MeasurementStreamEntity?, settings: UserSettings) {
+        self.dataSource = dataSource
+        self.stream = stream
+        self.settings = settings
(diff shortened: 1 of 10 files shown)


AirCasting/Graph/StatisticsContainer/Adapters/ConveringStatisticsDataSourceDecorator.swift
```diff
@@ -0,0 +1,38 @@
...
+class ConveringStatisticsDataSourceDecorator<D: MeasurementsStatisticsDataSource>: MeasurementsStatisticsDataSource, ObservableObject {
+    let dataSource: D
+    var stream: MeasurementStreamEntity?
+    private let settings: UserSettings
+    private var cancellables: [AnyCancellable] = []
+    
+    init(dataSource: D, stream: MeasurementStreamEntity?, settings: UserSettings) {
+        self.dataSource = dataSource
+        self.stream = stream
+        self.settings = settings
```
pawelgil, line 15: call `setupHooks()` here (forgot to)

AirCasting/SessionViews/SessionCardView.swift
```diff
@@ -101,7 +101,9 @@ struct SessionCardView: View {
         }
         .onChange(of: selectedStream, perform: { [weak graphStatsDataSource, weak mapStatsDataSource, weak chartViewModel] newStream in
             graphStatsDataSource?.stream = newStream
+            graphStatsDataSource?.dataSource.stream = newStream
```
anna1901, line 104: Could this be done on set of stream in converter?

pawelgil: not really. The `stream` is a property of both the `ConveringStatisticsDataSourceDecorator` and `Graph/MapStatsDataSource` concrete classes, but not the `MeasurementsStatisticsDataSource` itself. This is also why I exposed the original data source: `let dataSource: D` - in case we need access to the specific stuff that concrete datasources expose.

PR #572, opened 2022-01-26
Rewriting settings view and warning manage.

jankrzempek:
Whole logic was moved to SettingsViewModel now. Few warnings handled too.

### DIFF
--- a/AirCasting/Map/PathPoint.swift
+++ b/AirCasting/Map/PathPoint.swift
@@ -19,7 +19,9 @@ struct PathPoint: Equatable {
     let location: CLLocationCoordinate2D
     let measurementTime: Date
     var measurement: Double = 20
-    #warning("Do something with hard coded measurement")
+    #warning("Do something with hard coded measurement: https://github.com/HabitatMap/AirCastingiOS/issues/586")
+    // FIXME: PathPoints should not have hardcoded measurement, as it has now. It should only get any value, whenever the real measurement
+    // FIXME: appears in our app. Otherwise, it should not be considered getting any value, as it is now ( = 20).
 }
 
 extension PathPoint {
--- a/AirCasting/Map/PathPoint.swift
+++ b/AirCasting/Map/PathPoint.swift
@@ -18,9 +18,9 @@ struct PathPoint: Equatable {
     
     let location: CLLocationCoordinate2D
     let measurementTime: Date
-    let measurement: Double
+    var measurement: Double = 20
 }
 
 extension PathPoint {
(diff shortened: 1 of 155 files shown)


AirCasting/Settings/SettingsView.swift
```diff
@@ -254,3 +225,13 @@ struct SettingsView: View {
     }
     #endif
 }
+
+extension SettingsView {
+    func settingSwitch(toogle using: Binding<Bool>, label with: String) -> some View {
```
pawelgil, line 230: well done

AirCasting/Map/PathPoint.swift
```diff
@@ -18,9 +18,9 @@ struct PathPoint: Equatable {
     
     let location: CLLocationCoordinate2D
     let measurementTime: Date
-    let measurement: Double
+    var measurement: Double = 20
```
pawelgil, line 21: that's not resolving the issue - it changes its location from being in multiple places to being in a single place which **great**, but this is still a code smell that indicates we have a wrong abstraction here perhaps. Why is `googleLocation` using PathPoint? It is is basically a measurement with added location, maybe we need two objects instead of one?

Please keep the warning and add an issue to github accordingly

pawelgil (changes requested):
Great job, let's keep one warning and add an issue to GH for it and it will be ready to go

AirCasting/Map/PathPoint.swift
jankrzempek: Yeah, right. I am going to add now

AirCasting/Map/PathPoint.swift
```diff
@@ -19,6 +19,7 @@ struct PathPoint: Equatable {
     let location: CLLocationCoordinate2D
     let measurementTime: Date
     var measurement: Double = 20
+    #warning("Do something with hard coded measurement")
```
pawelgil, line 22: plase adhere to our new warning standards (see readme) :)

pawelgil (changes requested):

AirCasting/Map/PathPoint.swift
jankrzempek: I need to say that I cannot see it on readMe 😬, should I add FIXME or smth else?

pawelgil: > I need to say that I cannot see it on readMe 😬, should I add FIXME or smth else?

My bad 😅
README is updated now

jankrzempek: Thanks 🚀

pawelgil (approved):

PR #573, opened 2022-01-26
Automatic string detection using XCode 13.

jankrzempek:
Current Strings system and allow XCode to detect strings and then we are able to export then easly.


AirCasting/Utils/Strings.swift
```diff
@@ -310,285 +310,298 @@ struct Strings {
...
+        static let saveButton: String = LocalizedStringKey("Save changes").stringKey
+        static let deleteButton: String = LocalizedStringKey("Delete note").stringKey
+        static let cancelButton: String = LocalizedStringKey("Cancel").stringKey
     }
 }
+
+// Extension allows us to preserve current Strings system and allow XCode to detect strings and then we are able to export then easly.
+extension LocalizedStringKey {
+    var stringKey: String {
+    let description = "\(self)"
```
pawelgil, line 599: unindented code inside `var` body

pawelgil (approved):
Not sure if I like the new `LocalizedStringKey` stuff and the fact that it needs importing SwiftUI, but let's roll with it and see if it causes any problems. Should be easy to change if yes

PR #577, opened 2022-01-27
Location screen for fixed session moved.

jankrzempek:
The location screen which prompts the user to give the app access to the Location was moved to 'after SessionDetailsView' in the case when session is fixed. On the mobile it stays as it is.

https://trello.com/c/M3fRJ3wn/500-allow-users-to-create-fixed-session-even-if-they-have-location-tracking-disabled-but-keep-the-turn-on-location-view-as-a-reminde

### DIFF
--- a/AirCasting/CreateSessionViews/CreateSessionDetails/CreateSessionDetailsViewModel.swift
+++ b/AirCasting/CreateSessionViews/CreateSessionDetails/CreateSessionDetailsViewModel.swift
@@ -13,17 +13,21 @@ class CreateSessionDetailsViewModel: ObservableObject {
     @Published var wifiSSID: String = ""
     @Published var isConfirmCreatingSessionActive: Bool = false
     @Published var isLocationSessionDetailsActive: Bool = false
+    @Published var isLocationScreenNedeed: Bool = false
     @Published var showAlertAboutEmptyCredentials = false
     @Published var isSSIDTextfieldDisplayed: Bool = false
     @Published var showErrorIndicator: Bool = false
     var shouldShowError: Bool { sessionName.isEmpty && showErrorIndicator }
+    let locationHandler: LocationHandler
+    
     
     let baseURL: BaseURLProvider
     // It cannot be private as long as it is going to be passed
     // by every view which is included in session creation process
     
-    init(baseURL: BaseURLProvider) {
+    init(baseURL: BaseURLProvider, locationHandler: LocationHandler) {
         self.baseURL = baseURL
+        self.locationHandler = locationHandler
     }
     
     func onScreenEnter() {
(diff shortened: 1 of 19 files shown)


AirCasting/CreateSessionViews/CreateSessionDetails/CreateSessionDetailsViewModel.swift
```diff
@@ -64,8 +68,12 @@ class CreateSessionDetailsViewModel: ObservableObject {
     
     func compareIsIndoor(sessionContext: CreateSessionContext) -> CreateSessionContext {
         sessionContext.isIndoor = isIndoor
-        isLocationSessionDetailsActive = !isIndoor
-        isConfirmCreatingSessionActive = isIndoor
+        guard locationHandler.isLocationDenied() && !(sessionContext.isIndoor != nil) else {
```
anna1901, line 71: `!(sessionContext.isIndoor != nil)` what do we want to check here? I think `sessionContext.isIndoor == nil` would be easier to read.  However, do we want to move forward to `isLocationScreenNedeed = true` if location isn't denied but isIndoor is nil?

AirCasting/CreateSessionViews/PowerABView.swift
```diff
@@ -49,17 +50,9 @@ struct PowerABView: View {
...
     var continueButton: some View {
-        NavigationLink(destination: SelectPeripheralView(SDClearingRouteProcess: false, creatingSessionFlowContinues: $creatingSessionFlowContinues, urlProvider: urlProvider)) {
+        NavigationLink(destination: SelectPeripheralView(SDClearingRouteProcess: false, creatingSessionFlowContinues: $creatingSessionFlowContinues, urlProvider: urlProvider, locationHandler: locationHandler)) {
             Text(Strings.Commons.continue)
                 .frame(maxWidth: .infinity)
         }
     }
 }
-
-#if DEBUG
```
anna1901, line 59: Could you keep the preview please?

anna1901 (commented):
I'm trying to test this by uninstalling the app and installing again and then creating fixed session, but for some reason I'm never asked about permission to track location. It is automatically granted.

AirCasting/CreateSessionViews/CreateSessionDetails/CreateSessionDetailsViewModel.swift
jankrzempek: We were checking that to be sure that for the first session context must not be nil before checking the whole guard statement. After that, we were ensuring ourselves about location and indoor. I changed it anyway 😄. Thanks for asking

AirCasting/CreateSessionViews/PowerABView.swift
jankrzempek: I would like to delete it, as I talked to Paweł and after his merge of the architectural change, most of the previews will be gone from the app

anna1901 (approved):

AirCasting/CreateSessionViews/PowerABView.swift
pawelgil: Previews are really annoying to maintain at this point, we should remove all of them and re-introduce using more structured approach (if necessary), sharing Mock/Dummy data providers with unit test target.

PR #578, opened 2022-01-27
Save session url on upload

pawelgil:
# What was done
URLs received as a server response for uploaded sessions are now properly saved in the database

Trello: https://trello.com/c/3ti2pNwF

### DIFF
--- a/AirCasting/CoreData/NSManagedObjectContext+Utils.swift
+++ b/AirCasting/CoreData/NSManagedObjectContext+Utils.swift
@@ -52,6 +52,7 @@ extension NSManagedObjectContext {
     // and we decided to use a NSManagedObjectID instead. Somehow it's not working right, but when I added
     // additional session-id constraint it started to behave somewhat properly. But this is a blind patch
     // and should be investigated and resolved.
+    // Issue: https://github.com/HabitatMap/AirCastingiOS/issues/579
     
     // Checks if stream exists, if not, creates a new one
     func newOrExisting<T: NSManagedObject>(streamID: MeasurementStreamID, for session: SessionUUID? = nil) throws -> T  {
--- a/AirCasting/CoreData/NSManagedObjectContext+Utils.swift
+++ b/AirCasting/CoreData/NSManagedObjectContext+Utils.swift
@@ -39,9 +39,21 @@ extension NSManagedObjectContext {
         return new
     }
     
-    // Checks if stream exists, if not, creates a new one
+    
     #warning("This seems odd and is also working odd. I don't think this is correct implementation")
-    // NOTE: I added session id as an optional parameter to fix my scenario, but this is a patch
+    // Explaination:
+    // I'm not sure why is that happening, but the fetch inside this function is returning multiple streams
+    // not associated with a single session. It causes the calling code to sometimes modify an incorrect
+    // stream object. I found this when implementing the url saving feature after uploads
+    // (https://github.com/HabitatMap/AirCastingiOS/pull/578) - I used the `Databse`s
+    // `func insertOrUpdateSessions(_ sessions: [Database.Session], completion: ((Error?) -> Void)?)`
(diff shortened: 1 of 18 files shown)


AirCasting/CoreData/NSManagedObjectContext+Utils.swift
```diff
@@ -40,10 +40,17 @@ extension NSManagedObjectContext {
     }
     
     // Checks if stream exists, if not, creates a new one
-    func newOrExisting<T: NSManagedObject>(streamID: MeasurementStreamID) throws -> T  {
+    #warning("This seems odd and is also working odd. I don't think this is correct implementation")
```
anna1901, line 43: Do you have an idea for a better implementation? Cause I wouldn't know how to proceed with this warning and we are supposed to decrease the number of warnings since we already have a lot. 😅

AirCasting/Models/AirBeamCellularSessionCreator.swift
```diff
@@ -72,7 +72,8 @@ final class AirBeamCellularSessionCreator: SessionCreator {
                                                                 case .success(let output):
                                                                     measurementStreamStorage.accessStorage { storage in
                                                                         do {
-                                                                            try storage.createSession(session)
+                                                                            let sessionWithURL = session.withUrlLocation(output.location)
```
anna1901, line 75: Cool 👍🏻 What a nice solution

AirCasting/CoreData/Session.swift
```diff
@@ -144,3 +144,26 @@ public enum DeviceType: Int, CustomStringConvertible {
         }
     }
 }
+
+extension Session {
+    func withUrlLocation(_ newLocation) -> Self {
```
anna1901, line 149: `: String` is missing

anna1901 (changes requested):
Just add this `:String` so it builds. 😄 But with it it works perfectly! 👏🏻 Great job 🚀

AirCasting/CoreData/Session.swift
pawelgil: Weird

AirCasting/CoreData/NSManagedObjectContext+Utils.swift
pawelgil: The warning is there to be fixed asap, but this is too big of an issue to do it along this PR. I think we need to establish some sort of a scheme for dealing with those. I'd suggest:
**For potential bugs or severe code quality issues**
1) Place warnings in places that are *bad bad* (potentially breaking stuff, like here) and too big to resolve ad hoc.
2) When placing warning, add a comprehensive explanation of an issue in the comment. Also link the ticket from 3) in this comment.
3) Add a ticket to github issues section with correct tag (I suggest the tag "warning" + another one indicating severity high/mid/low or smth)

**For less severe stuff like minor code quality issues**
1) When the issue is not *bad bad*, but to big to resolve ad hoc, add a `// FIXME: ` marking, but explain it really well so that someone with more time that stumbles across this will be able to refactor/fix. Also link the ticket from 2) in this comment.
2) Add a ticket to github issues section with correct tag (I suggest the tag "quality" for quality issues and other tags for other issues. We can also create tags for parts of the app or to indicate if resolving this will require AirBeam).

What do you think about this @anna1901 @jankrzempek ? If/when we decide on some kind of a more formalized approach we shall add a section to the `README.md` file.

anna1901: I love the idea! This will help a lot to control all the warnings and issues that pop up and to start gathering parts of the app that need refactoring in one place.

anna1901 (approved):

AirCasting/CoreData/NSManagedObjectContext+Utils.swift
anna1901: Could you post this proposition on slack as well? Maybe we could move the discussion there and then pin it for future reference.

PR #584, opened 2022-01-31
ShareSheet handles mail case.

jankrzempek:
Now ShareSheet takes only URL as a share item parameters and based on the user decision it provides you with an entry email text or just a link to the session when copying only.

https://trello.com/c/3FTCLgbA/521-share-session-update-email-message-text-to-match-android


AirCasting/SessionViews/ShareSession/ShareLocationlessSessionView.swift
```diff
@@ -24,7 +24,7 @@ struct ShareLocationlessSessionView: View {
             .padding()
             .alert(item: $viewModel.alert, content: { $0.makeAlert() })
             .sheet(isPresented: $viewModel.showShareSheet, content: {
-                ActivityViewController(itemsToShare: [viewModel.file as Any]) { activityType, completed, returnedItems, error in
+                ActivityViewController(itemToShare: viewModel.file ?? URL(fileURLWithPath: "")) { activityType, completed, returnedItems, error in
```
pawelgil, line 27: is silently handling an empty `viewModel.file` good here? Maybe we need in that case?

jankrzempek: We have a warning when it is empty I think, so will it break anything in this way?

pawelgil: Won't break, but won't it be surprising to users? I mean - in this scenario they will be presented with a share sheet for empty URL. Don't we want to show an alert saying "oops, we couldn't fetch share URL for this session"? Or is it 100% guaranteed at this point that viewModel.file will be present?

anna1901 (approved):
Nice!

anna1901 (changes requested):
Sending zip file for locationless sessions is not working anymore, so that should be fixed

AirCasting/SessionViews/ShareSession/SharePage.swift
```diff
@@ -0,0 +1,52 @@
...
+        
+        init(_ parent: ActivityViewController) {
+            self.parent = parent
+        }
+        
+        func activityViewControllerPlaceholderItem(_ activityViewController: UIActivityViewController) -> Any {
+            return ""
+        }
+        
+        /// In email apps, other then Apple native one, the subject it taken/considered as the first line from the message body - which is writtien in the below func
```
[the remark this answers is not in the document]
jankrzempek: Thanks!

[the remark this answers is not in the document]
jankrzempek: Thanks!

AirCasting/SessionViews/ShareSession/ShareLocationlessSessionView.swift
jankrzempek: I made a few changes now and as I tested I can say that whenever there is no link there will be an alert presented to the user

pawelgil (approved):

AirCasting/SessionViews/ShareSession/ShareLocationlessSessionView.swift
```diff
@@ -24,7 +24,7 @@ struct ShareLocationlessSessionView: View {
             .padding()
             .alert(item: $viewModel.alert, content: { $0.makeAlert() })
             .sheet(isPresented: $viewModel.showShareSheet, content: {
-                ActivityViewController(itemsToShare: [viewModel.file as Any]) { activityType, completed, returnedItems, error in
+                ActivityViewController(isLocationless: true, itemToShare: viewModel.file!) { activityType, completed, returnedItems, error in
```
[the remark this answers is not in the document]
jankrzempek: Really like this idea, will do. thank a lot!

AirCasting/SessionViews/ShareSession/SharePage.swift
```diff
@@ -0,0 +1,53 @@
...
+        /// In email apps, other than Apple native one, the subject is taken/considered as the first line from the message body - which is written in the below func
+        func activityViewController(_ activityViewController: UIActivityViewController, itemForActivityType activityType: UIActivity.ActivityType?) -> Any? {
+            guard activityType == .mail else {
+                return """
+                        \(Strings.SessionShare.sharedEmailText)
+                        
+                        \(parent.itemToShare)
+                        """
+            }
+            return "\(Strings.SessionShare.sharedEmailText): \(parent.itemToShare)"
```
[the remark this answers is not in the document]
jankrzempek: I think that most of the users will use 'copy' option and paste it in the browser now so yes, thanks again!

AirCasting/SessionViews/ShareSession/ShareLocationlessSessionView.swift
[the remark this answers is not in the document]
anna1901: No problem. 😊 Maybe guard would be more swifty actually

AirCasting/SessionViews/ShareSession/SharePage.swift
```diff
@@ -0,0 +1,58 @@
+// Created by Lunar on 08/07/2021.
+//
+import MessageUI
+import SwiftUI
+
+struct ActivityViewController: UIViewControllerRepresentable {
+    typealias Callback = (_ activityType: UIActivity.ActivityType?, _ completed: Bool, _ returnedItems: [Any]?, _ error: Error?) -> Void
+    
+    let sharingFileisLocationless: Bool
```
anna1901, line 9: I think this generally apply to sharing file instead of a link, so no need to mention the locationless here

AirCasting/SessionViews/ShareSession/ShareLocationlessSessionViewModel.swift
```diff
@@ -54,6 +54,14 @@ class ShareLocationlessSessionViewModel: ObservableObject {
         exitRoute()
     }
     
+    func getSharePage() -> ActivityViewController? {
+        if file != nil {
```
anna1901, line 58: I know it is how I proposed it, but I noticed now that using guard would probably be more swifty version (at least that's what Paweł usually says 😄 )

anna1901 (approved):
Works very well! Great job 👏🏻🚀

AirCasting/SessionViews/ShareSession/ShareLocationlessSessionViewModel.swift
jankrzempek: Will do then 😎

PR #587, opened 2022-02-02
Strings preparation to be localized by XCode + use of NSLocalizedString.

jankrzempek:
LocalizedStringKey was changed to NSLocalizedString to allow automatic string detection and allows developers for easier exporting and importing. The pros of this implementation is, that we can always (at the end) get non-reformated string - just like we need.


anna1901 (approved):

AirCasting/Utils/Strings.swift
```diff
@@ -1,302 +1,436 @@
...
+        static let passwordPlaceholder: String = NSLocalizedString("Password",
+                                                                   comment: "")
+        static let connectButton: String = NSLocalizedString("Connect",
+                                                             comment: "")
+        static let passwordTitle: String = NSLocalizedString("WiFi network name & password:",
+                                                             comment: "")
+        static let nameAndPasswordTitle_1: String = NSLocalizedString("Password for ",
+                                                                      comment: "It consists of few parts. Please consider them together. Whole text is following: Password for network %@")
+        static let nameAndPasswordTitle_2: String = NSLocalizedString(" network:",
+                                                                      comment: "")
```
pawelgil, line 267: This is not enough to put a comment here. It won't work for languages that don't put the subject at the end of the sentence in this case. Can be worked around by providing an empty `nameAndPasswordTitle_2`, but this is not how it should be done. Use format string instead (`"Password for $@ network:"`).

AirCasting/Utils/Strings.swift
```diff
@@ -1,302 +1,436 @@
...
+                                                                  comment: "It consists of few parts. Please consider them together. Whole text is following: In mobile mode, the AirBeam captures personal exposures. In fixed mode, it can be installed indoors or outdoors to keep tabs on pollution levels in your home, office, backyard, or neighborhood 24/7.")
+        static let sheetDescription_2: String = NSLocalizedString("mobile ",
+                                                                  comment: "")
+        static let sheetDescription_3: String = NSLocalizedString("mode, the AirBeam captures personal exposures. In ",
+                                                                  comment: "")
+        static let sheetDescription_4: String = NSLocalizedString("fixed ",
+                                                                  comment: "")
+        static let sheetDescription_5: String = NSLocalizedString("mode, it can be installed indoors or outdoors to keep tabs on pollution levels in your home, office, backyard, or neighborhood 24/7.",
+                                                                  comment: "")
```
pawelgil, line 309: all this should be a single string too, but to incorporate this into SwiftUI it will be a bit of work, because `Text` doesn't let us style a single string with multiple formats. This split purpose is to distinguish "mobile" and "fixed" in this text. Knowing that we can detect those words in text in our view and split it or style it there. So steps to fix that:
1. Merge this string into one in the Strings file
2. Make sure in strings file we provide localized strings for "mobile" and "fixed" session types
3. Inside the view detect those words, split the string there and concat multiple `Text` accordingly

AirCasting/Utils/Strings.swift
```diff
@@ -305,300 +439,501 @@ struct Strings {
...
-
+ 
             Miscellaneous
-
+ 
             This Privacy Policy may be subject to updates and changes. In such case, HabitatMap will take steps to inform you of any such updates or changes by, for example, sending you a notification. Notwithstanding the foregoing, HabitatMap recommends that you review this page as often as possible.
-
- """).stringKey
+ 
+ """,
```
pawelgil, line 447: general note: it's bette to put privacy policy on a website and redirect there or make a special endpoint to provide policy text. Reasons:
1) Currently to update the policy we need to push new version to the appstore
2) If we update this and push new version, users that doesn't update the app won't be subjected to the new rules. This can cause some legal issues

Maybe smth to talk thru with M.

AirCasting/Utils/Strings.swift
```diff
@@ -305,300 +439,501 @@ struct Strings {
...
+                                                          comment: "")
+        static let alertMessage: String = NSLocalizedString("Please go to settings and allow location first.",
+                                                            comment: "")
+        static let alertSettings: String = NSLocalizedString("Settings",
+                                                             comment: "")
+        static let title: String = NSLocalizedString("What device are you using to record this session?",
+                                                     comment: "")
+        static let bluetoothLabel_1: String = NSLocalizedString("Bluetooth device",
+                                                                comment: "It consists of two parts. Please consider them together. Whole text is following: Bluetooth device for example AirBeam")
+        static let bluetoothLabel_2: String = NSLocalizedString("for example AirBeam",
```
pawelgil, line 502: those 2 should be merged too using one of the above methods

AirCasting/Utils/Strings.swift
```diff
@@ -305,300 +439,501 @@ struct Strings {
...
+                                                                comment: "")
+        static let micLabel_1: String = NSLocalizedString("Phone microphone",
+                                                          comment: "It consists of two parts. Please consider them together. Whole text is following: Phone microphone to measure sound level")
+        static let micLabel_2: String = NSLocalizedString("to measure sound level",
```
pawelgil, line 506: those 2 should be merged too using one of the above methods

AirCasting/Utils/Strings.swift
```diff
@@ -305,300 +439,501 @@ struct Strings {
...
+                                                                  comment: "")
+        static let finishAlertTitle_3_SYNC: String = NSLocalizedString("and sync from SD card?",
+                                                                       comment: "")
+        static let finishAlertMessage_1: String = NSLocalizedString("The session will be moved to ",
+                                                                    comment: "It consists of few parts. Please consider them together. Whole text is following: The session will be moved to Mobile Dormant tab and you won't be able to add new measurement to it.")
+        static let finishAlertMessage_2: String = NSLocalizedString("Mobile Dormant",
+                                                                    comment: "")
+        static let finishAlertMessage_3: String = NSLocalizedString(" tab and you won't be able to add new measurement to it.",
+                                                                    comment: "")
+        static let finishAlertMessage_4: String = NSLocalizedString("SD card will be cleared afterwards",
```
pawelgil, line 585: all multipart strings here should be merged into single ones using techniques described above

AirCasting/Utils/Strings.swift
```diff
@@ -305,300 +439,501 @@ struct Strings {
...
+                                                            comment: "")
+        static let mobileLabel_1: String = NSLocalizedString("Mobile session",
+                                                             comment: "It consists of few parts. Please consider them together. Whole text is following: Mobile session for moving around")
+        static let mobileLabel_2: String = NSLocalizedString("for moving around",
+                                                             comment: "")
+        static let orLabel: String = NSLocalizedString("or",
+                                                       comment: "")
+        static let syncTitle: String = NSLocalizedString("Sync data from AirBeam3",
+                                                         comment: "")
+        static let syncDescription: String = NSLocalizedString("if you recorded with AirBeam3",
```
pawelgil, line 633: all multipart strings here should be merged into single ones using techniques described above

AirCasting/Utils/Strings.swift
```diff
@@ -305,300 +439,501 @@ struct Strings {
...
+        static let contentViewText_1: String = NSLocalizedString("Your ", comment: "It consists of few parts. Please consider them together. Whole text is following: Your session is ready to start gathering data.")
+        static let contentViewText_2: String = NSLocalizedString(" session ",
+                                                                 comment: "")
+        static let contentViewText_3: String = NSLocalizedString(" is ready to start gathering data.",
+                                                                 comment: "")
+        static let contentViewText_4: String = NSLocalizedString("Hang your AirBeam in a secure position, then press the start recording button below.",
+                                                                 comment: "")
+        static let contentViewText_4Mobile: String = NSLocalizedString("Move to your starting location, confirm your location is accurate on the map, then press the start recording button below.",
+                                                                       comment: "")
+        static let startRecording: String = NSLocalizedString("Start recording",
```
pawelgil, line 713: all multipart strings here should be merged into single ones using techniques described above

AirCasting/Utils/Strings.swift
```diff
@@ -305,300 +439,501 @@ struct Strings {
...
+                                                        comment: "")
+        static let plusBlueIcon: String = NSLocalizedString("blueplus",
+                                                            comment: "")
+        static let settingsIcon: String = NSLocalizedString("settings",
+                                                            comment: "")
+        static let settingsBlueIcon: String = NSLocalizedString("bluesettings",
+                                                                comment: "")
+        static let loggingOut: String = NSLocalizedString("Logging out, please wait...",
+                                                          comment: "")
+    }
```
pawelgil, line 739: we don't want strings used to name button icons here at all

`HomeIcon`/`PlusIcon`/`SettingsIcon`

AirCasting/Utils/Strings.swift
```diff
@@ -305,300 +439,501 @@ struct Strings {
...
-        static let all: String = LocalizedStringKey("All").stringKey
+        static let all: String = NSLocalizedString("All",
+                                                   comment: "")
     }
-
+    
     enum DeviceHandler {
-        static let alertTitle = LocalizedStringKey("Not supported device").stringKey
-        static let alertMessage = LocalizedStringKey("To use the AirBeam3 in standalone mode and sync the SD card, an iPhone8 or higher is required").stringKey
+        static let alertTitle = NSLocalizedString("Not supported device",
```
pawelgil, line 830: a bit funky wording here. Maybe we should change to something better like `unsupprted device` or `this device is not supported`

AirCasting/Utils/Strings.swift
```diff
@@ -305,300 +439,501 @@ struct Strings {
...
-
+    
     enum InAppAlerts {
-        static let assertError: String = LocalizedStringKey("Unsupported button count! For SwiftUI implementation max of 2 buttons is supported").stringKey
-        static let unableToLogOutTitle: String = LocalizedStringKey("No internet connection").stringKey
-        static let unableToLogOutMessage: String = LocalizedStringKey("To sign out, you must be connected to the Internet.").stringKey
-        static let unableToLogOutButton: String = LocalizedStringKey("Got it!").stringKey
-    }
-
+        static let assertError: String = NSLocalizedString("Unsupported button count! For SwiftUI implementation max of 2 buttons is supported",
```
pawelgil, line 851: we don't really want to localize strings we use internally for asserts. This file should only contain user-facing strings

AirCasting/Utils/Strings.swift
```diff
@@ -1,302 +1,436 @@
...
+        static let clearSDTitle = NSLocalizedString("Clear SD card",
+                                                    comment: "")
+        static let appInfoTitle = NSLocalizedString("AirCasting App v",
+                                                    comment: "")
+        static let buildText = NSLocalizedString("build",
+                                                 comment: "")
+        static let betaBuild = NSLocalizedString("Beta build",
+                                                 comment: "")
+        static let debugBuild = NSLocalizedString("Debug build",
+                                                  comment: "")
```
pawelgil, line 68: I think we can remove beta/debug only strings from localization at all. It's our internal thing and would be a waste of money or/and time to localize those.

pawelgil (changes requested):
Good job!

To properly prepare for localizing the strings there are some much needed steps. We want to make the translation process as easy as it can be for translators. I know it's laborious, but let's do it.

AirCasting/Utils/Strings.swift
jankrzempek: I will write a message to him and discuss it, thanks!

AirCasting/Utils/Strings.swift
jankrzempek: Sure, changing!

AirCasting/Utils/StringsCustomizer.swift
```diff
@@ -0,0 +1,41 @@
+// Created by Lunar on 07/02/2022.
+//
+
+import SwiftUI
+
+final class StringCustomizer {
+    static func customizeString(_ s: String, using words: [String], fontWeight: Font.Weight = .bold, color: Color = .aircastingGray, font: Font = .body, makeNewLineAfterCustomized: Bool = false) -> Text {
```
pawelgil, line 7: that is neat, very well thought out!

AirCasting/Utils/StringsCustomizer.swift
```diff
@@ -0,0 +1,41 @@
...
+        
+        var descriptionText: Text!
+        let separatedText = s.components(separatedBy: " ")
+        var total = [String]()
+        var newL = false
+        
+        words.forEach { wd in
+            let sp = wd.components(separatedBy: " ")
+            sp.forEach({ total.append( $0 ) })
+        }
```
pawelgil, line 17: this takes a while to decipher. I would suggest using [flatMap(_:)](https://developer.apple.com/documentation/swift/sequence/2905332-flatmap). It will make it a one-liner:
```Swift
let total = words.flatMap { $0.components(separatedBy: " ") }
```

But please rename this `total` variable into something more descriptive as it is saying the reader nothing ;)

AirCasting/Utils/StringsCustomizer.swift
```diff
@@ -0,0 +1,41 @@
...
+        
+        for w in separatedText {
+            if total.contains(w) {
+                if let index = total.firstIndex(of: w) {
+                    total.remove(at: index)
+                }
```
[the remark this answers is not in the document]
jankrzempek: You are right, but this was intended. In every case, we have it the app should be covered, but without this, we have a special case: "----- AirBeam3 ----- AirBeam3---" where the only first word should be blue. That was the reason for this change

AirCasting/Utils/StringsCustomizer.swift
```diff
@@ -0,0 +1,41 @@
...
+        var total = [String]()
+        var newL = false
+        
+        words.forEach { wd in
+            let sp = wd.components(separatedBy: " ")
+            sp.forEach({ total.append( $0 ) })
+        }
+        
+        for w in separatedText {
+            if total.contains(w) {
```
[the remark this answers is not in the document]
jankrzempek: I think I won't be able to remove it as it is used in a few places and thanks to that I think a new line can be easily added. I till try to split then :)

AirCasting/Utils/StringsCustomizer.swift
```diff
@@ -0,0 +1,61 @@
...
+                    .fontWeight(.regular)
+                    .foregroundColor(.aircastingGray)
+                newLine = false
+            }
+            customizedText = addSpace(after: customizedText)
+        }
+        return customizedText
+    }
+    
+    //MARK: - Private
```
pawelgil, line 35: space after //

pawelgil (approved):

anna1901 (changes requested):
Screenshot 2022-02-17 at 13 29 52
While going through fixed session creation I got this fatal error when I confirmed session detailes.

AirCasting/Utils/StringsCustomizer.swift
```diff
@@ -0,0 +1,61 @@
...
+        separatedText.forEach { word in
+            if separatedKeyWords.contains(word) {
+                removeUsedWord(separatedKeyWords: &separatedKeyWords, word: word)
+                newLine = true
+                customizedText = appended(word: word, to: customizedText)
+                    .fontWeight(fontWeight)
+                    .foregroundColor(color)
+                    .font(font)
+            } else {
+                customizedText = shouldAddEmptyLine(after: customizedText,
```
anna1901, line 23: CustomizedText doesn't have any value here

AirCasting/Utils/StringsCustomizer.swift
```diff
@@ -0,0 +1,61 @@
+// Created by Lunar on 07/02/2022.
+//
+
+import SwiftUI
+
+final class StringCustomizer {
+    static func customizeString(_ s: String, using words: [String], fontWeight: Font.Weight = .bold, color: Color = .aircastingGray, font: Font = .body, makeNewLineAfterCustomized: Bool = false) -> Text {
+        
+        var customizedText: Text!
```
anna1901, line 9: Why "!" ?

AirCasting/Utils/StringsCustomizer.swift
anna1901: The crash happens both for mobile and fixed sessions

AirCasting/Utils/StringsCustomizer.swift
jankrzempek: Otherwise, we would get an error, and using unwrapping we are ensuring compiler that customizedText will have a value in the future

AirCasting/Dashboard/EmptyMobileDashboardView.swift
```diff
@@ -49,11 +49,14 @@ private extension EmptyMobileDashboardViewMobile {
     
     private var emptyMobileDashboardText: some View {
         VStack(spacing: 14) {
-            Text(Strings.EmptyDashboardMobile.title)
-                .font(Fonts.boldTitle4)
-                .foregroundColor(Color.darkBlue)
-                .minimumScaleFactor(0.1)
-            
+            StringCustomizer.customizeString(Strings.EmptyDashboardMobile.title,
+                                             using: ["a"],
```
pawelgil, line 53: will "a" be universal word here across all languages? If not it needs to be moved out

AirCasting/Utils/StringsCustomizer.swift
anna1901: But if it doesn't then the app crashes and the compiler doesn't notify us about that. I thought generally it's better not to use the force unwrap

AirCasting/Utils/StringsCustomizer.swift
```diff
@@ -0,0 +1,60 @@
...
+
+final class StringCustomizer {
+    static func customizeString(_ s: String, using words: [String], fontWeight: Font.Weight = .bold, standardFontWeight: Font.Weight = .regular, color: Color = .aircastingGray, standardColor: Color = .aircastingGray, font: Font = .body, standardFont: Font = Fonts.muliHeading3, makeNewLineAfterCustomized: Bool = false) -> Text {
+        
+        var customizedText: Text!
+        let separatedText = s.components(separatedBy: " ")
+        var newLine = false
+        var separatedKeyWords = words.flatMap { $0.components(separatedBy: " ") }
+        
+        separatedText.forEach { word in
```
anna1901, line 14: If s would be an empty string this function would crash, right?

AirCasting/Utils/StringsCustomizer.swift
anna1901: Exactly, I was going to propose the same.

AirCasting/Utils/StringsCustomizer.swift
```diff
@@ -0,0 +1,60 @@
...
+            separatedKeyWords.remove(at: index)
+        }
+    }
+    
+    private static func addSpace(after text: Text) -> Text {
+        text + Text(" ")
+    }
+    
+    private static func appended(word: String, to text: Text?) -> Text {
+        return (text == nil ? Text(word) : text! + Text(word))
```
anna1901, line 49: This can be just `text + Text(word))` and the text doesn't have to be optional now

anna1901 (approved):
👏🏻

PR #593, opened 2022-02-06
Additional trigger for ChartViewModel added

pawelgil:
Fixes bug where chart view would not refresh properly by adding database triggers to viewmodel


AirCasting/SessionViews/ChartViewModel.swift
```diff
@@ -13,9 +13,15 @@ final class ChartViewModel: ObservableObject {
...
     var stream: MeasurementStreamEntity? {
         didSet {
-            guard session.isActive || session.isFollowed || session.status == .NEW else { return }
+            databaseObserver = nil
+            guard shouldBeUpdatingChartForCurrentSession(), let sensor = stream?.sensorName else { return }
+            databaseObserver = ChartDatabaseObserver(session: session.uuid, sensor: sensor) {
+                Log.info("Measurements change detected [\(self.session.uuid ?? "??")]")
+                self.generateEntries()
+            }
             generateEntries()
```
jankrzempek, line 25: Do we want to use  `generateEntries()`, despite we used it above? If so, could you tell me why, please

jankrzempek (approved):
Testing right now!

AirCasting/SessionViews/ChartViewModel.swift
pawelgil: note that one call will be executed right away (the latter), but the other will execute only after a measurements change.

AirCasting/SessionViews/ChartViewModel.swift
```diff
@@ -13,9 +13,15 @@ final class ChartViewModel: ObservableObject {
...
 
+    private var databaseObserver: ChartDatabaseObserver?
     var stream: MeasurementStreamEntity? {
         didSet {
-            guard session.isActive || session.isFollowed || session.status == .NEW else { return }
+            databaseObserver = nil
+            guard shouldBeUpdatingChartForCurrentSession(), let sensor = stream?.sensorName else { return }
+            databaseObserver = ChartDatabaseObserver(session: session.uuid, sensor: sensor) {
+                Log.info("Measurements change detected [\(self.session.uuid ?? "??")]")
+                self.generateEntries()
```
anna1901, line 23: Could you explain why we don't need a weak self here?

AirCasting/SessionViews/ChartDatabaseObserver.swift
```diff
@@ -0,0 +1,59 @@
...
+final class ChartDatabaseObserver {
+    @Injected private var persistence: PersistenceController
+    private let onMeasurementsChange: (() -> Void)
+    private let session: SessionUUID
+    private let sensor: String
+    private let filteringComponent: Calendar.Component?
+    
+    private var lastMeasurementChangeAt: Date?
+    
+    init(session: SessionUUID, sensor: String, timedFilter: ChartDatabaseObserverFilter, onMeasurementsChange: @escaping (() -> Void)) {
```
anna1901, line 23: What does `timedFilter` mean? Should it be `timeFilter`?

AirCasting/SessionViews/ChartViewModel.swift
```diff
@@ -13,9 +13,17 @@ final class ChartViewModel: ObservableObject {
...
     @ObservedObject var session: SessionEntity
 
+    private var databaseObserver: ChartDatabaseObserver?
     var stream: MeasurementStreamEntity? {
         didSet {
-            guard session.isActive || session.isFollowed || session.status == .NEW else { return }
+            databaseObserver = nil
+            guard shouldBeUpdatingChartForCurrentSession(), let sensor = stream?.sensorName else { return }
+            let filterBy: ChartDatabaseObserverFilter = session.isFixed ? .hour : .minute
+            databaseObserver = ChartDatabaseObserver(session: session.uuid, sensor: sensor, timedFilter: filterBy) { [weak self] in
```
anna1901, line 22: Perhaps we could have one ChartDatabaseObserver per chart that would notify chartViewModel about new measurements for this session and call the generateEntries function, but the charViewModel would know for which stream it should generate entries?

AirCasting/SessionViews/ChartDatabaseObserver.swift
```diff
@@ -0,0 +1,59 @@
...
+        lastMeasurementChangeAt = DateBuilder.getRawDate()
+        onMeasurementsChange()
+    }
+    
+    private func shouldFireObserver() -> Bool {
+        guard let lastMeasurementChangeAt = lastMeasurementChangeAt, let filteringComponent = filteringComponent else { return true }
+        let now = DateBuilder.getRawDate()
+        guard Calendar.current.isDate(lastMeasurementChangeAt, inSameDayAs: now) else { return true }
+        let nowComponent = Calendar.current.component(filteringComponent, from: now)
+        let lastMeasurementComponent = Calendar.current.component(filteringComponent, from: lastMeasurementChangeAt)
```
anna1901, line 46: This is a nice idea, but it would work as intended only for fixed sessions. For mobile sessions, we don't want measurements for a specific minute of time but for a specific minute of session duration. I mean, for a session which started eg. at 9:01:30 we don't want measurements from 9:02:00 to 9:02:59 but from 9:01:30 to 9:02:30

AirCasting/SessionViews/ChartDatabaseObserver.swift
pawelgil: changed to `filtered`

AirCasting/SessionViews/ChartViewModel.swift
pawelgil: We could, but IMO putting it into the minute/hour enum and moving the now-generic filtering logic inside observer logic is a bit more neat and will keep low-level logic away from viewmodel

anna1901: I'm not sure I understand your answer and if I explained myself well so I'll try again. I meant that we could not pass the sensor information to the chartDatabaseObserver and have just one observer for the whole session since we get measurements in each stream at once. The hour/minute filter would still be relevant, as it depends on the session type. I think we don't need to filter measurements in the observer, as we are doing it only doing it now to check if there are more than 0 measurements for that stream.

AirCasting/SessionViews/ChartDatabaseObserver.swift
```diff
@@ -33,27 +33,61 @@ final class ChartDatabaseObserver {
...
+        case .minute(let countingFrom): return isInNewMinute(measurement: newMeasurement, countingFrom: countingFrom)
+        }
+    }
+    
+    private func isInNewHour(measurement: MeasurementEntity) -> Bool {
+        guard let latestMeasurementTime = latestMeasurementTime else { return true }
+        guard Calendar.current.isDate(latestMeasurementTime, inSameDayAs: measurement.time) else { return true }
+        let nowComponent = Calendar.current.component(.hour, from: measurement.time)
+        let latestComponent = Calendar.current.component(.hour, from: latestMeasurementTime)
+        return latestComponent == nowComponent
```
anna1901, line 60: Based on name shouldn't it be `!=`?

AirCasting/SessionViews/ChartViewModel.swift
pawelgil: > "we get measurements in each stream at once"

this is the key part IMO here. By doing so we'll be implicitly depending on this behavior and there is a very important (but not very obvious) code quality issue with that:
This is not a set-in-stone, core type of behavior that we should rely on. It's more like a side effect of our (or backend) previous design choices. It might change due to some completely different task in completely different module in the future and this class will break all of the sudden. This is called **fragile code** and should be avoided. We do so by being very verbose and specific here. We're operating in a context of specific session and stream, let's design this interface to be conformant with that.

AirCasting/SessionViews/ChartDatabaseObserver.swift
pawelgil: good catch, changed it (and corresponding call) to match the minute specific func naming

AirCasting/SessionViews/ChartViewModel.swift
pawelgil: In the future we can actually move this observer to be hidden behind a protocol and inject different implementations based on session:

`Resolver.resolve(ChartMeasurementsObserver.self, args: [sessionId, sensorName])`

This way we can decouple viewmodel from database and also remove the need for conditional `if fixed`

AirCasting/SessionViews/ChartViewModel.swift
```diff
@@ -15,79 +15,134 @@ final class ChartViewModel: ObservableObject {
...
+        guard shouldBeUpdatingChartForCurrentSession(), let sensor = stream?.sensorName else { return }
+        let filter: ChartDatabaseObserverFilter = session.isFixed ? .hour : .minute(countingFrom: session.startTime?.convertedFromUTCToLocal ?? DateBuilder.getRawDate())
+        databaseObserver = ChartDatabaseObserver(session: session.uuid, sensor: sensor, filtered: filter) { [weak self] in
+            guard let self = self else { return }
+            self.log("Measurements change detected")
+            self.generateEntries()
+        }
+    }
+    
+    // MARK: Timers
```
anna1901, line 71: Great idea with those marks. We can probably remove the timers version soon. 👍🏻

AirCasting/SessionViews/ChartViewModel.swift
```diff
@@ -15,79 +15,134 @@ final class ChartViewModel: ObservableObject {
...
+    
+    private var timeUnit: TimeInterval {
```
anna1901, line 73: The timeUnit is relevant in both implementation (we use it in generate entries method)

AirCasting/SessionViews/ChartViewModel.swift
```diff
@@ -15,79 +15,134 @@ final class ChartViewModel: ObservableObject {
...
     }
+    
+    // MARK: Chart refresh
 
     func refreshChart() {
         generateEntries()
     }
 
     private func generateEntries() {
+        Log.verbose("Generating entries")
```
anna1901, line 137: Could you add session information here as well?

AirCasting/SessionViews/ChartViewModel.swift
```diff
@@ -15,79 +15,134 @@ final class ChartViewModel: ObservableObject {
...
         // Set up begning and end of the interval for the first average
         //  - for fixed sessions we are taking the last full hour of the session, which has any measurements
         //  - for mobile we are taking into account full minutes since the session started and we are taking the most recent one
         guard
             stream != nil,
             var intervalEnd = intervalEndTime()
         else {
+            Log.warning("Generating entried failed!")
```
anna1901, line 145: and here

AirCasting/SessionViews/ChartViewModel.swift
```diff
@@ -15,79 +15,134 @@ final class ChartViewModel: ObservableObject {
...
+        log("Starting chart timers (time of first entry regen: \(timeOfNextAverage) (\(Date().addingTimeInterval(timeOfNextAverage)))")
         firstTimer = Timer.scheduledTimer(withTimeInterval: TimeInterval(timeOfNextAverage), repeats: false) { [weak self] timer in
+            self?.log("Initial timer fired for chart")
             self?.generateEntries()
             self?.startMainTimer()
         }
     }
+    
+    private func stopTimers() {
+        Log.verbose("Stopping chart timers")
```
anna1901, line 101: Could you add session info here?

AirCasting/SessionViews/SessionCartFollowing.swift
```diff
@@ -32,7 +32,8 @@ extension SessionFollowing {
          set {
              #warning("❌This 'setter' is asynchronous!!!")
```
anna1901, line 33: Btw what do you think about this warning? Is it still relevant? Is it a problem that it is asynchronous?

anna1901 (commented):
The whole thing works almost perfectly! Great job with the whore PR. I just noticed one bug: when I logged in and I went to fixed sessions tab and followed a session and went to the followed tab and expanded the card before we downloaded the measurements, then there wasn't any chart entries (I saw only "waiting for the first averages") even after the measurements were downloaded and displayed in last minute measurements section. So that's one edge case we could still look into.

AirCasting/SessionViews/SessionCartFollowing.swift
pawelgil: It can be a problem. If we set the value of this property and than shortly after (before database queue is processed) read from it. The value returned will still be an old one. It needs some analysis if that's a problem, but it should be somehow communicated to the user of this class (warning is not great)

pawelgil:
Note: timers have been completely removed now in this PR. It needs some testing. I think database observing is sufficient and removes many problems, but I'm not sure I tested it all

AirCasting/Utils/DataFormatters.swift
```diff
@@ -116,6 +116,14 @@ enum DateFormatters {
             return dateFormatter
         }()
     }
+    
+    enum Debug {
+        static let logsFormatter: DateFormatter = {
+            let df = DateFormatter()
+            df.dateFormat = "yyyy-MM-dd HH:mm:ss"
```
anna1901, line 123: Can we display this time in UTC so that it is consistent with what we store in db?

anna1901 (approved):

PR #598, opened 2022-02-10
Clean boot option added

pawelgil:
# What was done?

Clean boot option added to environment variables.

When turned on it will clean all database data and log current user out on app launch. Really convenient when debugging/developing stuff that needs fresh login each time.
Zrzut ekranu 2022-02-10 o 12 32 22

To turn it on go to the "edit scheme menu" and turn it on:

### DIFF
--- a/AirCasting/AirCastingApp.swift
+++ b/AirCasting/AirCastingApp.swift
@@ -63,6 +63,15 @@ final class SynchronizationScheduler {
         authorization
             .$isLoggedIn
             .removeDuplicates()
+            .scan((nil, nil)) {
+                ($0.1, $1)
+            }
+            .logVerbose { "Logged in state changed from \(String(describing: $0.0)) to \(String(describing: $0.1))" }
+            .compactMap {
+                guard let _ = $0.0 else { return nil }
+                guard let new = $0.1 else { return nil }
+                return new
+            }
             .filter { $0 }
             .eraseToVoid()
             .sink {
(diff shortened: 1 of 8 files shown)


jankrzempek (approved):

AirCasting/AirCastingApp.swift
```diff
@@ -63,6 +63,15 @@ final class SynchronizationScheduler {
         authorization
             .$isLoggedIn
             .removeDuplicates()
+            .scan((nil, nil)) {
```
anna1901, line 66: Could you explain more or less what does it do?

pawelgil: of course :)

I noticed this authorization observer will trigger even when authorization is set for the first time during runtime. This happens right when the app launches - for that we already have an observer. To make this clearer, step by step scenario of this false-positive case:
1. User launches the app and they've been already logged in during previous session
2. We notice that and set `authorization.isLoggedin` to `true` as we should.
3. This observer gets triggered. And it should not since the user didn't actually perform a login.

This one should **only** trigger when we switch from logged-out state to logged-in state. To accomplish that I'm using [scan(_:_:)](https://developer.apple.com/documentation/combine/fail/scan(_:_:)) in order to transform a `Bool` publisher into a `(Bool, Bool)` publisher which contains previous value (`$0.0`) and the newly set one (`$0.1`) so we can detect desired scenario.

PR #599, opened 2022-02-11
Manage UIState using CoreData and UIStateEntity.

jankrzempek:
The aim of this branch is to keep track of the UI state, especially the card state and selected measurement state. It is done, using the CoreData and newly created UIStateEntity which has two needed parameters.

https://trello.com/c/bOwMTfOG/388-following-remember-last-known-state


AirCasting/CoreData/UIStorage.swift
```diff
@@ -0,0 +1,69 @@
...
+    func cardStateToggle(for sessionUUID: SessionUUID) throws {
+        let sessionEntity = try context.existingSession(uuid: sessionUUID)
+        checkUIStateAvaibility(entity: sessionEntity)
+        sessionEntity.userInterface?.expandedCard.toggle()
+        try context.save()
+    }
+    
+    func changeStream(for sessionUUID: SessionUUID, stream: String) throws {
+        let sessionEntity = try context.existingSession(uuid: sessionUUID)
+        checkUIStateAvaibility(entity: sessionEntity)
```
anna1901, line 54: I'm thinking if there is a name that better describes the behavior of this method. As I understand it creates the UIState entity for session if it didn't have it yet, so maybe it could be named: `createUIStateForSessionIfNeeded` or something like that?

AirCasting/CoreData/MeasurementStreamStorage.swift
```diff
@@ -380,9 +380,16 @@ final class HiddenCoreDataMeasurementStreamStorage: MeasurementStreamStorageCont
                     long: note.long,
                     number: Int(note.number))
     }
+    
+    private func newSessionEntity() -> SessionEntity {
```
anna1901, line 384: I thought we were supposed to remember the state only for followed sessions, based on the trello ticket title. Was that changed? If we should do it only for followed sessions, maybe we should add UIState to session only on follow button click?

anna1901 (commented):
Very nicely done! 👏🏻 I'm just having doubts if it wasn't supposed to be done only for following tab

AirCasting/CoreData/SessionEntity.swift
```diff
@@ -39,6 +38,9 @@ public class SessionEntity: NSManagedObject, Identifiable {
     @NSManaged public var changesCount: Int32
     @NSManaged public var rowOrder: Int64
     
+    // UserInterface state
+    @NSManaged public var userInterface: UIStateEntity?
```
anna1901, line 42: I'm thinking whether it wouldn't be better not to add this relationship to session entity but rather add SessionUUID property to UIStateEntity and operate on that. What do you think about that? This way we wouldn't have to add this UIState logic to Session logic and just keep it in the session card view for followed sessions

AirCasting/CoreData/UIStorage.swift
jankrzempek: I need to say your idea sound better than mine. Thank you 😄 I will change in a moment

AirCasting/CoreData/MeasurementStreamStorage.swift
jankrzempek: I missed it totally. I will talk to Michael, thanks!

AirCasting/CoreData/SessionEntity.swift
jankrzempek: I need to say I would like to leave it like it is if you agree. The reasons are:
- now it is well connected with a relation 1:1 and everything is clear
- whenever we would need to find a proper UIState (when using UUID) we would need to iterate and find a proper one then. I think it is not needed
- The pros is whenever we delete the session card we get rid of the UIState too. In the UUID way, we would need to iterate and delete it separately

What do you think?

pawelgil: I agree with @jankrzempek here. On CoreData level this tradeoff seems really beneficial - pros outweigh cons. As long as we keep it well separated on the interface level so the rest of the app see the ui state database and core database as separate modules.

AirCasting/CoreData/UIStorage.swift
```diff
@@ -0,0 +1,69 @@
...
+    
+    init(context: NSManagedObjectContext) {
+        self.context = context
+    }
+    
+    func cardStateToggle(for sessionUUID: SessionUUID) throws {
+        let sessionEntity = try context.existingSession(uuid: sessionUUID)
+        createUIStateIfNeeded(entity: sessionEntity)
+        sessionEntity.userInterface?.expandedCard.toggle()
+        try context.save()
```
pawelgil, line 49: Isn't `save()` guaranteed by `acessStorage(_:)`?

AirCasting/CoreData/UIStorage.swift
```diff
@@ -0,0 +1,69 @@
...
+    }
+    
+    func changeStream(for sessionUUID: SessionUUID, stream: String) throws {
+        let sessionEntity = try context.existingSession(uuid: sessionUUID)
+        createUIStateIfNeeded(entity: sessionEntity)
+        sessionEntity.userInterface?.sensorName = stream
+        try context.save()
```
pawelgil, line 56: Isn't `save()` guaranteed by `acessStorage(_:)`?

AirCasting/UIState/UIState.swift
```diff
@@ -0,0 +1,10 @@
+// Created by Lunar on 10/02/2022.
+//
+
+import Foundation
+import CoreLocation
+
+struct UIState {
```
pawelgil, line 7: `SessionCardUIState` maybe?

AirCasting/UIState/UIState.swift
```diff
@@ -0,0 +1,10 @@
...
+    let expandedCard: Bool
```
pawelgil, line 8: if this struct is called `SessionCardUIState` we can rename it to just be `isExpanded`

AirCasting/UIState/UIState.swift
```diff
@@ -0,0 +1,10 @@
...
+    let sensorName: String
```
pawelgil, line 9: `selectedSensorName` perhaps?

AirCasting/CoreData/UIStorage.swift
```diff
@@ -0,0 +1,69 @@
+// Created by Lunar on 10/02/2022.
+//
+
+import CoreData
+import Combine
+import SwiftUI
```
pawelgil, line 6: do we use `SwiftUI` here? (and `Combine`?)

AirCasting/CoreData/UIStorage.swift
```diff
@@ -0,0 +1,69 @@
...
+
+protocol UIStorageContextUpdate {
+    func cardStateToggle(for sessionUUID: SessionUUID) throws
+    func changeStream(for sessionUUID: SessionUUID, stream: String) throws
+    func save() throws
+}
+
+final class CoreDataUIStorage: UIStorage {
+    
+    @Injected private var persistenceController: PersistenceController
```
pawelgil, line 21: hmm... Let's maybe not inject it here, instead initialize this class with context and move figuring it out to Resolver's `register` func for `UIStorage`

AirCasting/CoreData/UIStorage.swift
```diff
@@ -0,0 +1,69 @@
...
+    private lazy var updateSessionParamsService = UpdateSessionParamsService()
```
pawelgil, line 22: would be nice to move `UpdateSessionParamsService` to `Resolver` (be a nice scout ;) )

AirCasting/UIState/UIStateHandler.swift
```diff
@@ -0,0 +1,35 @@
+// Created by Lunar on 10/02/2022.
+//
+
+import Foundation
+import Resolver
+
+protocol UIStateHandler {
```
pawelgil, line 7: `SessionCardUIStateHandler` maybe? By naming it this way we discourage from putting more and more methods here and instead create separate interfaces for separate parts

AirCasting/UIState/UIStateHandler.swift
```diff
@@ -0,0 +1,35 @@
...
+    func cardToggle(sessionUUID: SessionUUID)
```
pawelgil, line 8: `toggleCardExpanded` maybe? `cardToggle` is not very descriptive

AirCasting/UIState/UIStateHandler.swift
```diff
@@ -0,0 +1,35 @@
...
+    func streamChange(sessionUUID: SessionUUID, newStream: String)
```
pawelgil, line 9: `changeSelectedStream` maybe?

AirCasting/UIState/UIStateHandler.swift
```diff
@@ -0,0 +1,35 @@
...
+}
+
+class UIStateHandlerDefault: UIStateHandler {
+    @Injected private var UIStorage: UIStorage
```
pawelgil, line 13: we always name proeprties `lowerCamelCased`, no capital letters in front plz :)

AirCasting/UIState/UIStateHandler.swift
```diff
@@ -0,0 +1,35 @@
...
+            }
+        }
+    }
+    
+    func streamChange(sessionUUID: SessionUUID, newStream: String) {
+        UIStorage.accessStorage { storage in
+            do {
+                try storage.changeStream(for: sessionUUID, stream: newStream)
+            } catch {
+                Log.info("(UIState) Error changing stream \(error)")
```
pawelgil, line 30: `(UIState)` is not needed since logger will log automatically which file, function and line the log originates from (when my PR is merged). Also this should be warning or error log, not info

AirCasting/UIState/UIStateHandler.swift
```diff
@@ -0,0 +1,35 @@
...
+
+class UIStateHandlerDefault: UIStateHandler {
+    @Injected private var UIStorage: UIStorage
+    
+    func cardToggle(sessionUUID: SessionUUID) {
+        UIStorage.accessStorage { storage in
+            do {
+                try storage.cardStateToggle(for: sessionUUID)
+            } catch {
+                Log.info("(UIState) Error toggling card state \(error)")
```
pawelgil, line 20: `(UIState)` is not needed since logger will log automatically which file, function and line the log originates from (when my PR is merged). Also this should be warning or error log, not info

AirCasting/CoreData/UIStorage.swift
```diff
@@ -0,0 +1,69 @@
...
+    func changeStream(for sessionUUID: SessionUUID, stream: String) throws {
+        let sessionEntity = try context.existingSession(uuid: sessionUUID)
+        createUIStateIfNeeded(entity: sessionEntity)
+        sessionEntity.userInterface?.sensorName = stream
+        try context.save()
+    }
+    
+    private func createUIStateIfNeeded(entity: SessionEntity) {
+        if entity.userInterface == nil {
+            let UIState = UIStateEntity(context: context)
```
pawelgil, line 61: `lowerCased` variable name plz

pawelgil (changes requested):
Good job! Let's make some small changes, mostly naming

AirCasting/CoreData/UIStorage.swift
jankrzempek: Yeah, it is - thanks!

AirCasting/CoreData/UIStorage.swift
jankrzempek: 😬 copying leftover

AirCasting/CoreData/UIStorage.swift
jankrzempek: I started to do this, but I am not sure if I get it as it should be. Could you tell me when you will be back from ⛷?
Why do we want to skip Injection in this case?

jankrzempek: I mean, should I initialize context using the persistenceController that I will require in init?

AirCasting/CoreData/UIStorage.swift
```diff
@@ -16,11 +16,14 @@ protocol UIStorageContextUpdate {
 
 final class CoreDataUIStorage: UIStorage {
     
-    @Injected private var persistenceController: PersistenceController
+    private var context: NSManagedObjectContext
```
pawelgil, line 19: now it can be a `let` :)

pawelgil (approved):

PR #602, opened 2022-02-14
Sd sync > fixed sessions: upload only sessions that belong to the user

anna1901:
https://trello.com/c/1Y4cJzlC


AirCasting/ABConnector/SDSyncController.swift
```diff
@@ -115,10 +111,20 @@ class SDSyncController {
...
+            switch result {
+            case .success(let sessions):
+                uploadedSessions = sessions
+                completion(true)
+            case .failure(let error):
+                Log.error("[SD Sync] Failed to upload sessions to backend: \(error.localizedDescription)")
+                completion(false)
+            }
+        }
+        return uploadedSessions
```
pawelgil, line 127: isn't processAnUpload asynchronous tho? It mens this will always return `nil`. You need to put this into completion rather than standard return.

On that note we would be able to start using the new async/await sweetness now since it's supported for iOS13+.

AirCasting/SDSync/FixedSessions/SDCardFixedSessionsSavingService.swift
```diff
@@ -2,19 +2,90 @@
...
+        
+        var readLines = 0
+        
+        measurementStreamStorage.accessStorage { storage in
+            do {
+                try self.fileLineReader.readLines(of: fileURL, progress: { line in
+                    switch line {
+                    case .line(let content):
+                        readLines += 1
+                        Log.info("LINE: \(content)")
```
pawelgil, line 35: won't this flood our logs **real hard**?

AirCasting/SDSync/FixedSessions/SDCardFixedSessionsSavingService.swift
```diff
@@ -2,19 +2,90 @@
...
+                        let measurementsRow = self.parser.parseMeasurement(lineString: content)
+                        guard let measurements = measurementsRow, !sessionsToIgnore.contains(measurements.sessionUUID) else { return }
+                        
+                        let session = processedSessions.first(where: { $0 == measurements.sessionUUID })
```
pawelgil, line 39: won't it be easier and faster to make `processedSessions` a set instead of an array?

Edit: I see you used Set, kudos. Use `processedSessions.contains(measurements.sessionUUID)` then plz (it will be faster and easier to read)

AirCasting/SDSync/FixedSessions/SDCardFixedSessionsSavingService.swift
```diff
@@ -2,19 +2,90 @@
...
+        // We don't want to upload sessions of other users
+        var sessionsToIgnore: [SessionUUID] = []
+        
+        var readLines = 0
+        
+        measurementStreamStorage.accessStorage { storage in
+            do {
+                try self.fileLineReader.readLines(of: fileURL, progress: { line in
+                    switch line {
+                    case .line(let content):
```
pawelgil, line 33: this case is huge, should be split into multiple functions

pawelgil (changes requested):
Light CR performed as this module is pending a slight rework

pawelgil (approved):

PR #608, opened 2022-02-17
Implement new version for saving session's urlLocation

anna1901:
This fixes this bug:
https://trello.com/c/kqXzDa8I

### DIFF
--- a/AirCasting/CoreData/PersistenceController+Database.swift
+++ b/AirCasting/CoreData/PersistenceController+Database.swift
@@ -113,7 +113,7 @@ extension PersistenceController: SessionUpdateable {
         let context = self.editContext
         context.perform {
             do {
-                let sessionEntity: SessionEntity = try context.newOrExisting(uuid: sessionUUID)
+                let sessionEntity = try context.existingSession(uuid: sessionUUID)
                 sessionEntity.urlLocation = url
                 try context.save()
                 completion?(nil)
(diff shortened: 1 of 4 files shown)


AirCasting/CoreData/PersistenceController+Database.swift
```diff
@@ -108,6 +108,25 @@ extension PersistenceController: SessionRemovable {
     }
 }
 
+extension PersistenceController: SessionUpdateable {
+    func updateSessionUrl(_ url: String, for sessionUUID: SessionUUID, completion: ((Error?) -> Void)?) {
+        let context = self.editContext
+        context.perform {
+            do {
+                let sessionEntity: SessionEntity = try context.newOrExisting(uuid: sessionUUID)
```
pawelgil, line 116: I'm wondering if we want to use `newOrExisting` here... It's clearly an error when we want to update URL of a non-existent session and by creating new session in that case we're pushing the error further down the line.

pawelgil (approved):

AirCasting/CoreData/PersistenceController+Database.swift
anna1901: Good catch, that's definitely true. I'll change that.

AirCasting/CoreData/PersistenceController+Database.swift
```diff
@@ -108,6 +108,25 @@ extension PersistenceController: SessionRemovable {
...
+            } catch {
+                Log.error("Error adding urlLocation for session: \(error)")
+                completion?(error)
+            }
+        }
+    }
+}
+
+
+
```
jankrzempek, line 129: Do we need that many empty lines here 😅?

jankrzempek (approved):
🚀

pawelgil (approved):

PR #613, opened 2022-02-23
Get rid of the bug causing app crash.

jankrzempek:
We are now using appropriate protocols in the Resolver.

Done based on: https://github.com/hmlongco/Resolver/blob/master/Documentation/Protocols.md

https://trello.com/c/S0HRDQyM/558-mobile-dormant-fixed-edit-session-app-crashes


anna1901 (approved):
Awesome! You've done it so fast 🚀

PR #614, opened 2022-02-23
Applied missing changes from the mother branch of that task - SearchView.

jankrzempek:
Few changes were requested in SearchView from the search and follow feature.

### DIFF
--- a/AirCasting/CreateSessionViews/ChooseLocationPickerService.swift
+++ b/AirCasting/CreateSessionViews/ChooseLocationPickerService.swift
@@ -6,11 +6,11 @@ import Resolver
 import SwiftUI
 
 class ChooseLocationPickerService: PlacePickerService {
-    @InjectedObject private var tracker: LocationTracker
+    @Injected private var tracker: LocationTracker
     @Binding private var address: String
     
     init(address: Binding<String>) {
-        self._address = address
+        self._address = .init(projectedValue: address)
     }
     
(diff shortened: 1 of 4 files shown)


AirCasting/CreateSessionViews/ChooseLocationPickerService.swift
```diff
@@ -6,11 +6,11 @@ import Resolver
 import SwiftUI
 
 class ChooseLocationPickerService: PlacePickerService {
-    @InjectedObject private var tracker: LocationTracker
+    @Injected private var tracker: LocationTracker
     @Binding private var address: String
     
     init(address: Binding<String>) {
-        self._address = address
+        self._address = .init(projectedValue: address)
```
anna1901, line 13: What's the reason for using here `projectedValue` instead of `wrppedValue`? Is it a read-only variable?

anna1901 (approved):
Works fine, I just have one question

AirCasting/CreateSessionViews/ChooseLocationPickerService.swift
jankrzempek: I really have not deep dove into it. Paweł once said that it should be done this way

jankrzempek: I will let you know in the near future when I will inspect it

PR #615, opened 2022-02-28
Search complete view

anna1901:
https://trello.com/c/p8fBClRb

This is a basic implementation of search and follow feature complete view. It doesn't include necessary logic, just view based on mocked data.


AirCasting/SearchAndFollow/CompleteScreen/SearchCompleteScreenMapView.swift
```diff
@@ -0,0 +1,55 @@
+// Created by Lunar on 25/02/2022.
+//
+
+import SwiftUI
+import UIKit
+import GoogleMaps
+import GooglePlaces
```
jankrzempek, line 7: Do we need GooglePlaces here?

AirCasting/SearchAndFollow/CompleteScreen/SearchSession.swift
```diff
@@ -0,0 +1,50 @@
...
+    let longitude: Double
+    let latitude: Double
+    let streams: [SearchSessionStream]
+    
+    struct SearchSessionStream: Identifiable {
+        var id: Int
+        let sensorPackageName: String?
+        let sensorName: String
+        let measurements: [SearchSessionMeasurement]
+
```
jankrzempek, line 20: Can be deleted I think

AirCasting/SearchAndFollow/CompleteScreen/SearchSession.swift
```diff
@@ -0,0 +1,50 @@
...
+        
+    }
+    
+    struct SearchSessionMeasurement {
+        let time: Date
+        let value: Double
+    }
+    
+    
+//    #if DEBUG
```
jankrzempek, line 29: Why it is here? and the //end below

jankrzempek (changes requested):
I placed a few comments, small ones. Geat, big work 🚀
The only suggestion I have is to change the complete screen a bit. More size we have, the more information we can show - what do you think about making that mapView bigger there. I think that the button should be also on the bottom of the screen. Suggesting [screenshot], if not please skip at all 😇

AirCasting/SearchAndFollow/CompleteScreen/SearchSession.swift
anna1901: This will probably be usefull

AirCasting/SearchAndFollow/CompleteScreen/SearchSession.swift
anna1901: Cause it is a mocked session which in the future will be used only in the preview, but for now I am using it to pass session data to the view (it will be changed when we'll be able to pass session data from map view)

AirCasting/SearchAndFollow/CompleteScreen/SearchSession.swift
jankrzempek: Oh, I was talking about the empty line, sorry

AirCasting/SearchAndFollow/CompleteScreen/CompleteScreenViewModel.swift
```diff
@@ -0,0 +1,32 @@
+// Created by Lunar on 22/02/2022.
+//
+
+import Foundation
+import SwiftUI
+
+class CompleteScreenViewModel: ObservableObject {
+    let session: SearchSession
```
pawelgil, line 8: underlying session should be private and viewmodel should only expose ready to use stuff like text strings, UI-related structs etc.

AirCasting/SearchAndFollow/CompleteScreen/CompleteScreenViewModel.swift
```diff
@@ -0,0 +1,32 @@
...
+    @Published var selectedStream: Int?
+    @Published var isMapSelected: Bool = true
+    
+    init(session: SearchSession) {
+        self.session = session
+        selectedStream = session.streams.first?.id
+    }
+    
+    func stream(withID id: Int?) -> SearchSession.SearchSessionStream {
```
pawelgil, line 17: 2 things here:

1. ViewModels should not have this kind of in-out functions. We use bindings as outputs of viewmodels and functions as inputs. So viewmodel only has UI related input functions (that don't return anything) reflecting user interactions and `@Published` outputs which we bind into inside our views.

2. `SearchSessionStream` seems like a data layer object and should not be exposed in a viewmodel

AirCasting/SearchAndFollow/CompleteScreen/CompleteScreenViewModel.swift
```diff
@@ -0,0 +1,32 @@
...
+        if let streamId = id, let rightStream = session.streams.first(where: { $0.id == streamId}) {
+            return rightStream
+        } else {
```
pawelgil, line 20: we already return in `if let`, no need for `else` here

AirCasting/SearchAndFollow/CompleteScreen/SearchAndFollowChartViewModel.swift
```diff
@@ -0,0 +1,21 @@
...
+import Charts
+
+class SearchAndFollowChartViewModel: ObservableObject {
+    @Published var entries: [ChartDataEntry] = []
+    
+    init(stream: SearchSession.SearchSessionStream) {
+        generateEntries(for: stream)
+    }
+    
+    func generateEntries(for stream: SearchSession.SearchSessionStream) {
```
pawelgil, line 14: same as before - `generateEntries` is not a user interaction so it should not be a public method on a viewmodel. It also takes a data object as a parameter which the view should not know about (except for passing it from its init to a viewmodel init in SwiftUI approach)

AirCasting/SearchAndFollow/CompleteScreen/SearchCompleteScreenHeader.swift
```diff
@@ -0,0 +1,68 @@
+// Created by Lunar on 22/02/2022.
+//
+
+import SwiftUI
+
+struct SearchCompleteScreenHeader: View {
+    var session: SearchSession
```
pawelgil, line 7: this is a data layer object, should not be handled by the view layer objects

AirCasting/SearchAndFollow/CompleteScreen/SearchCompleteScreenMapView.swift
```diff
@@ -0,0 +1,53 @@
...
+        // 5. Continent
+        // 10. Cities
+        // 15. Streets
+        // 20. Buildings
+        
+        let camera = GMSCameraPosition.camera(withLatitude: latitude, longitude: longitude, zoom: 15.0)
+        let mapView = GMSMapView.map(withFrame: .zero, camera: camera)
+        
+        do {
+            if let styleURL = Bundle.main.url(forResource: "style", withExtension: "json") {
```
pawelgil, line 31: this code is mixing low level stuff with the view layer. It's not terrible and can stay IMO, but just pointing it out. Proper design would be to create a `StyleStore` protocol which provides us with styles, hide this `Bundle` related logic inside the implementation and inject it here using dependency injection.

AirCasting/SearchAndFollow/CompleteScreen/StaticSingleMeasurementView.swift
```diff
@@ -0,0 +1,56 @@
+// Created by Lunar on 28/02/2022.
+//
+
+import SwiftUI
+
+struct StaticSingleMeasurementView: View {
+    @Binding var selectedStreamId: Int?
+    var streamId: Int
+    var streamName: String
+    var value: Double
```
pawelgil, line 10: all this looks like a good viewmodel candidate

AirCasting/SearchAndFollow/CompleteScreen/StaticSingleMeasurementView.swift
```diff
@@ -0,0 +1,56 @@
...
+        Color.aircastingGray
+            .clipShape(Circle())
+            .frame(width: 5, height: 5)
+    }
+    
+    func showStreamName(_ streamName: String) -> String {
+        streamName
+            .replacingOccurrences(of: ":", with: "-")
+            .drop { $0 != "-" }
+            .replacingOccurrences(of: "-", with: "")
```
pawelgil, line 48: this also looks like it wants to be in a viewmodel. Also - don't we use this exact code all over the app? Maybe we should make an extension for it or smth?

pawelgil (changes requested):

AirCasting/SearchAndFollow/CompleteScreen/CompleteScreenViewModel.swift
anna1901: 2. This function is used only to pass the right stream to chartView. How else should this be implemented?

AirCasting/SearchAndFollow/CompleteScreen/CompleteScreenViewModel.swift
anna1901: We need to pass this session to other views, like SessionHeaderView. How should we do this without exposing sessions from ViewModel?

AirCasting/SearchAndFollow/CompleteScreen/CompleteScreenViewModel.swift
anna1901: In this case @Binding is so comfortable to use, when we want to pass `selectedStream` as binding to `StaticSingleMeasurementView` so that when user taps the stream name in that view, they would see the right stream in chartView and the right stream name would have a border. I understand it could probably be done with a closure passed to this view, like `onStreamSelection`, but weren't bindings added to SwiftUI to make this process simpler?

anna1901: What are your thoughts on that? Could we discuss in on a call today?

AirCasting/SearchAndFollow/CompleteScreen/SearchSessionStreamsDownstream.swift
```diff
@@ -0,0 +1,16 @@
...
+
+protocol SearchSessionStreamsDownstream {
+    func downloadSession(uuid: SessionUUID, completion: @escaping (Result<SearchSession, Error>) -> Void)
+}
+
+class SearchSessionStreamsDownstreamMock: SearchSessionStreamsDownstream {
+    func downloadSession(uuid: SessionUUID, completion: @escaping (Result<SearchSession, Error>) -> Void) {
+        DispatchQueue.main.asyncAfter(deadline: .now() + .seconds(5)) {
+            completion(.success(.mock))
+        }
```
jankrzempek, line 14: This part is crashing on the simulator of iPhone SE 3rd gen. On the simulator of iPhone 12 everything is working fine, as well as on the physical device. I am leaving it up to you, if it is worth to make the effort to fix it now or just after you will be working with real data

AirCasting/SearchAndFollow/CompleteScreen/Views/SearchAndFollowChartView.swift
```diff
@@ -0,0 +1,61 @@
...
+        // format line and dots
+        formatDataSet(dataSet: dataSet)
+    }
+    
+    private func formatData(data: LineChartData) {
+        let formatter = NumberFormatter()
+        formatter.maximumFractionDigits = 0
+        data.setValueFormatter(DefaultValueFormatter(formatter: formatter))
+        data.setValueFont(Fonts.muliHeadingUIFont2)
+        data.setValueTextColor(UIColor.aircastingGray)
```
jankrzempek, line 45: I think we can add this: `data.highlightEnabled = false` to disable possibility of highlighting the dots

AirCasting/SearchAndFollow/CompleteScreen/Views/SearchSession.swift
```diff
@@ -0,0 +1,46 @@
...
+        let sensorPackageName: String
+        let sensorName: String
+        let measurements: [SearchSessionMeasurement]
+    }
+    
+    struct SearchSessionMeasurement {
+        let time: Date
+        let value: Double
+    }
+
```
jankrzempek, line 25: One new line can be deleted I think

jankrzempek (approved):
Really detailed and well developed! 😎
Please consider only those comments and apply is possible

AirCasting/SearchAndFollow/CompleteScreen/Views/SearchAndFollowChartView.swift
anna1901: Done, thanks for the suggestion!

anna1901:
[SEARCH FEATURE - Complete screen - views](https://trello.com/c/p8fBClRb/534-search-feature-complete-screen-views)

PR #617, opened 2022-03-01
Crashlytics logger added

pawelgil:


AirCasting/Settings/SettingsView.swift
```diff
@@ -106,6 +106,9 @@ struct SettingsView: View {
             Section() {
                 navigateToAppConfigurationButton
                 shareLogsButton
+                Text("Crashlytics integration testing:")
+                crashButton
```
anna1901, line 110: This is added to test logging?

anna1901 (approved):
Nice!

AirCasting/Settings/SettingsView.swift
pawelgil: yes, we can force-crash the app here and see if crashlytics have the logs. Note that it can take a while for crashlytics to register a crash

jankrzempek (approved):
Awesome!

PR #627, opened 2022-03-06
Adjust tests to use Resolver

pawelgil:
## What was done?

Our tests suites were broken after we introduced a more sophisticated DI mechanism. This PR contains a fix for this.


jankrzempek (approved):
Working well, every test passed ✨
I just have one question, you have used "ASD" in the code twice. I think this is intentional as it is not affecting anything. Asking to be sure this is the way

pawelgil:
> Working well, every test passed ✨
> I just have one question, you have used "ASD" in the code twice. I think this is intentional as it is not affecting anything. Asking to be sure this is the way

This is a very good question @jankrzempek and I'll answer it in a broader way. I think you're referring to the DRY principe (do not repeat yourself) which is often misunderstood. Not all **repeated** code is really **the same**. For us to generalize repeated occurrences of the same code we need to figure out whether this piece of code is doing logically the same thing in all places. The simple test for this is to ask yourself _"if I change the way this code works here, will I also need to change this in all other places?"_. If the answer is yes it means you probably need to generalize it as it`s a repeated logic. If the answer is no than we don't really have repeated behavior, just repeated bytes of text (which we still may want to generalize, but not always).
In this particular case `"ASD"` is just a coincidental thing, I didn't want to think much of it. To make those tests clearer and emphasize non-importance of those strings I could've used a constant named `.any` or `.nevermind` which I sometimes do in unit tests. This way the reader is not bothered by the particular string (or any other data type) I'm using to test a certain behavior.

```Swift
extension String {
    static var nevermind: Self { "nevermind" }
}
```

This is one of the advanced clean-tests techniques, but I didn't refactor any test in this PR, we might think of such helpers as we move forward.

anna1901 (approved):

PR #628, opened 2022-03-10
Select device screen - scrolling improved

jankrzempek:
Improved the way the screen for selecting the device is scrolling. For now, the whole screen was able to scroll which could result in moving out of the screen the continue button when there were too many devices.

### DIFF
--- a/AirCasting/CreateSessionViews/SelectPeripheralView.swift
+++ b/AirCasting/CreateSessionViews/SelectPeripheralView.swift
@@ -51,12 +51,7 @@ struct SelectPeripheralView: View {
                     refreshButton
                         .frame(alignment: .trailing)
                 }
-                
-                if selection != nil {
-                    connectButton.disabled(false)
-                } else {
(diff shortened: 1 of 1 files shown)


anna1901 (approved):
Nice one👏🏻

AirCasting/CreateSessionViews/SelectPeripheralView.swift
```diff
@@ -43,37 +42,34 @@ struct SelectPeripheralView: View {
...
+                    refreshButton
+                        .frame(alignment: .trailing)
                 }
-                .onDisappear {
-                    bluetoothManager.centralManager.stopScan()
+                
+                if selection != nil {
+                    connectButton.disabled(false)
+                } else {
+                    connectButton.disabled(true)
```
pawelgil, line 58: one liner instead of if/else: ```connectButton.disabled(selection == nil)```

pawelgil (approved):

PR #629, opened 2022-03-10
PersistentController - fix applied.

jankrzempek:
Improved edit context which was the issuing provider. Made additional changes that were needed after the edit context fix was applied.

https://trello.com/c/jWnrXXLo/576-1504-mobile-active-crash
https://trello.com/c/5OaCc4sW/567-140-persistencecontrollerswift-line-118-crash

### DIFF
--- a/AirCasting/CoreData/MeasurementStreamStorage.swift
+++ b/AirCasting/CoreData/MeasurementStreamStorage.swift
@@ -38,19 +38,17 @@ extension HiddenCoreDataMeasurementStreamStorage {
 }
 
 final class CoreDataMeasurementStreamStorage: MeasurementStreamStorage {
-
     @Injected private var persistenceController: PersistenceController
     @Injected private var updateSessionParamsService: UpdateSessionParamsService
-    private lazy var context: NSManagedObjectContext = persistenceController.editContext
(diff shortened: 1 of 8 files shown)


AirCasting/CoreData/MeasurementStreamStorage.swift
```diff
@@ -135,7 +133,7 @@ final class HiddenCoreDataMeasurementStreamStorage: MeasurementStreamStorageCont
     func addMeasurement(_ measurement: Measurement, toStreamWithID id: MeasurementStreamLocalID) throws {
         let stream = try context.existingObject(with: id.id) as! MeasurementStreamEntity
 
-        let newMeasurement = MeasurementEntity(context: context)
+        let newMeasurement = NSEntityDescription.insertNewObject(forEntityName: "MeasurementEntity", into: context) as! MeasurementEntity
```
anna1901, line 136: Why did that need to be changed?

pawelgil:
@jankrzempek is this a sable change? I think we should put this on hold for now and see if crashes still occur after #631 and #630

AirCasting/CoreData/MeasurementStreamStorage.swift
jankrzempek: Paweł said to me that this is the correct known way of inserting and object to core data, that's the main reason for that 😄

PR #630, opened 2022-03-11
Fix editing

anna1901:
## Description
This PR fixed bug with duplicated streams after editing.
Turns out there were two underlying issues:
* When the user entered the editing screen, the app performed sync with backend just for this one session to download its most recent data from backend (the goal was to have the newest version of the name and tags). However, while saving this session data, the app was also saving the streams from backend and streams got duplicated because they had different id locally (0) and in backend database (some number assigned on backend). Now, we are not saving all of the session data, just its name and tags, and this eliminated the stream duplication at this stage of the edit, but in the future, the underlying problem should also be fixed, i.e. we should save stream id from backend locally.
* When we were sending the updated name ang tags to backend, backend bumped session's version by 1. We didn't save this new version number locally. Then, when the app performed sync, there was discrepancy between the local version number for the session and the version number on backend, and because of that, the session was downloaded again, and the streams were duplicated (because of the id mismatch). Now, we are saving the new version number after updating session data to backend, and this eliminated this issue (but still, we should really get the right ids for streams).

## Code changes
* Rewriting logic in editViewModel
* Adding handling of response in sessionUpdateService
* Removing single session synchroniser
* Changing back insertOrUpdateSessions to insertSessions

https://trello.com/c/NgCUPYg4

### DIFF
--- a/AirCasting/CoreData/NSManagedObjectContext+Utils.swift
+++ b/AirCasting/CoreData/NSManagedObjectContext+Utils.swift
@@ -40,38 +40,38 @@ extension NSManagedObjectContext {
     }
     
     
-    #warning("This seems odd and is also working odd. I don't think this is correct implementation")
-    // Explaination:
-    // I'm not sure why is that happening, but the fetch inside this function is returning multiple streams
-    // not associated with a single session. It causes the calling code to sometimes modify an incorrect
-    // stream object. I found this when implementing the url saving feature after uploads
-    // (https://github.com/HabitatMap/AirCastingiOS/pull/578) - I used the `Databse`s
-    // `func insertOrUpdateSessions(_ sessions: [Database.Session], completion: ((Error?) -> Void)?)`
-    // to achieve session URL update and it started acting weird - adding streams to existing dB sessions
-    // etc.. I think the root of the problem is that session streams don't have a unique identifiers
-    // and we decided to use a NSManagedObjectID instead. Somehow it's not working right, but when I added
-    // additional session-id constraint it started to behave somewhat properly. But this is a blind patch
-    // and should be investigated and resolved.
-    // Issue: https://github.com/HabitatMap/AirCastingiOS/issues/579
-    
-    // Checks if stream exists, if not, creates a new one
-    func newOrExisting<T: NSManagedObject>(streamID: MeasurementStreamID, for session: SessionUUID? = nil) throws -> T  {
-        let className = NSStringFromClass(T.classForCoder())
-        let fetchRequest = NSFetchRequest<T>(entityName: className)
-        fetchRequest.predicate = NSPredicate(format: "id == \(streamID)")
-        if let session = session {
-            fetchRequest.predicate = NSCompoundPredicate(type: .and,
-                                                         subpredicates: [fetchRequest.predicate!,
-                                                                         NSPredicate(format: "session.uuid == %@", session.rawValue)])
-        }
-        
-        let results = try self.fetch(fetchRequest)
-        if let existing  = results.first { return existing }
-        
-        let new: T = T(context: self)
-        new.setValue(streamID, forKey: "id")
-        return new
-    }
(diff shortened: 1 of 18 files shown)


anna1901:
[1.5.0(4) > mobile dormant > edit session: after editing a session name, the streams are duplicated](https://trello.com/c/NgCUPYg4/575-1504-mobile-dormant-edit-session-after-editing-a-session-name-the-streams-are-duplicated)

AirCasting/SingleSessionSynchronizer/SessionUpdateService.swift
```diff
@@ -5,19 +5,26 @@ import Foundation
...
 
 protocol SessionUpdateService {
-    func updateSession(session: SessionEntity, completion: @escaping () -> Void)
+    func updateSession(session: SessionEntity, completion: @escaping (Result<UpdateSessionReturnedData, Error>) -> Void)
 }
 
-class DefaultSessionUpdateService: SessionUpdateService {
+struct UpdateSessionReturnedData: Codable {
+    let version: Int
+}
```
jankrzempek, line 13: Maybe we could make a new file for this as for other structs we do?

jankrzempek (approved):
Perfect 🚀

pawelgil:
Can we be a bit more verbose in PR titles? Something in the realm of "Fix duplicated streams problem on session edit" or smth

AirCasting/CoreData/NSManagedObjectContext+Utils.swift
```diff
@@ -39,40 +39,6 @@ extension NSManagedObjectContext {
...
-        }
-        
-        let results = try self.fetch(fetchRequest)
-        if let existing  = results.first { return existing }
-        
-        let new: T = T(context: self)
-        new.setValue(streamID, forKey: "id")
-        return new
-    }
-
```
pawelgil, line 75: 😍

AirCasting/CoreData/PersistenceController+Database.swift
```diff
@@ -25,12 +25,11 @@ extension PersistenceController: SessionsFetchable {
 }
 
 extension PersistenceController: SessionInsertable {
-    func insertOrUpdateSessions(_ sessions: [Database.Session], completion: ((Error?) -> Void)?) {
+    func insertSessions(_ sessions: [Database.Session], completion: ((Error?) -> Void)?) {
```
pawelgil, line 28: 😍

AirCasting/SessionViews/EditViewModel.swift
```diff
@@ -10,23 +10,23 @@ protocol EditViewModel: ObservableObject {
...
 }
 
 class EditSessionViewModel: EditViewModel {
-    
     @Published var isSessionDownloaded = false
     @Published var sessionName = ""
     @Published var sessionTags = ""
     @Published var shouldShowError = false
-    @Published var didSave = false
+    @Published var shouldDismiss = false
```
pawelgil, line 26: 😍

AirCasting/SingleSessionSynchronizer/SessionUpdateService.swift
pawelgil: I don't think that's necessary, we can move this struct above the SessionUpdateService protocol definition because we're using it there so it would be easier to read this file top-down

AirCasting/SessionsSynchronization/Controller/SessionSynchronizer.swift
```diff
@@ -16,7 +16,7 @@ struct SessionSynchronizationOptions: OptionSet {
 }
 
 /// Defines the interface for objects that provide session list synchronization to the app
-protocol SessionSynchronizer: SingleSessionSynchronizer {
```
pawelgil, line 19: 😍

pawelgil (approved):
So many good changes, great stuff!

PR #631, opened 2022-03-12
Fix CoreData mishandle in averaging

pawelgil:


jankrzempek (approved):
✨

AirCasting/CoreData/MeasurementStreamStorage.swift
```diff
@@ -195,8 +195,10 @@ final class HiddenCoreDataMeasurementStreamStorage: MeasurementStreamStorageCont
         return session
     }
 
-    func updateMeasurements(stream: MeasurementStreamEntity, newMeasurements: NSOrderedSet) throws {
-        stream.measurements = newMeasurements
+    func removeAllMeasurements(in stream: MeasurementStreamEntity, except measurementsToLeave: [MeasurementEntity]) {
```
anna1901, line 198: Great, thank you for changing that! This function name and logic was very misleading

AirCasting/Utils/AveragingService.swift
```diff
@@ -171,6 +173,7 @@ final class AveragingService: NSObject {
     }
     
     private func startPeriodicAveraging(uuid: SessionUUID, window: AveragingWindow) {
+        Log.info("Starting periodic averaging with window of \(window.rawValue)s for \(uuid)")
         let timer = Timer.publish(every: TimeInterval(window.rawValue), on: .main, in: .common)
```
anna1901, line 177: Is it possible and does it make sense to schedule the timer on the background thread?

anna1901 (approved):
Nice!! 🚀

AirCasting/Utils/AveragingService.swift
pawelgil: the `on:` parameter here is actually a `RunLoop`, not a `DispatchQueue`. And this is very important. Timers are scheduled on runloops which means that if we want to have a background threaded timer set up, we would need to start a runloop manually on that thread and manage the whole thing by ourselves (stopping the runloop after timer invalidation perhaps. We also couldn't use GCD since it doesn't guarantee blocks are being called on the same thread and runloops are attached to threads etc.). This is why in a vast majority of cases timers should be just scheduled on a main runloop and if we want the handler to operate on a background queue we should just dispatch there. From what I see this is the case here since periodic update func is doing:
```Swift
measurementStreamStorage.accessStorage { ... }
```
which guarantees background execution of a block

I also noticed now that timers in averaging are not being properly removed from the runloop. Will fix this issue in a separate PR.

PR #632, opened 2022-03-14
Standalone - update card UI

SzymonGorowski:
The text has been adjusted and centered. Space has been added before each card.


anna1901 (approved):
Great 🚀

PR #635, opened 2022-03-15
Add missing background task end marking code when invalidated by the system

pawelgil:
## What was done?

The code that handles background tasks for session sync might have caused some crashes in background


jankrzempek (approved):
Hope it will help with all of those bugs 🙌

anna1901 (approved):

PR #638, opened 2022-03-16
Add web view for displaying privacy policy

anna1901:
https://trello.com/c/FNOc7hDZ

### DIFF
--- a/AirCasting/Extensions/WKWebView.swift
+++ b/AirCasting/Extensions/WKWebView.swift
@@ -1,0 +1,15 @@
+// Created by Lunar on 03/03/2022.
+//
+
+import Foundation
+import WebKit
+
+extension WKWebView {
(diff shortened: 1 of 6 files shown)


AirCasting/Extensions/WKWebView.swift
```diff
@@ -0,0 +1,15 @@
...
+//
+
+import Foundation
+import WebKit
+
+extension WKWebView {
+    func load(_ url: URL?) {
+        guard let url = url else {
+            return
+        }
```
jankrzempek, line 11: I think this can be one liner here 😄

jankrzempek (approved):
SUGGESTION: Can we please add a spinner when the WebView is loading, sometimes (in my case) it is taking a long time. The whole PR - awesome 😎

PR #639, opened 2022-03-16
Change default location after tapping the location button 

SzymonGorowski:
https://trello.com/c/2bfrWXgC/551-fix-map-centering-button-to-go-immediately-to-the-right-location


jankrzempek (changes requested):
I need to say that this is not going to work in some cases :( . You are using this:
`parent.tracker.locationManager.location!.coordinate.latitude : parent.pathPoints.last?.location.latitude ??parent.tracker.locationManager.location!.coordinate.latitude` 
so whenever you don't have the coordinates from locationManager it will try to use (the last code line) - so the same manager although it was considered as the one that does not have the value at first. In the simulator case it will crash and on the fixed sessions where we do not need to request location (we are only suggesting that it is good to have it) it will crash too. In case of me not being clear please let me know :)

jankrzempek (commented):
Good way I think , but still not working as expected on the fixed session.
Please avoid force unwrapping as sometimes we will have the case when the user is in the process of creating a new session but his location will be denied which then can lead to a crash.
A great way of adding `parent.isCreatingSessionScreen` I think, and when adding that please think if we still do need liveModeOn variable.

I am going to paste a "kind of" solution that I think is working. I shortened the Double value as it was not working for the initial (long one) value. Maybe that was just me so please confirm.

```
func centerMap(for mapView: GMSMapView) {
            
            let lat = parent.isCreatingSessionScreen ?
            parent.tracker.locationManager.location?.coordinate.latitude ?? 37.35 :
            parent.pathPoints.last?.location.latitude ?? 37.35
            
            let long = parent.isCreatingSessionScreen ?
            parent.tracker.locationManager.location?.coordinate.longitude ?? -122.05 :
            parent.pathPoints.last?.location.longitude ?? -122.05
            
            let shorted_lat = String(format:"%.5f", lat)
            let shorted_long = String(format:"%.5f", long)
            
            let camera = GMSCameraPosition.camera(withLatitude: Double(shorted_lat)!, longitude: Double(shorted_long)!, zoom: 16)
            mapView.animate(to: camera)
        }
```

anna1901:
@jankrzempek I think liveModeOn is still necessary. I think it was added to differentiate between active and dormant sessions. If the session is active, we want to center the camera on the most recent location (`parent.tracker.locationManager.location?.coordinate`) but if it is dormant, we want to center camera on the last path point. However, when we are creating a session, we want to center the camera on the current location (or on the location provided in the location picker). Isn't that right?

anna1901:
Actually, I think it would be best to create a separate map class for creatingSessionView and avoid those problems this way. The map in this view doesn't need the whole pathPoints logic, so it might be a smaller view specifically made for the session creation purposes. What do you think @jankrzempek and @SzymonGorowski ?

AirCasting/Map/CreatingSessionMapView.swift
```diff
@@ -0,0 +1,80 @@
...
+import Resolver
+
+struct CreatingSessionMapView: UIViewRepresentable {
+    
+    typealias UIViewType = GMSMapView
+    @InjectedObject private var tracker: LocationTracker
+    var isMyLocationEnabled = false
+    
+    init(isMyLocationEnabled: Bool = false) {
+        GMSServices.provideAPIKey(GOOGLE_MAP_KEY)
```
anna1901, line 17: I know it was done this way in those PRs I told you about, but actually, we should provide this API key once for the whole app. So I think this could be moved to AppBootstrap class, what do you think @pawelgil ?

anna1901 (commented):
Great work 🚀 Now this map view is so much cleaner

AirCasting/Map/CreatingSessionMapView.swift
anna1901: @jankrzempek , could you move this and delete the duplication from GoogleMapView?

jankrzempek (approved):
Great map, great work 😎

anna1901 (approved):

PR #641, opened 2022-03-17
Fix crashes for background context

anna1901:
This PR fixes the app crash that was happening when user entered background after creating new mobile AB session before the app received first measurements.

### DIFF
--- a/AirCasting/MobilePeripheralSessionManager.swift
+++ b/AirCasting/MobilePeripheralSessionManager.swift
@@ -21,12 +21,10 @@ class MobilePeripheralSessionManager {
         measurementStreamStorage.accessStorage { [weak self] storage in
             do {
                 try storage.createSession(session)
-                DispatchQueue.main.async {
-                    if !session.locationless {
-                        self?.locationProvider.requestLocation()
-                    }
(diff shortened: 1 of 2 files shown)


AirCasting/MobilePeripheralSessionManager.swift
```diff
@@ -147,16 +145,18 @@ class MobilePeripheralSessionManager {
             do {
                 let existingStreamID = try storage.existingMeasurementStream(sessionUUID, name: stream.sensorName)
                 guard let id = existingStreamID else {
-                    return try self.createSessionStream(stream, sessionUUID)
+                    let streamId = try self.createSessionStream(stream, sessionUUID, storage: storage)
+                    try storage.addMeasurementValue(stream.measuredValue, at: location, toStreamWithID: streamId)
```
pawelgil, line 149: no biggie, but since we're doing the `try storage.addMeasurementValue(stream.measuredValue, at: location, toStreamWithID: id)` anyway if guard is not hit, we can just change guard to `if` and only do the session stream creation inside (the measurement adding line will be executed in both cases then)

pawelgil (approved):

jankrzempek (approved):
🥳

PR #642, opened 2022-03-17
Disable measurement update when app is on background.

jankrzempek:


anna1901 (commented):
Nicely done! It works as expected and this download it not triggered in the background. However, now when a user returns to the app, the following session don't have most recent measurements until the scheduled measurements download fires again. I think it would be a good idea to add a call to updateAllSessionsMeasurements when the app returns from background

PR #644, opened 2022-03-17
Get only followed sessions in DownloadMeasurementsService

anna1901:
I think it will improve performance a lot. Currently we are downloading measurements for all of the fixed sessions and for each of those sessions we are triggering call to api and then updating session params in core data. I think it is probably not necessary to do it for sessions which are not followed.

## Code changes
* limit periodic measurements download to followed sessions only
* download new measurements for followed sessions on card expand
* remove saving cardToggle state for fixed not followed sessions

Additional improvements:
* remove 2 seconds wait before calling completion in session downloader (it was causing the long waiting for measurements download and I think most likely it was not necessary - should be tested to check if it didn't break anything important)

### DIFF
--- a/AirCasting/SessionViews/ABMeasurementsView.swift
+++ b/AirCasting/SessionViews/ABMeasurementsView.swift
@@ -97,8 +97,8 @@ struct _ABMeasurementsView: View {
                 }
             }
         }
-        .onChange(of: isCollapsed, perform: { new in
-            if isCollapsed == false && (!hasAnyMeasurements || (session.isFixed && !session.isFollowed)) {
+        .onChange(of: isCollapsed, perform: { _ in
+            if isCollapsed == false && (!hasAnyMeasurements || session.isFixedNotFollowed) {
                 measurementsViewModel.syncMeasurements()
             }
         })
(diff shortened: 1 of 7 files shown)


AirCasting/SessionViews/ABMeasurementsView.swift
```diff
@@ -96,7 +98,7 @@ struct _ABMeasurementsView: View {
             }
         }
         .onChange(of: isCollapsed, perform: { new in
```
jankrzempek, line 100: We can get rid of the new keyword and replace it with underscore I think

AirCasting/SessionViews/SessionCardView.swift
```diff
@@ -132,7 +132,9 @@ private extension SessionCardView {
             action: {
                 withAnimation {
                     isCollapsed.toggle()
-                    uiState.toggleCardExpanded(sessionUUID: session.uuid)
+                    if !(session.isFixed && !session.isFollowed) {
```
jankrzempek, line 135: Here. As we have this again and possibly somewhere else too in the future, could we add some extension and then maybe use session.FixedNotFollowed. That is just idea, if now please skip :)

jankrzempek (approved):
Awesome! Wonderfully! Just added to comments for consideration. Tested and seems to be working as intended

AirCasting/SessionViews/SessionCardView.swift
pawelgil: `unfollowedFixed` perhaps?

pawelgil (approved):

PR #646, opened 2022-03-22
Fix creating microphone session

anna1901:


AirCasting/CreateSessionViews/SelectDeviceView.swift
```diff
@@ -108,6 +108,7 @@ struct SelectDeviceView: View {
             micLabels
         })
         .buttonStyle(WhiteSelectingButtonStyle(isSelected: selected == 2))
+        .disabled(microphoneManager.isRecording)
```
jankrzempek, line 111: Maybe we could add and alert that can inform user about the thing that he can only make only one session and that's why it is grayed out?

jankrzempek (approved):
🥳🥳🥳

AirCasting/MicrophoneSession/MicrophoneManager.swift
```diff
@@ -52,22 +52,22 @@ final class MicrophoneManager: NSObject, ObservableObject {
         try audioSession.setActive(true)
         recorder.isMeteringEnabled = true
         
-        createMeasurementStream(for: session) { result in
+        createMeasurementStream(for: session) { [unowned self] result in
             switch result {
-            case .success(let id):
-                self.measurementStreamLocalID = id
+            case .success():
```
pawelgil, line 57: you don`t need `()` here

pawelgil (approved):
I think there might be some unsafe reads/writes from random threads to public properties in microphone manger, but this is something that was also present before this PR so I'll not include any remarks (I'm also going to refactor all this soon, so no point in fixing all that)

AirCasting/MicrophoneSession/MicrophoneManager.swift
```diff
@@ -52,22 +52,22 @@ final class MicrophoneManager: NSObject, ObservableObject {
         try audioSession.setActive(true)
         recorder.isMeteringEnabled = true
         
-        createMeasurementStream(for: session) { result in
+        createMeasurementStream(for: session) { [unowned self] result in
```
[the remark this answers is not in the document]
anna1901: Thanks, now I will remember about this for sure 😄

PR #654, opened 2022-03-25
Sd sync fixed sessions upload refactor

anna1901:
Previously we were reading the whole file content to memory before sending streams to backend. Now we are introducing a buffer mechanism and sending measurements in 50 elements chunks.


jankrzempek (approved):
🥳🥳🥳

PR #656, opened 2022-03-29
SearchFollow bottom cards

jankrzempek:
‼️ Files which are presented first are duplicated from develop, please skip
Start a review from: MapSessionMarker.swift ‼️

Every card corresponds to its dot on the map. Whenever the card is clicked, the modal screen with details pops up. Whenever the dot is clicked, it gets bigger and the coresponding card is marked with blue border.

https://trello.com/c/yTJPyndz/535-search-feature-search-map-only-bottom-cards


AirCasting/SearchAndFollow/SearchMap/SearchMapView.swift
```diff
@@ -52,49 +57,88 @@ private extension SearchMapView {
...
+                HStack(alignment: .bottom, spacing: 12) {
+                    ForEach(viewModel.sessionsList, id: \.id) { session in
+                        BottomCardView(cardPointer: $viewModel.cardPointerID,
+                                       dataModel: .init(id: session.id,
+                                                        title: session.title,
+                                                        startTime: session.startTime,
+                                                        endTime: session.endTime,
+                                                        latitude: session.location.latitude,
+                                                        longitude: session.location.longitude))
+                        .border((viewModel.cardPointerID == session.id ? .blue : .clear), width: 1)
```
anna1901, line 128: Can we give the border an aircasting blue color?

anna1901, line 128: Can we give the border an aircasting blue color?

AirCasting/SearchAndFollow/SearchMap/SearchMapView.swift
```diff
@@ -52,49 +57,88 @@ private extension SearchMapView {
...
+                        
+                    }
+                }
+                .onReceive(viewModel.$cardPointerID) { _ in
+                    DispatchQueue.main.asyncAfter(deadline: .now() + .milliseconds(200)) {
```
anna1901, line 133: Hmmm I'm not sure this kind of dispatch workaround is a safe way to solve this problem. I can see that without it this scrolling gets triggered before the cardPointerID is set by the map. Maybe this could be solved by adding `onMarkerTapped` function to the map, similar to the `onPositionChange` one and we could inform the viewModel that the marker was tapped and then the viewModel could set its pointerID value itself? I didn't try if it would work, but maybe it would be a cleaner way of handling that interaction.

anna1901, line 133: Hmmm I'm not sure this kind of dispatch workaround is a safe way to solve this problem. I can see that without it this scrolling get's triggered before the cardPointerID is set by the map. Maybe this could be solved by adding `onMarkerTapped` function to the map, similar to the `onPositionChange` one and we could inform the viewModel that the marker was tapped and then the viewModel could set it's pointerID value itself? I didn't try if it would work, but maybe it would be a cleaner way of handling that interaction.

anna1901 (commented):
Great job Janek!!! It looks awesome and works awesome as well. 👏🏻 I added a couple of comments with some ideas for you to consider

AirCasting/SearchAndFollow/SearchMap/SearchMapView.swift
pawelgil: I agree, using `DispatchQueue.main.asyncAfter(deadline:)` should be a last resort and should be used only when there is absolutely no other way (and in most cases there is)

pawelgil (changes requested):
those `UIScreen` references need to go :)

AirCasting/SearchAndFollow/SearchAndFollowMap/SearchAndFollowMap.swift
```diff
@@ -51,16 +54,18 @@ struct SearchAndFollowMap: UIViewRepresentable {
     }
     
     func updateUIView(_ uiView: GMSMapView, context: Context) {
-        if !showRedoButton {
+        if pointerID != context.coordinator.pointerIdHolder {
```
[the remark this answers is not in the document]
jankrzempek: Sure, will change!

AirCasting/SearchAndFollow/SearchMap/SearchMapView.swift
```diff
@@ -52,49 +57,88 @@ private extension SearchMapView {
...
+                                                       .white.opacity(0.9),
+                                                       .white]),
+                           startPoint: .bottom,
+                           endPoint: .top)
+            .frame(width: UIScreen.main.bounds.width, height: UIScreen.main.bounds.height / 4.5, alignment: .top)
+        }
+    }
+    
+    var cardsTitle: some View {
+        HStack(alignment: .bottom) {
```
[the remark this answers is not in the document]
jankrzempek: Thaaanks, it is not 😄

AirCasting/SearchAndFollow/SearchMap/SearchMapView.swift
jankrzempek: Suree 🥳

AirCasting/SearchAndFollow/SearchMap/SearchMapView.swift
```diff
@@ -52,49 +57,88 @@ private extension SearchMapView {
...
-            .opacity(viewModel.searchAgainButton ? 1.0 : 0.0)
+        } label: {
+            Text("\(Strings.SearchMapView.redoText) \(Image(systemName: "goforward"))")
+                .font(Fonts.boldHeading3)
+                .lineLimit(1)
+                .scaledToFill()
+                .multilineTextAlignment(.leading)
+        }
+        .foregroundColor(.white)
+        .frame(width: UIScreen.main.bounds.width / 2.2, height: 8, alignment: .center)
```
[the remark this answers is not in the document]
jankrzempek: Sure, I thought that this time is okey to do so, but I will change

AirCasting/SearchAndFollow/SearchMap/SearchMapViewModel.swift
```diff
@@ -16,6 +16,7 @@ class SearchMapViewModel: ObservableObject {
     @Published var showLoadingIndicator: Bool = false
     @Published var alert: AlertInfo?
     @Published var shouldDismissView: Bool = false
+    @Published var cardPointerID: Int = -1
```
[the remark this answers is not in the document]
jankrzempek: Here it needs to stay I think, becouse I need to pass it somehow to the map. I will do my best to reduce it as much as possible

AirCasting/SearchAndFollow/MapSessionsDownloader/APIStruct/MeasurementType.swift
```diff
@@ -0,0 +1,18 @@
...
+enum MeasurementType: Codable {
+    case particulateMatter
+    case ozone
+}
+
+extension MeasurementType {
+    var name: String {
+        switch self {
+        case .particulateMatter: return "Particulate Matter"
+        case .ozone: return ""
```
anna1901, line 15: This file is different than on develop branch, so I think it would be better to sort out this problem with files that shouldn't be included in this PR, because there might be some unwanted changes

AirCasting/Extensions/UIImage.swift
```diff
@@ -36,3 +36,33 @@ extension UIImage {
...
+        
+        // Compute the new image size that preserves aspect ratio
+        let scaledImageSize = CGSize(
+            width: size.width * scaleFactor,
+            height: size.height * scaleFactor
+        )
+
+        // Draw and return the resized UIImage
+        let renderer = UIGraphicsImageRenderer(
+            size: scaledImageSize
```
pawelgil, line 56: formatting is odd in this function

pawelgil (approved):

AirCasting/SearchAndFollow/SearchMap/SearchMapView.swift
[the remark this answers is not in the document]
pawelgil: calculating frames based on `UIScreen` in any case is smelly IMO. Using local coordinates makes our code way more future proof

AirCasting/SearchAndFollow/SearchMap/SearchMapView.swift
```diff
@@ -16,21 +16,39 @@ struct SearchMapView: View {
...
+                            cards
+                                .frame(width: reader.size.width, height: reader.size.height / 7, alignment: .leading)
+                        }
+                        .padding(.horizontal, 5)
+                        .padding(.bottom, 5)
+                    }
+                }
+                VStack(alignment: .center, content: {
+                    addressTextField
+                    measurementTypeText
```
anna1901, line 36: The measurement type text is still "particular matter" instead of "particulate matter". Could you change it in this PR before we forget to do it?

AirCasting/SearchAndFollow/SearchAndFollowMap/SearchAndFollowMap.swift
```diff
@@ -51,16 +60,18 @@ struct SearchAndFollowMap: UIViewRepresentable {
     }
     
     func updateUIView(_ uiView: GMSMapView, context: Context) {
-        if !showRedoButton {
+        if selectedPointerID != context.coordinator.pointerIdHolder {
+            context.coordinator.pointerIdHolder = selectedPointerID
+            placeDots(uiView, context: context)
+        }
+        if !showSearchAgainButton {
             placeDots(uiView, context: context)
```
anna1901, line 68: So this line is triggered whenever the user selects a new marker, so placeDots is called twice. I think we can avoid this by putting this line in an else if

AirCasting/SearchAndFollow/SearchAndFollowMap/SearchAndFollowMap.swift
```diff
@@ -111,13 +123,15 @@ struct SearchAndFollowMap: UIViewRepresentable {
         }
         
         func mapView(_ mapView: GMSMapView, didTap marker: GMSMarker) -> Bool {
-            // Will be implemented in next task
+            guard let newPointerID = marker.userData as? Int else {
+                Log.error("To continue the program ensure yourself why newPointerID is not and INT!")
```
anna1901, line 127: What do you think about changing the message to "Unexpectedly found value other than Int in marker userData"?

AirCasting/SearchAndFollow/SearchMap/BottomCard.swift
```diff
@@ -0,0 +1,67 @@
...
+        _viewModel = .init(wrappedValue: .init(id: id, title: title, startTime: startTime, endTime: endTime, latitude: latitude, longitude: longitude))
+    }
+    
+    var body: some View {
+        sessionCard
+    }
+    
+    var sessionCard: some View {
+        Button {
+            viewModel.setIsModalScreenPresented()
```
anna1901, line 21: As I understand, the function should be called more something like `sessionCardTapped` as the view tells the viewModel what happed rather than what it should do. Did I get it right or is it the other way around @pawelgil ?

AirCasting/SearchAndFollow/SearchMap/BottomCard.swift
```diff
@@ -0,0 +1,67 @@
...
+                    .foregroundColor(.aircastingGray)
+                    .minimumScaleFactor(0.1)
+                    .scaledToFit()
+            }
+        }
+        .sheet(isPresented: .init(get: {
+            viewModel.getIsModalScreenPresented()
+        }, set: { _ in
+            viewModel.setIsModalScreenPresented()
+        }), content: { viewModel.initCompleteScreen() })
```
anna1901, line 42: The `x` mark on the complete screen stopped working with this implementation, so that needs to be fixed

AirCasting/SearchAndFollow/SearchMap/BottomCardViewModel.swift
```diff
@@ -0,0 +1,47 @@
...
+        return d
+    }
+    
+    func initCompleteScreen() -> CompleteScreen {
+        CompleteScreen(session: .init(uuid: .init(rawValue: "\(dataModel.id)") ?? .init(),
+                                      name: dataModel.title,
+                                      startTime: startTimeAsDate(),
+                                      endTime: endTimeAsDate(),
+                                      longitude: dataModel.longitude,
+                                      latitude: dataModel.latitude))
```
anna1901, line 45: Probably we can add some dismiss function here to fix the `X` mark. Or we can delete the `X` mark, cause it doesn't really look so good. I included it in the complete screen cause it was present in the designs

AirCasting/SearchAndFollow/SearchMap/SearchMapView.swift
```diff
@@ -52,38 +70,33 @@ private extension SearchMapView {
...
+                                   showSearchAgainButton: $viewModel.searchAgainButton,
+                                   sessions: $viewModel.sessionsList,
+                                   selectedPointerID: $viewModel.cardPointerID)
+                .onPositionChange(action: { geoSquare in
                     viewModel.mapPositionsChanged(geoSquare: geoSquare)
                 })
-                    .padding(.top, 50)
-                    .ignoresSafeArea(.all, edges: [.bottom])
+                .onMarkerChange(action: { pointer in
+                    viewModel.markerPositionChanged(using: pointer)
```
anna1901, line 96: I think the position of the marker didn't change, but a new marker got selected 🙂

anna1901 (changes requested):
Awesome job again! There are some small changes that need to be done to remove some small bugs. When the redo button gets tapped we need to set the cardPointerID to nil, because otherwise the previously selected pointer is still selected and when it gets tapped the scroll animation doesn't get triggered.

AirCasting/SearchAndFollow/SearchMap/SearchMapView.swift
jankrzempek: For now here it's just a mock, but I will change not to mess up anything. Thanks!

AirCasting/SearchAndFollow/SearchAndFollowMap/SearchAndFollowMap.swift
jankrzempek: That was it, totally amazing. Thx a lot!

AirCasting/SearchAndFollow/SearchMap/BottomCard.swift
jankrzempek: You right, my mistake

AirCasting/SearchAndFollow/SearchMap/BottomCardViewModel.swift
jankrzempek: I think it does look ok here. I do only use swipe-down gestures but who knows about others. I will fix it, and if we could leave it, it would be great

AirCasting/SearchAndFollow/SearchMap/SearchMapView.swift
```diff
@@ -52,38 +68,33 @@ private extension SearchMapView {
...
     var map: some View {
-        GeometryReader { geometry in
+        GeometryReader { reader in
             ZStack(alignment: .top) {
                 SearchAndFollowMap(startingPoint: viewModel.passedLocationAddress,
-                                   showRedoButton: $viewModel.searchAgainButton,
-                                   sessions: $viewModel.sessionsList).onPositionChange(action: { geoSquare in
+                                   showSearchAgainButton: $viewModel.searchAgainButton,
+                                   sessions: $viewModel.sessionsList,
+                                   selectedPointerID: $viewModel.cardPointerID)
```
anna1901, line 89: It's not causing any issues, but I think that it is not necessary to pass the selectedPointerID to the map. I think it's a map that needs to inform the viewModel which marker was tapped, and it can do it with `onMarkerChange` function. Then the map itself can decide which marker should be larger, and it could be triggered in the `didTap` function (as I have sent you on slack). As we want to move forward with the feature, I think you could merge this PR anyway, but we could change this in another PR, to make code cleaner and remove unnecessary complexity.

anna1901 (approved):

PR #664, opened 2022-04-11
Adding notes in the current user location

SzymonGorowski:
https://trello.com/c/Ni2Q9sdl/583-notes-notes-are-not-always-added-to-the-current-location

Made some changes which should fix the problem with the notes


AirCasting/Notes/AddNote/AddNoteViewModel.swift
```diff
@@ -4,18 +4,26 @@ import Foundation
...
     }
     
     func continueTapped() {
-        notesHandler.addNote(noteText: noteText)
+        if let latitude = locationTracker.locationManager.location?.coordinate.latitude,
+           let longitude = locationTracker.locationManager.location?.coordinate.longitude,
+           trackLocation {
+            notesHandler.addNote(noteText: noteText, latitude: latitude, longitude: longitude)
+        }
+        notesHandler.addNote(noteText: noteText, latitude: 20.0, longitude: 20.0)
```
jankrzempek, line 26: Here when you add one note you will continue to next step, so whenever the user add a new note - he wil add a proper one as well as the fake one. I think we should avoid that, so maybe we could use a guard here? If you prefer `if let` then please combine it with `return` and use `exitRoute()` in both cases. Still I would like to say that seems to be working really nice 😊 

Example:

`guard let latitude = locationTracker.locationManager.location?.coordinate.latitude,
              let longitude = locationTracker.locationManager.location?.coordinate.longitude,
              trackLocation else {
            notesHandler.addNote(noteText: noteText, latitude: 20.0, longitude: 20.0)
            exitRoute()
            return
        }
        notesHandler.addNote(noteText: noteText, latitude: latitude, longitude: longitude)
        exitRoute() `

jankrzempek, line 26: Here when you add one note you will continue to next step, so whenever the user add a new note - he wil add a proper one as well as the fake one. I think we should avoid that, so maybe we could use a guard here? If you prefer `if let` then please combine it with `return` and use `exitRoute()` in both cases. Still I would like to say that seems to be working really nice 😊 

Example:

`guard let latitude = locationTracker.locationManager.location?.coordinate.latitude,
              let longitude = locationTracker.locationManager.location?.coordinate.longitude,
              trackLocation else {
            notesHandler.addNote(noteText: noteText, latitude: 20.0, longitude: 20.0)
            exitRoute()
            return
        }
        notesHandler.addNote(noteText: noteText, latitude: latitude, longitude: longitude)
        exitRoute() `

jankrzempek (changes requested):
Few small changes only, great work and awesome code 😎

I will try to test today afternoon when I will be travelling my sweet home

jankrzempek (approved):
Code seems great!! I will test soon and report the results to you 😄

AirCasting/Notes/AddNote/AddNoteViewModel.swift
```diff
@@ -9,14 +9,14 @@ class AddNoteViewModel: ObservableObject {
...
-    init(sessionUUID: SessionUUID, exitRoute: @escaping () -> Void) {
+    init(sessionUUID: SessionUUID, withLocation: Bool, exitRoute: @escaping () -> Void) {
         self.exitRoute = exitRoute
         self.notesHandler = Resolver.resolve(NotesHandler.self, args: sessionUUID)
     }
     
     func continueTapped() {
         notesHandler.addNote(noteText: noteText)
-        exitRoute()
+            exitRoute()
```
jankrzempek, line 19: The indentation here plz

AirCasting/Notes/AddNote/AddNoteViewModel.swift
```diff
@@ -9,14 +9,14 @@ class AddNoteViewModel: ObservableObject {
     private let notesHandler: NotesHandler
     private let exitRoute: () -> Void
     
-    init(sessionUUID: SessionUUID, exitRoute: @escaping () -> Void) {
+    init(sessionUUID: SessionUUID, withLocation: Bool, exitRoute: @escaping () -> Void) {
```
jankrzempek, line 12: Doń;t you want to use the info about location: `withLocation`? Now I think it is not used, as before you were providing fake location info when this `var` was `true`

jankrzempek (commented):
Please consider those two comments only 😄

anna1901 (changes requested):

AirCasting/Notes/NotesHandler.swift
```diff
@@ -44,13 +44,15 @@ class NotesHandlerDefault: NSObject, NotesHandler, NSFetchedResultsControllerDel
     }
     
     func addNote(noteText: String) {
+        let latitude = locationTracker.locationManager.location?.coordinate.latitude ?? locationTracker.googleLocation.last?.location.latitude ?? 20.0
```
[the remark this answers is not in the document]
SzymonGorowski: I add google location because Janek told me about the case when the user turns off the location in settings when the mobile session is active, it's not a rare case but it covers it. These three values (2 default) cover all cases 1st session with location, the second when the user turns off the location in settings when the mobile session is active, and the third in location less session when it doesn't matter where the note will be placed. 

Also, I removed withlocation parameter but I don't know why it is included in this commit ...

I hope everything will be clear if not please let me know.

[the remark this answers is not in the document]
anna1901: I think we would want to use this `withLocation` parameter, because when user turns on `disable location` is settings, the location manager still has it's location, so we are saving a real location instead of the fake one. At least that's what I found with those logs 

Screenshot 2022-04-14 at 15 40 00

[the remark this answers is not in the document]
SzymonGorowski: Maybe I do not understand it correctly but I checked which value is used and when the user turns off location during the recording session the location is used from the code `locationTracker.googleLocation.last?.location.longitude`. When I removed the above code the location was used with default values.

I removed the "with location" property because is not used when we fetch localization only in NotesHandler.

jankrzempek (approved):
Everything seems fine, working, and well structured 🔥

anna1901 (approved):

PR #665, opened 2022-04-12
Created and used protection screen.

jankrzempek:
We should inform the user, that force closing the app can lead to unwanted issues. There we go!

https://trello.com/c/UuRsghpc/607-consider-custom-app-switcher-screen-to-ensure-that-app-wont-be-terminated-during-recording-a-session

### DIFF
--- a/AirCasting/AirCastingApp.swift
+++ b/AirCasting/AirCastingApp.swift
@@ -14,6 +14,7 @@ struct AirCastingApp: App {
     @UIApplicationDelegateAdaptor(AppDelegate.self) var appDelegate
     
     @Environment(\.scenePhase) var scenePhase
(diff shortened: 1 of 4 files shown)


AirCasting/AirCastingApp.swift
```diff
@@ -14,6 +14,7 @@ struct AirCastingApp: App {
     @UIApplicationDelegateAdaptor(AppDelegate.self) var appDelegate
     
     @Environment(\.scenePhase) var scenePhase
+    @State var shoulProtect = false
```
SzymonGorowski, line 17: I suppose this is a typo mistake that should be changed in a few places. Anyway, everything is okay. 👍

PR #666, opened 2022-04-12
Searchfollow - buttons parameters

jankrzempek:
The idea behind this PR is to allow user to select different parameters and sensor accordingly.

https://trello.com/c/00sU4Sw1/537-add-searching-for-purpleair-and-airbeam-data-to-search-feature


AirCasting.xcodeproj/project.pbxproj
```diff
@@ -2374,6 +2386,7 @@
 			attributes = {
 				LastSwiftUpdateCheck = 1240;
 				LastUpgradeCheck = 1320;
+				ORGANIZATIONNAME = Lunar;
```
anna1901, line 2389: Could you tell me what is this change?

jankrzempek: Yeah, so whenever I was creating a new file I have there my name. After adding this I am back with a proper stamp that says "Lunar ..." . This was the easiest solution

AirCasting/SearchAndFollow/SearchLocation/SearchViewModel.swift
```diff
@@ -5,17 +5,37 @@ import Foundation
...
     @Published var addressName = ""
     @Published var addresslocation = CLLocationCoordinate2D(latitude: 20.0, longitude: 20.0)
     @Published var alert: AlertInfo?
-    var continueDisabled: Bool { addressName == "" }
+    @Published private var parameter: MapDownloaderMeasurementType = .ozone
+    @Published private var sensor: MapDownloaderSensorType = .OzoneSensor
     
+    var continueDisabled: Bool { addressName == "" }
+    var shoudShowPMChoiceSheet: Bool { parameter == .particulateMatter }
+    var getParameter: MapDownloaderMeasurementType { parameter }
```
anna1901, line 22: Why do we need those variables separately?

anna1901 (approved):
Amazing! 🎉 Very well done, I'm impressed! 🚀

anna1901:
I'm just wondering why you had to add this organization name in this PR all of the sudden 🤔

AirCasting/SearchAndFollow/MapSessionsDownloader/API/SessionsForLocationDownloader.swift
```diff
@@ -34,9 +34,9 @@ class SessionsForLocationDownloaderDefault: SessionsForLocationDownloader {
                                        north: geoSquare.north,
                                        limit: 100,
                                        offset: 0,
-                                       sensorName: "openaq-pm2.5",
-                                       measurementType: MapDownloaderMeasurementType.particulateMatter.name,
-                                       unitSymbol: MapDownloaderUnitSymbol.uqm3.name)
+                                       sensorName: sensor.apiName,
+                                       measurementType: parameter.apiName,
+                                       unitSymbol: (parameter == .particulateMatter) ? MapDownloaderUnitSymbol.uqm3.name : MapDownloaderUnitSymbol.ppb.name)
```
pawelgil, line 39: I know it's probably temporary here, but it has a flow I need to point out:
It we add a third unit symbol to `MapDownloaderMeasurementType` this code will break. We want it to not break :) It'd go with an extension

AirCasting/SearchAndFollow/SearchLocation/PMSensorType.swift
```diff
@@ -0,0 +1,22 @@
+// Created by Lunar on 12/04/2022.
+//
+
+import Foundation
+
+enum PMSensorType: Codable, CaseIterable, Identifiable {
```
pawelgil, line 6: This is complicating things. Problem with this is that with each added unit you'd have to create a new enum type, and than adjust all the classes. Here we are combining a unit type with a sensor type, why not go with the good 'ol struct or even a tuple?

```Swift
struct Cośtam {
  let sensorType: SensorType // AB3and2, OpenAQ etc.
  let unitType: UnitType // PM2, Ozone etc.
}
```

AirCasting/SearchAndFollow/SearchLocation/OzoneSensorType.swift
```diff
@@ -0,0 +1,18 @@
+// Created by Lunar on 12/04/2022.
+//
+
+import Foundation
+
+enum OzoneSensorType: Codable, CaseIterable, Identifiable {
```
pawelgil, line 6: Same as for the `PMSensorType`

AirCasting/SearchAndFollow/SearchLocation/SearchView.swift
```diff
@@ -18,6 +18,42 @@ struct SearchView: View {
         VStack(alignment: .leading) {
             title
             textField
+            parametersQuestion
+            HStack(spacing: 12) {
+                ForEach(ParameterType.allCases) { param in
```
pawelgil, line 23: this is an interesting technique for `enum` usage, but let's stick to the `ViewModel` approach so we don't mix 10 different coding styles across the project. I also think it would be more readable if we just had `viewModel.searchParameters` exposed:
```Swift
string SearchParameter {
  let isSelected: Bool
  let name: String
}
```

AirCasting/SearchAndFollow/SearchLocation/SearchView.swift
```diff
@@ -18,6 +18,42 @@ struct SearchView: View {
...
+                        }
+                    } label: {
+                        Text(param.capitalizedName)
+                            .padding([.all], 5)
+                    }.buttonStyle(MultiSelectButtonStyle(isSelected: param.capitalizedName == viewModel.getParameter.capitalizedName))
+                        .padding(.bottom, 5)
+                }
+            }
+            sensorQuestion
+            if viewModel.shoudShowPMChoiceSheet {
```
pawelgil, line 36: I think that should be way more flexible. Please design viewmodel so that if we want to add another sensor choice we won't need to make a single change to this view.

pawelgil (changes requested):
Let's get rid of enum based `ForEach` loops and redesign viewmodel to be less set in stone and more welcoming for future changes

AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift
```diff
@@ -84,14 +84,20 @@ class CompleteScreenViewModel: ObservableObject {
     
     let exitRoute: () -> Void
     
-    init(session: SearchSessionResult, sensorType: String) {
+    private let session: PartialExternalSession
```
anna1901, line 87: I'm wondering why those changes are here if they are already on develop? 🤔

anna1901 (approved):

anna1901:
[Add searching for PurpleAir and Airbeam data to search feature](https://trello.com/c/00sU4Sw1/537-add-searching-for-purpleair-and-airbeam-data-to-search-feature)

PR #668, opened 2022-04-12
Search complete screen logic

anna1901:
https://trello.com/c/NhaFepTn

This PR adds logic to the complete screen for the search and follow feature.
For now, we are only displaying the stream for which we get the information from map view. In the future, we need to implement the call to backend to get other streams ids (for aribeam sessions). As I understand, there is no such endpoint now for getting stream information for sessions of other users, so this will require some additional work.


AirCasting/APICommunicator/StreamDownloader.swift
```diff
@@ -0,0 +1,74 @@
+// Created by Lunar on 29/03/2022.
+//
+
+import Foundation
+import Resolver
+
+struct StreamWithMeasurementsDownstream: Decodable {
```
pawelgil, line 7: a little naming issue: `Downstream` is usually used as a suffix for the `protocol` that is responsible for getting data, not data itself. This one can be named either `StreamWithMeasurements` or `StreamWithMeasurementsDownstreamData`

AirCasting/APICommunicator/StreamDownloader.swift
```diff
@@ -0,0 +1,74 @@
...
+    
+    struct Measurements: Decodable {
+        let value: Double
+        let time: Int
+        let longitude: Double
+        let latitude: Double
+    }
+}
+
+protocol StreamDownloader {
```
pawelgil, line 24: `StreamDownstream` here would be appropriate (although funny naming coincidence)

AirCasting/APICommunicator/StreamDownloader.swift
```diff
@@ -0,0 +1,74 @@
...
+        let decoder = JSONDecoder()
+        decoder.keyDecodingStrategy = .convertFromSnakeCase
+        return decoder
+    }()
+    
+    func downloadStreamWithMeasurements(id: String, completion: @escaping (Result<StreamWithMeasurementsDownstream, Error>) -> Void) {
+        let urlComponentPart = urlProvider.baseAppURL.appendingPathComponent("api/fixed/streams/\(id).json")
+        var urlComponents = URLComponents(string: urlComponentPart.absoluteString)!
+        urlComponents.queryItems = [
+            URLQueryItem(name: "measurements_limit", value: "1140")
```
pawelgil, line 44: I think the limit value should be passed in. Either via func parameter or an init value for the whole service.

AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift
```diff
@@ -9,32 +9,69 @@ struct SessionStreamViewModel: Identifiable {
...
+            return session
+        }
+    }
+    
     @Published var selectedStream: Int?
+    @Published var selectedStreamUnitSymbol: String?
+    @Published var chartStartTime: Date?
+    @Published var chartEndTime: Date?
     @Published var isMapSelected: Bool = true
+    @Published var showAlert = false
```
pawelgil, line 74: I think this var is unused

AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift
```diff
@@ -44,41 +81,42 @@ class CompleteScreenViewModel: ObservableObject {
     let sessionEndTime: Date
     let sensorType: String
     @Published var sessionStreams: Loadable<[SessionStreamViewModel]> = .loading
-    @Published var chartViewModel = SearchAndFollowChartViewModel(stream: nil)
+    @Published var chartViewModel = SearchAndFollowChartViewModel()
     
-    private let session: SearchSessionResult
-    @Injected private var service: SearchSessionStreamsDownstream
+    private let session: PartialExternalSession
+    private var service = DefaultStreamDownloader()
```
pawelgil, line 87: please inject this one

AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift
```diff
@@ -44,41 +81,42 @@ class CompleteScreenViewModel: ObservableObject {
...
+    private var isPresented: Binding<Bool>
```
pawelgil, line 88: I'm not a fan of using bindings for dismissal and much rather prefer the `exitRoute` strategy that we use in other parts of the code

AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift
```diff
@@ -44,41 +81,42 @@ class CompleteScreenViewModel: ObservableObject {
...
-                        .init(id: $0.id,
-                              sensorName: Self.showStreamName($0.sensorName),
-                              lastMeasurementValue: $0.measurements.last?.value ?? 0)
-                    })
-                )
-                self.selectedStream = downloadedSession.streams.first?.id
-                if let stream = downloadedSession.streams.first {
-                    self.chartViewModel.setStream(to: stream)
+                switch error {
+                case .noThresholdsFound:
```
pawelgil, line 113: hmm, makes me think if this "no thresohlds" scenario is really an error. Maybe we should change the store to be able to successfully return nil value? Smth like `Result<Thresholds?, Error> `

AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift
```diff
@@ -96,7 +134,65 @@ class CompleteScreenViewModel: ObservableObject {
         selectedStream = id
     }
     
-    private static func showStreamName(_ streamName: String) -> String {
+    func xMarkTapped() {
+        isPresented.wrappedValue = false
+    }
+    
+    func dismissView() {
```
pawelgil, line 141: such method should not be exposed for viewmodels. Public methods should always be a response of some user action and should reflect that in a name. "Dismiss view" is not an action, it's an order, and a view should never tell its viewmodel what to do, only report what happened.

Ofc if this is just an internal method than just put a `private` in front :)

pawelgil (changes requested):
Some minor stuff to correct and we'll be good to go

pawelgil (approved):

PR #669, opened 2022-04-12
Adding building outlines to the maps

SzymonGorowski:
change visibility of buildings on the map in the map style JSON file


jankrzempek (approved):
Perfect buildings outline, it's a lot better 🥳

anna1901 (approved):

PR #672, opened 2022-04-14
Protect screen - only mobile.

jankrzempek:
Protect screen shows only when mobile session is on.

https://trello.com/c/UuRsghpc/607-consider-custom-app-switcher-screen-to-ensure-that-app-wont-be-terminated-during-recording-a-session -- comment

### DIFF
--- a/AirCasting/AirCastingApp.swift
+++ b/AirCasting/AirCastingApp.swift
@@ -18,6 +18,8 @@ struct AirCastingApp: App {
     private let syncScheduler: SynchronizationScheduler
     @Injected private var sessionSynchronizer: SessionSynchronizer
     @Injected private var persistenceController: PersistenceController
+    @Injected private var mobilePeripheralSessionManager: MobilePeripheralSessionManager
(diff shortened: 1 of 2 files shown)


AirCasting/AirCastingApp.swift
```diff
@@ -39,11 +41,15 @@ struct AirCastingApp: App {
         }.onChange(of: scenePhase) { newScenePhase in
             switch newScenePhase {
             case .active:
-                shouldProtect = false
+                if mobilePeripheralSessionManager.isMobileSessionActive || microphoneManager.isRecording {
```
anna1901, line 44: I think this check is not necessary? And maybe it is safer to not include it, because for example if the session stops being active when the app is in the background, then shouldProtect won't be set to false when the app comes back to foreground and we would want that to happen

jankrzempek: Nice point of looking at that. I will delete this condition from here

SzymonGorowski (approved):
Good job, nice feature! 🚀

PR #676, opened 2022-04-15
Don't allow saving session with duplicated uuid and fix updating session version

anna1901:
This PR was created to fix a bug with duplicated sessions in the database. Two changes were made:
1. context save after version update. The call to this method was called in a different thread so the context save in `accessStorage` was called even before this method got executed
2. in persistenceController in instertSessions function we are adding a check if the session with given uuid doesn't already exist in the database


AirCasting/CoreData/MeasurementStreamStorage.swift
```diff
@@ -269,6 +270,8 @@ final class HiddenCoreDataMeasurementStreamStorage: MeasurementStreamStorageCont
     func updateVersion(for sessionUUID: SessionUUID, to version: Int) throws {
         let sessionEntity = try context.existingSession(uuid: sessionUUID)
         sessionEntity.version = Int16(version)
+        // I needed to add this, because in the note editing function this get's called from another thread and the context save in `accessStorage` function was getting called before this function got exectuted.
+        try context.save()
```
pawelgil, line 274: I think the error is on the client code side and should be fixed there rather than patched here.
The proper solution would be to add another `accessStorage` inside the **async callback** in our note editing functions like so (`deleteNote` function, but it needs to be done in other places too):
```Swift
measurementStreamStorage.accessStorage { [self] storage in
    do {
        try storage.deleteNote(note, for: sessionUUID)
        fetchSession(storage: storage) { session in
            self.sessionUpdateService.updateSession(session: session) { result in
                switch result {
                case .success(let updateData):
                    // HERE!
                    self.measurementStreamStorage.accessStorage {
                        try? $0.updateVersion(for: self.sessionUUID, to: updateData.version)
                        Log.info("Notes successfully updated")
                        completion()
                    }
                    //~~~
                case .failure(let error):
                    Log.info("Failed updating session while updating notes: \(error.localizedDescription)")
                    completion()
                }
            }
        }
    } catch {
        Log.info("Error when deleting note: \(error)")
    }
}
```

AirCasting/CoreData/PersistenceController+Database.swift
```diff
@@ -29,6 +29,11 @@ extension PersistenceController: SessionInsertable {
         let context = self.editContext
         context.perform {
             sessions.forEach {
+                    // This is added to ensure that we don't add session if a session with this uuid already existis in the database
+                    guard (try? context.existingSession(uuid: $0.uuid)) == nil else {
+                        Log.warning("Tried to add new session with same UUID")
```
pawelgil, line 34: I think this should be an `error` log (the difference is that `error` logs are logged to firebase as non-crashing errors. I think we'd like to be informed of this case)

pawelgil (changes requested):

AirCasting/Notes/NotesHandler.swift
```diff
@@ -60,18 +60,22 @@ class NotesHandlerDefault: NSObject, NotesHandler, NSFetchedResultsControllerDel
...
+                try storage.deleteNote(note, for: self.sessionUUID)
+                self.fetchSession(storage: storage) { session in
                     self.sessionUpdateService.updateSession(session: session) { result in
                         switch result {
                         case .success(let updateData):
-                            try? storage.updateVersion(for: self.sessionUUID, to: updateData.version)
-                            Log.info("Notes successfully updated")
-                            completion()
+                            self.measurementStreamStorage.accessStorage { [weak self] storage in
+                                guard let self = self else { return }
```
pawelgil, line 74: hmm, using `weak self` in this `deleteNote` might not be the best. I don't see any retain cycles possible, and by using `weak` we don't hold on to the `self` till delete is finished. So this kind of code:
```Swift
func asd() {
    let handler = NotesHandlerDefault(...)
    handler.deleteNote(note: note) {
        print("Deleted!")
    }
}
```

will never execute `print("Deleted!")`.

IDK if that's realistic here

pawelgil (approved):

PR #677, opened 2022-04-15
Search and follow chart averaging

anna1901:


anna1901:
[Implement averaging in the chart](https://trello.com/c/IbhLI7o8/618-implement-averaging-in-the-chart)

AirCasting/SearchAndFollow/CompleteScreen/Views/SearchAndFollowChartViewModel.swift
```diff
@@ -19,13 +19,32 @@ class SearchAndFollowChartViewModel: ObservableObject {
...
+            }
+            
+            if buffer.isEmpty || hourIsAlreadyPresent(in: buffer.map({ $0.time }), date: measurement.time) {
+                buffer.append(measurement)
+                continue
+            }
+            
+            let average = (buffer.map { $0.value }.reduce(0, +)) / Double(buffer.count)
+            
+            times.append(buffer.last!.time.roundedUpToHour)
```
jankrzempek, line 37: The only thing is force unwrapping here, but after my long mind-analyze, I think it should be fine. Mentioning this only for the sake of having this in mind :)

jankrzempek (approved):
Seems fine 😎

AirCasting/SearchAndFollow/CompleteScreen/Views/SearchAndFollowChartViewModel.swift
pawelgil: idk, for me it's ok, but on the edge. The code that guarantees that `buffer.last` won't ever be `nil` is like 4 lines up

pawelgil (approved):

PR #690, opened 2022-04-26
Ability to download missing streams for SearchFollow feature.

jankrzempek:
https://trello.com/c/Jd6CYYFH/626-display-all-of-the-airbeam-streams

### DIFF
--- a/AirCasting/SearchAndFollow/AirBeamMeasurements/AirBeamMeasurementsDownloader.swift
+++ b/AirCasting/SearchAndFollow/AirBeamMeasurements/AirBeamMeasurementsDownloader.swift
@@ -1,0 +1,59 @@
+// Created by Lunar on 26/04/2022.
+//
+import SwiftUI
+import Resolver
+
+final class AirBeamMeasurementsDownloader {
+    @Injected private var urlProvider: URLProvider
+    @Injected private var client: APIClient
+    @Injected private var responseValidator: HTTPResponseValidator
+    private let responseHandler = AuthorizationHTTPResponseHandler()
+    
+    func downloadStreams(using sessionID: Int, completion: @escaping (Result<[MeasurementsDownloaderResultModel], Error>) -> Void) throws {
+        let decoder = JSONDecoder()
+        let group = DispatchGroup()
+        var requests = [URLRequest]()
+        var combinedResult = [MeasurementsDownloaderResultModel]()
+        
+        AirBeamStreamSuffixes.allCases.forEach { s in
+            let urlComponentPart = urlProvider.baseAppURL.appendingPathComponent("api/fixed/sessions/\(sessionID).json")
+            var urlComponents = URLComponents(string: urlComponentPart.absoluteString)!
+            urlComponents.queryItems = [
+                URLQueryItem(name: "sensor_name", value: "airbeam3-\(s.rawName)"),
+                // optional parameter: measurements_limit (on the web is equal to 1440)
+                URLQueryItem(name: "measurements_limit", value: "10"),
+            ]
+            let url = urlComponents.url!
+            var request = URLRequest(url: url)
+            request.httpMethod = "GET"
+            request.addValue("application/json", forHTTPHeaderField: "Content-Type")
+            request.addValue("application/json", forHTTPHeaderField: "Accept")
+            requests.append(request)
+        }
+        
+        for request in requests {
+            group.enter()
+            client.requestTask(for: request) { [responseValidator] result, _ in
+                switch result {
+                case .success(let response):
+                    do {
+                        try responseValidator.validate(response: response.response, data: response.data)
+                        let resultPart = try decoder.decode(MeasurementsDownloaderResultModel.self, from: response.data)
+                        combinedResult.append(resultPart)
+                        group.leave()
+                    } catch {
+                        completion(.failure(error))
+                        fatalError("Error when downloading one of the AirBeam streams: \(error)")
+                    }
+                case .failure(let error):
+                    completion(.failure(error))
+                    fatalError("Error when downloading one of the AirBeam streams: \(error)")
+                }
+            }
+        }
+        group.notify(queue: .main) {
+            Log.info("Downloading all of the streams for AirBeam - completed.")
+            completion(.success(combinedResult))
+        }
+    }
+}
--- a/AirCasting/SearchAndFollow/AirBeamMeasurements/MeasurementsDownloaderResultModel.swift
+++ b/AirCasting/SearchAndFollow/AirBeamMeasurements/MeasurementsDownloaderResultModel.swift
@@ -1,0 +1,16 @@
+// Created by Lunar on 26/04/2022.
+//
+
+import Foundation
+
+struct MeasurementsDownloaderResultModel: Decodable {
+    let sensorName: String
+    let measurements: [theMeasurement]
+}
+
+struct theMeasurement: Decodable {
+    let value: Double
+    let time: Double
+    let longitude: Double
+    let latitude: Double
+}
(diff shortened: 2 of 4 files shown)


AirCasting/SearchAndFollow/AirBeamMeasurements/AirBeamMeasurementsDownloader.swift
```diff
@@ -0,0 +1,59 @@
...
+    @Injected private var responseValidator: HTTPResponseValidator
+    private let responseHandler = AuthorizationHTTPResponseHandler()
+    
+    func downloadStreams(using sessionID: Int, completion: @escaping (Result<[MeasurementsDownloaderResultModel], Error>) -> Void) throws {
+        let decoder = JSONDecoder()
+        let group = DispatchGroup()
+        var requests = [URLRequest]()
+        var combinedResult = [MeasurementsDownloaderResultModel]()
+        
+        AirBeamStreamSuffixes.allCases.forEach { s in
```
SzymonGorowski, line 18: One tiny cosmetic change maybe you should use `stream` instead of `s`.

AirCasting/SearchAndFollow/AirBeamMeasurements/AirBeamMeasurementsDownloader.swift
```diff
@@ -0,0 +1,59 @@
...
+            client.requestTask(for: request) { [responseValidator] result, _ in
+                switch result {
+                case .success(let response):
+                    do {
+                        try responseValidator.validate(response: response.response, data: response.data)
+                        let resultPart = try decoder.decode(MeasurementsDownloaderResultModel.self, from: response.data)
+                        combinedResult.append(resultPart)
+                        group.leave()
+                    } catch {
+                        completion(.failure(error))
```
pawelgil, line 45: `group.leave()` missing in the `catch` block

AirCasting/SearchAndFollow/AirBeamMeasurements/MeasurementsDownloaderResultModel.swift
```diff
@@ -0,0 +1,16 @@
...
+//
+
+import Foundation
+
+struct MeasurementsDownloaderResultModel: Decodable {
+    let sensorName: String
+    let measurements: [theMeasurement]
+}
+
+struct theMeasurement: Decodable {
```
pawelgil, line 11: capitalize the name plz

pawelgil: Also - you named it `TheMeasurement` because `Measurement` was already taken. Why not putting this inside the ResultModel struct and just call it `Measurement`?

```Swift
struct MeasurementsDownloaderResultModel: Decodable {
     let sensorName: String
     let measurements: [Measurement]

    struct Measurement: Decodable { ... }
 }
```

pawelgil (changes requested):

AirCasting/SearchAndFollow/AirBeamMeasurements/AirBeamMeasurementsDownloader.swift
```diff
@@ -0,0 +1,59 @@
...
+                        fatalError("Error when downloading one of the AirBeam streams: \(error)")
+                    }
+                case .failure(let error):
+                    completion(.failure(error))
+                    fatalError("Error when downloading one of the AirBeam streams: \(error)")
```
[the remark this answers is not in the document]
jankrzempek: I was reading lately that sometimes the best way to stop errors and to avoid propagating them further is to crash. That was the reason I implemented it that way. I will change it then 😄

AirCasting/SearchAndFollow/AirBeamMeasurements/AirBeamMeasurementsDownloader.swift
```diff
@@ -0,0 +1,59 @@
...
+                        try responseValidator.validate(response: response.response, data: response.data)
+                        let resultPart = try decoder.decode(MeasurementsDownloaderResultModel.self, from: response.data)
+                        combinedResult.append(resultPart)
+                        group.leave()
+                    } catch {
+                        completion(.failure(error))
+                        fatalError("Error when downloading one of the AirBeam streams: \(error)")
+                    }
+                case .failure(let error):
+                    completion(.failure(error))
```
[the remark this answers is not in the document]
jankrzempek: I was not using it as I decided to crash the app. It was fine then, right? I will change it as I am going to delete fataError

AirCasting/SearchAndFollow/AirBeamMeasurements/AirBeamMeasurementsDownloader.swift
```diff
@@ -0,0 +1,71 @@
...
+                case .failure(let error):
+                    Log.info("Error when downloading one of the AirBeam streams: \(error)")
+                    group.leave()
+                    completion(.failure(error))
+                }
+            }
+        }
+        group.notify(queue: .main) {
+            Log.info("Downloading all of the streams for AirBeam - completed.")
+            completion(.success(combinedResult))
```
pawelgil, line 47: problem here is - this `notify` block will get called always. So for the scenario that 2 out of 5 stream downloads will fail the completion will get called twice with `.failure(error)` and after that it will get called with `.success(combinedResult)`. Sorry I didn't pick that before... Usually its done like this:
```Swift
var accumulatedResults : [Result] = []
var allErrors: [Error] = []
...
group.notify(queue: .main) {
 // Here, if allErrors is not empty we can decide what to do next
}
```

This way we won't get multiple `completion` calls

AirCasting/SearchAndFollow/AirBeamMeasurements/AirBeamMeasurementsDownloader.swift
```diff
@@ -0,0 +1,71 @@
...
+    
+    private func prepareRequests(using sessionID: Int, sensor: AirBeamStreamPrefix) -> [URLRequest] {
+        var requests = [URLRequest]()
+        AirBeamStreamSuffixes.allCases.forEach { s in
+            let urlComponentPart = urlProvider.baseAppURL.appendingPathComponent("api/fixed/sessions/\(sessionID).json")
+            let url = structureURLComponents(with: urlComponentPart, name: s, sensor: sensor)
+            let request = URLRequest.jsonGET(url: url)
+            requests.append(request)
+        }
+        return requests
```
pawelgil, line 59: good job with this function, but please use `map` for `AirBeamStreamSuffixes.allCases` instead of a for loop with this `var requests` helper

pawelgil (changes requested):

AirCasting/SearchAndFollow/AirBeamMeasurements/AirBeamMeasurementsDownloader.swift
```diff
@@ -31,32 +32,31 @@ final class AirBeamMeasurementsDownloaderDefault: AirBeamMeasurementsDownloader
...
                 case .failure(let error):
+                    allErrors.append(error)
                     Log.info("Error when downloading one of the AirBeam streams: \(error)")
                     group.leave()
-                    completion(.failure(error))
                 }
             }
         }
         group.notify(queue: .main) {
+            if !allErrors.isEmpty { completion(.failure(allErrors.first!)) }
```
pawelgil, line 47: we're only passing the first error here. I think that's reasonable here, but the *perfect* solution would be to pass all under a [composite](https://refactoring.guru/design-patterns/composite) error. Could look like this:
```Swift
struct CompositeError: Error, LocalizedError {
    let underlyingErrors: [Error]
    public var errorDescription: String? {
        // nice description that prints out all underlying descriptions
    }
}
```

pawelgil (approved):

PR #693, opened 2022-05-04
Updating slider and graph for celsius

SzymonGorowski:
https://trello.com/c/bGlibqUD/520-celsius-update-hlu

### DIFF
--- a/AirCasting/Graph/Graph.swift
+++ b/AirCasting/Graph/Graph.swift
@@ -64,7 +64,7 @@ struct Graph: UIViewRepresentable {
         let uiView = AirCastingGraph(onDateRangeChange: { newRange in
             rangeChangeAction?(newRange)
         })
-        try? uiView.updateWithThreshold(thresholdValues: thresholds.rawThresholdsBinding.wrappedValue)
+        try? uiView.updateWithThreshold(thresholdValues: thresholds.sensorName == MeasurementStreamSensorName.f.rawValue && userSettings.convertToCelsius ? thresholds.rawThresholdsBindingCelsius.wrappedValue : thresholds.rawThresholdsBinding.wrappedValue)
         let entries = stream.allMeasurements?.sorted(by: { $0.time < $1.time }).compactMap({ measurement -> ChartDataEntry? in
             let timeInterval = Double(measurement.time.timeIntervalSince1970)
             let chartDataEntry = ChartDataEntry(x: timeInterval, y: getValue(of: measurement))
@@ -102,7 +102,7 @@ struct Graph: UIViewRepresentable {
                 context.coordinator.totalNumberOfMeasurements != stream.allMeasurements?.count ||
                 stream != context.coordinator.stream else { return }
         
-        try? uiView.updateWithThreshold(thresholdValues: thresholds.rawThresholdsBinding.wrappedValue)
+        try? uiView.updateWithThreshold(thresholdValues: thresholds.sensorName == MeasurementStreamSensorName.f.rawValue && userSettings.convertToCelsius ? thresholds.rawThresholdsBindingCelsius.wrappedValue : thresholds.rawThresholdsBinding.wrappedValue)
         let allLimitLines = getLimitLines()
         uiView.limitLines = allLimitLines
         
--- a/AirCasting/Map/AirMapView.swift
+++ b/AirCasting/Map/AirMapView.swift
@@ -93,7 +93,7 @@ struct AirMapView: View {
                     
                     if let selectedStream = selectedStream {
                         NavigationLink(destination: ThresholdsSettingsView(thresholdValues: threshold.thresholdsBinding,
-                                                                           initialThresholds: selectedStream.thresholds)) {
+                                                                           initialThresholds: selectedStream.thresholds, threshold: threshold)) {
                             EditButtonView()
                         }.padding([.bottom, .leading, .trailing])
                     }
--- a/AirCasting/SessionViews/MeasurementDotView.swift
+++ b/AirCasting/SessionViews/MeasurementDotView.swift
@@ -2,12 +2,14 @@
 //
 
 import SwiftUI
+import Resolver
 
 struct MeasurementDotView: View {
     
     let value: Double
     @ObservedObject var thresholds: SensorThreshold
-    
+    @InjectedObject private var userSettings: UserSettings
+
     var body: some View {
         color
             .clipShape(Circle())
@@ -15,17 +17,32 @@ struct MeasurementDotView: View {
     }
     
     var color: Color {
-        switch Int32(value) {
-        case thresholds.thresholdVeryLow ..< thresholds.thresholdLow:
-            return Color.aircastingGreen
-        case thresholds.thresholdLow ..< thresholds.thresholdMedium:
-            return Color.aircastingYellow
-        case thresholds.thresholdMedium ..< thresholds.thresholdHigh :
-            return Color.aircastingOrange
-        case thresholds.thresholdHigh ... thresholds.thresholdVeryHigh :
-            return Color.aircastingRed
-        default:
-            return Color.aircastingGray
+        if thresholds.sensorName == MeasurementStreamSensorName.f.rawValue && userSettings.convertToCelsius {
+            switch Int32(value) {
+            case Int32(TemperatureConverter.calculateCelsius(fahrenheit: Double(thresholds.thresholdVeryLow))) ..< Int32(TemperatureConverter.calculateCelsius(fahrenheit: Double(thresholds.thresholdLow))):
+                return Color.aircastingGreen
+            case Int32(TemperatureConverter.calculateCelsius(fahrenheit: Double(thresholds.thresholdLow))) ..< Int32(TemperatureConverter.calculateCelsius(fahrenheit: Double(thresholds.thresholdMedium))):
+                return Color.aircastingYellow
+            case Int32(TemperatureConverter.calculateCelsius(fahrenheit: Double(thresholds.thresholdMedium))) ..< Int32(TemperatureConverter.calculateCelsius(fahrenheit: Double(thresholds.thresholdHigh))):
+                return Color.aircastingOrange
+            case Int32(TemperatureConverter.calculateCelsius(fahrenheit: Double(thresholds.thresholdHigh))) ... Int32(TemperatureConverter.calculateCelsius(fahrenheit: Double(thresholds.thresholdVeryHigh))):
+                return Color.aircastingRed
+            default:
+                return Color.aircastingGray
+            }
+        } else {
+            switch Int32(value) {
+            case thresholds.thresholdVeryLow ..< thresholds.thresholdLow:
+                return Color.aircastingGreen
+            case thresholds.thresholdLow ..< thresholds.thresholdMedium:
+                return Color.aircastingYellow
+            case thresholds.thresholdMedium ..< thresholds.thresholdHigh :
+                return Color.aircastingOrange
+            case thresholds.thresholdHigh ... thresholds.thresholdVeryHigh :
+                return Color.aircastingRed
+            default:
+                return Color.aircastingGray
+            }
         }
     }
 }
--- a/AirCasting/SessionViews/MultiSliderView.swift
+++ b/AirCasting/SessionViews/MultiSliderView.swift
@@ -6,6 +6,7 @@
 //
 
 import SwiftUI
+import Resolver
 
 struct ThresholdsSliderView: View {
     
@@ -17,7 +18,7 @@ struct ThresholdsSliderView: View {
 }
 
 struct MultiSliderView: View {
-    
+    @InjectedObject private var userSettings: UserSettings
     @Binding var thresholds: [Float]
     
     private var thresholdButtonValues: [Float] {
@@ -93,7 +94,7 @@ struct MultiSliderView: View {
     func labels(geometry: GeometryProxy) -> some View {
         let y = geometry.frame(in: .local).size.height / 2
         return ForEach(thresholds.indices, id: \.self) { index in
-            let ints = Int(thresholds[index])
+            let ints = userSettings.convertToCelsius ? Int(TemperatureConverter.calculateCelsius(fahrenheit: Double(thresholds[index]))) : Int(thresholds[index])
             Text("\(ints)")
                 .position(x: calculateXAxisSize(thresholdValue: thresholds[index], geometry: geometry),
                           y: y)
--- a/AirCasting/SessionViews/ThresholdSettings/ThresholdsSettingsViewModel.swift
+++ b/AirCasting/SessionViews/ThresholdSettings/ThresholdsSettingsViewModel.swift
@@ -2,6 +2,7 @@
 //
 
 import Foundation
+import Resolver
 
 class ThresholdSettingsViewModel: ObservableObject {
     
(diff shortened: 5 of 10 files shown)


AirCasting/Graph/Graph.swift
```diff
@@ -64,7 +64,7 @@ struct Graph: UIViewRepresentable {
         let uiView = AirCastingGraph(onDateRangeChange: { newRange in
             rangeChangeAction?(newRange)
         })
-        try? uiView.updateWithThreshold(thresholdValues: thresholds.rawThresholdsBinding.wrappedValue)
+        try? uiView.updateWithThreshold(thresholdValues: stream.isTemperature && userSettings.convertToCelsius ? thresholds.rawThresholdsBindingCelsius.wrappedValue : thresholds.rawThresholdsBinding.wrappedValue)
```
jankrzempek, line 67: Here maybe we could use only var to make it clearer?

` var properThresholds: [Float] {
        stream.isTemperature && userSettings.convertToCelsius ? thresholds.rawThresholdsBindingCelsius.wrappedValue : thresholds.rawThresholdsBinding.wrappedValue
    }`

AirCasting/Graph/Graph.swift
```diff
@@ -102,7 +102,7 @@ struct Graph: UIViewRepresentable {
                 context.coordinator.totalNumberOfMeasurements != stream.allMeasurements?.count ||
                 stream != context.coordinator.stream else { return }
         
-        try? uiView.updateWithThreshold(thresholdValues: thresholds.rawThresholdsBinding.wrappedValue)
+        try? uiView.updateWithThreshold(thresholdValues: stream.isTemperature && userSettings.convertToCelsius ? thresholds.rawThresholdsBindingCelsius.wrappedValue : thresholds.rawThresholdsBinding.wrappedValue)
```
jankrzempek, line 105: Thanks to the above here we could use the same var

AirCasting/Map/AirMapView.swift
```diff
@@ -93,11 +93,11 @@ struct AirMapView: View {
                     
                     if let selectedStream = selectedStream {
                         NavigationLink(destination: ThresholdsSettingsView(thresholdValues: threshold.thresholdsBinding,
-                                                                           initialThresholds: selectedStream.thresholds)) {
+                                                                           initialThresholds: selectedStream.thresholds, threshold: threshold)) {
```
jankrzempek, line 96: Please use a new line after coma 😄

AirCasting/Map/Heatmap/GridSquare.swift
```diff
@@ -48,7 +50,7 @@ struct GridSquare {
         calculateAverage()
         
         guard let averagedValue = averagedValue else { return }
-        let color: UIColor = GoogleMapView.color(value: Int32(averagedValue), threshold: sensorThreshold).withAlphaComponent(0.5)
+        let color: UIColor = GoogleMapView.color(value: sensorThreshold.sensorName == MeasurementStreamSensorName.f.rawValue && userSettings.convertToCelsius ? Int32(TemperatureConverter.calculateFahrenheit(celsius: averagedValue)) : Int32(averagedValue), threshold: sensorThreshold).withAlphaComponent(0.5)
```
jankrzempek, line 53: Here again maybe we could consider placing it into some computed var?

jankrzempek (changes requested):
One thing which is really required I think is the decimal keyboard. Could you please change the keyboard style that we use when changing the threshold in the threshold setting panel? 😄 (now I am able to type text there)

Working everything great, cool great code 🥳

AirCasting/SessionViews/ThresholdSettings/ThresholdsSettingsViewModel.swift
```diff
@@ -11,18 +12,23 @@ class ThresholdSettingsViewModel: ObservableObject {
...
+    func resetToDefault() -> ThresholdsValue {
+        return initialThresholds
+    }
     
     func updateToNewThresholds() -> ThresholdsValue {
         let stringThresholdValues = [thresholdVeryHigh, thresholdHigh, thresholdMedium, thresholdLow, thresholdVeryLow]
         var newThresholdValues: [Int32] = []
         for value in stringThresholdValues {
-            let convertedValue = convertToInt(value)
+            let convertedValue = threshold.sensorName == MeasurementStreamSensorName.f.rawValue && userSettings.convertToCelsius ? Int32(TemperatureConverter.calculateFahrenheit(celsius: Double(convertToInt(value)))) : convertToInt(value)
```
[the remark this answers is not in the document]
SzymonGorowski: I think here is impossible because we use the value directly from the loop or probably I do not know how to solve this.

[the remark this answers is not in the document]
jankrzempek: Eh sure, I have missed that. Sorry 😅

AirCasting/Map/Heatmap/GridSquare.swift
```diff
@@ -47,8 +55,7 @@ struct GridSquare {
         number += 1
         calculateAverage()
         
-        guard let averagedValue = averagedValue else { return }
-        let color: UIColor = GoogleMapView.color(value: Int32(averagedValue), threshold: sensorThreshold).withAlphaComponent(0.5)
+        let color: UIColor = GoogleMapView.color(value: properThreshold, threshold: sensorThreshold).withAlphaComponent(0.5)
```
anna1901, line 58: As noted above, `properThreshold` should be renamed to reflect what it actually is, and I think we should keep it as an optional and keep the guard, instead of passing 0 forward.

AirCasting/SessionViews/MeasurementDotView.swift
```diff
@@ -2,30 +2,47 @@
...
-            return Color.aircastingGreen
-        case thresholds.thresholdLow ..< thresholds.thresholdMedium:
-            return Color.aircastingYellow
-        case thresholds.thresholdMedium ..< thresholds.thresholdHigh :
-            return Color.aircastingOrange
-        case thresholds.thresholdHigh ... thresholds.thresholdVeryHigh :
-            return Color.aircastingRed
-        default:
-            return Color.aircastingGray
+        if thresholds.sensorName == MeasurementStreamSensorName.f.rawValue && userSettings.convertToCelsius {
```
anna1901, line 20: I think you wrote a function which you can use it instead of duplicating the code.

```
var color: Color {
        if thresholds.sensorName == MeasurementStreamSensorName.f.rawValue && userSettings.convertToCelsius {
            return thresholds.colorForCelsius(value: Int32(value))
        } else {
            return thresholds.colorFor(value: Int32(value))
        }
    }
```

AirCasting/SessionViews/MultiSliderView.swift
```diff
@@ -6,19 +6,24 @@
 //
 
 import SwiftUI
+import Resolver
 
 struct ThresholdsSliderView: View {
     
     @ObservedObject var threshold: SensorThreshold
+    @Binding var selectedStream: MeasurementStreamEntity?
```
anna1901, line 14: I think you can remove this `@Binding`. The purpose of bindings is to allow to change a State variable of some by the child view. In this case, the slider doesn't change the value of `selectedStream`. I checked, and everything seems to be working fine without it.

AirCasting/SessionViews/MultiSliderView.swift
```diff
@@ -6,19 +6,24 @@
...
+            MultiSliderView(thresholds: threshold.rawThresholdsBinding, selectedStream: selectedStream)
+        }
     }
 }
 
 struct MultiSliderView: View {
-    
+    @InjectedObject private var userSettings: UserSettings
     @Binding var thresholds: [Float]
+    @ObservedObject var selectedStream: MeasurementStreamEntity
```
anna1901, line 26: I think you can remove the `@ObservedObject` wrapper. I checked, and everything seems to be working fine without it.

AirCasting/SessionViews/ThresholdSettings/ThresholdsSettingsViewModel.swift
```diff
@@ -11,18 +12,23 @@ class ThresholdSettingsViewModel: ObservableObject {
...
+    var threshold: SensorThreshold
+    @InjectedObject private var userSettings: UserSettings
+    
+    init(initialThresholds: ThresholdsValue, threshold: SensorThreshold) {
         self.initialThresholds = initialThresholds
+        self.threshold = threshold
     }
 
-    func resetToDefault() -> ThresholdsValue { initialThresholds }
+    func resetToDefault() -> ThresholdsValue {
```
anna1901, line 23: This can be changed back to the way it was done (no explicit `return`)

anna1901 (commented):
Nice progress. :) I noticed that we seem to have wrong thresholds colors in chart and in statistics container. Could you take a look also at that?

AirCasting/Map/Heatmap/GridSquare.swift
```diff
@@ -19,6 +20,13 @@ struct GridSquare {
     private var southEastLatLng: CLLocationCoordinate2D
     var northEastLatLng: CLLocationCoordinate2D
     private var northWestLatLng: CLLocationCoordinate2D
+    @InjectedObject private var userSettings: UserSettings
+    
+    private var properThreshold: Int32 {
```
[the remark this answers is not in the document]
anna1901: Why are we returning 0, when there is no averageValue? If this variable is here to "mirror" the averagedValue, but adjust it to celsius settings, then shouldn't it also be of type `Double?`?

AirCasting/SessionViews/StatisticsContainerView.swift
```diff
@@ -6,11 +6,14 @@
 //
 
 import SwiftUI
+import Resolver
 
 struct StatisticsContainerView<ViewModelType>: View where ViewModelType: StatisticsContainerViewModelable {
     @ObservedObject var statsContainerViewModel: ViewModelType
     @ObservedObject var threshold: SensorThreshold
-    
+    var selectedStream: MeasurementStreamEntity?
```
pawelgil, line 14: we have a `viewModel` here, let's not mix in `Entity` objects

AirCasting/SessionViews/StatisticsContainerView.swift
```diff
@@ -6,11 +6,14 @@
...
+    @InjectedObject private var userSettings: UserSettings
```
pawelgil, line 15: settings should not be used directly here, rather in a `viewModel`

AirCasting/SessionViews/StatisticsContainerView.swift
```diff
@@ -71,6 +74,15 @@ struct StatisticsContainerView<ViewModelType>: View where ViewModelType: Statist
         }.frame(width: 68, height: 33, alignment: .center)
     }
     
+    func color(value: Double) -> Color {
```
pawelgil, line 77: this feels like a `viewModel` responsibility

AirCasting/SessionViews/ChartViewModel.swift
```diff
@@ -20,7 +20,7 @@ final class ChartViewModel: ObservableObject {
         }
     }
     @Injected private var persistence: PersistenceController
-    @Injected private var settings: UserSettings
+    @InjectedObject var settings: UserSettings
```
pawelgil, line 23: this needs to stay private, the whole purpose of `ViewModels` is to hide that kind of objects behind a UI-domain interface

AirCasting/Graph/Graph.swift
```diff
@@ -41,6 +41,10 @@ struct Graph: UIViewRepresentable {
     var isAutozoomEnabled: Bool
     let simplifiedGraphEntryThreshold = 1000
     
+    private var properThresholds: [Float] {
+        stream.isTemperature && userSettings.convertToCelsius ? thresholds.rawThresholdsBindingCelsius.wrappedValue : thresholds.rawThresholdsBinding.wrappedValue
```
pawelgil, line 45: I see this kind of code repeating multiple times across this PR. This is not scalable and very fragile solution. Imagine we'll be tasked to add more settings like that in the future (the simplest example would be "add the possibility to display temperature in Kelvin"). All this code would inflate to a pile of extremely hard to read `if/else` statements real quick and would be really fragile (meaning a mistake would be super easy to make). This is a code smell and it should be packed into a single place somehow. (maybe some `ThresholdFormatter` class that would take a session in and have all the variable stuff like values, colors etc.?)
Please think about something and replace all places you're doing this kind of checks with this better solution

AirCasting/SessionViews/MeasurementDotView.swift
```diff
@@ -2,11 +2,14 @@
...
 
 import SwiftUI
+import Resolver
 
 struct MeasurementDotView: View {
     
     let value: Double
     @ObservedObject var thresholds: SensorThreshold
+    @InjectedObject private var userSettings: UserSettings
+    @ObservedObject var stream: MeasurementStreamEntity
```
pawelgil, line 12: let's keep simple views simple and come up with better solutions than adding more dependencies and logic in here

AirCasting/SessionViews/MultiSliderView.swift
```diff
@@ -6,19 +6,24 @@
 //
 
 import SwiftUI
+import Resolver
 
 struct ThresholdsSliderView: View {
     
     @ObservedObject var threshold: SensorThreshold
+    var selectedStream: MeasurementStreamEntity?
```
pawelgil, line 14: we're trying to move away from using `CoreData` objects directly in `View`s. Lets come up with something properly separated (applies to all `View`s in this PR that introduce `MeasurementStreamEntity`)

AirCasting/SessionViews/MultiSliderView.swift
```diff
@@ -6,19 +6,24 @@
...
+            MultiSliderView(thresholds: threshold.rawThresholdsBinding, selectedStream: selectedStream)
+        }
     }
 }
 
 struct MultiSliderView: View {
-    
+    @InjectedObject private var userSettings: UserSettings
     @Binding var thresholds: [Float]
+    var selectedStream: MeasurementStreamEntity
```
pawelgil, line 26: same as above - let's not pull big dependencies into simple Views. They should stay simple and they should not perform this kind of logic.

pawelgil (changes requested):

AirCasting/Models/ThresholdFormatter.swift
```diff
@@ -0,0 +1,119 @@
...
+            return .celsius
+        } else {
+            return .fahrenheit
+        }
+    }
+
+    private func convertToInt(_ value: String) -> Int32 { Int32(value) ?? 0 }
+}
+
+
```
jankrzempek, line 119: One of those can ve deleted 😄 but it is like minor of minors

jankrzempek (approved):
😎🥳

AirCasting/Map/GoogleMapView.swift
```diff
@@ -166,8 +161,9 @@ struct GoogleMapView: UIViewRepresentable {
     }
     
     func color(point: PathPoint) -> UIColor {
-        guard let thresholds = threshold, let selectedStream = selectedStream else { return .white }
-        let measurement = selectedStream.isTemperature && userSettings.convertToCelsius ? Int32(TemperatureConverter.calculateFahrenheit(celsius: point.measurement)) : Int32(point.measurement)
+        guard let thresholds = threshold else { return .white }
+        let formatter = ThresholdFormatter(for: thresholds)
+        let measurement = formatter.formattedToFahrenheit(for: point)
```
pawelgil, line 166: good stuff, but the caller code should not decide to what it shall be formatting (the `Fahrenheit` part here).
Perfect-ish interface would be something like that:
```Swift
protocol ThresholdFormatter {
    func formattedValue(for: SensorThreshold) -> String
    func formattedUnit(for: SensorThreshold) -> String
}
```

Which will completely hide the fact that we're formatting temperature units under the hood, the view would only do something like:

```Swift
let formatter = Resolver.resolve(ThresholdFormatter.self, args: stream) //stream is of MeasurementStream type
let value = formatter.formattedValue(for: threshold)
let unit = formatter.formattedUnit(for: threshold)
```

See how now we have all formatting under control in a single place, if we wanted to add Kelvins next, we would only modify the formatter file/class without the need to visit all the places like here when we explicitly ask for "hey, format this value to fahrenheit". Instead we do it like "hey, give me a proper formatter for this stream" which is hiding all this.

pawelgil (changes requested):

AirCasting/Map/GoogleMapView.swift
SzymonGorowski: Thank you for your input, but honestly, I cannot find a way to solve this as you mentioned. 

The only thing that comes to my mind is to change function names to hide the fact that we format our values to exact temperature but I cannot find a way to change function parameters as you advised using a just `threshold`. Every function used a different parameter type like `Path Point`, `Double`, or `String`. 

Now If we want to add a new unit like Kelvins we will have to change only our file with formatter. - maybe I am wrong

AirCasting/Graph/StatisticsContainer/ViewModel/StatisticsContainerViewModel.swift
```diff
@@ -2,6 +2,7 @@
 //
 
 import Foundation
+import SwiftUI
```
SzymonGorowski, line 5: @pawelgil is this necessary there? It works without it.

jankrzempek (changes requested):
Unfortunately it is not working as it should I think. When I am trying to switch between C° and F° values on the threshold slider have not changes. See movie please: https://www.icloud.com/iclouddrive/0eeHrEgu6WoPLG24Knq2_p_zw#Szymon_slider

PR #694, opened 2022-05-05
Adding finish button to SearchMap

SzymonGorowski:
https://trello.com/c/iCpkJZW0/643-add-finish-button-in-the-search-map-view

### DIFF
--- a/AirCasting/CreateSessionViews/ChooseSessionType/ChooseSessionTypeView.swift
+++ b/AirCasting/CreateSessionViews/ChooseSessionType/ChooseSessionTypeView.swift
@@ -98,7 +98,11 @@ struct ChooseSessionTypeView: View {
                         viewModel.setSearchAndFollow(using: new)
                     })) {
                         CreatingSessionFlowRootView {
-                            SearchView()
+                            SearchView(isSearchAndFollowFlowActive: .init(get: {
+                                viewModel.isSearchAndFollowLinkActive
+                            }, set: { new in
+                                viewModel.setSearchAndFollow(using: new)
+                            }))
                         }
                     }
                     .fullScreenCover(isPresented: .init(get: {
@@ -204,7 +208,11 @@ struct ChooseSessionTypeView: View {
                                     viewModel.setSearchAndFollow(using: new)
                                 })) {
                                     CreatingSessionFlowRootView {
-                                        SearchView()
+                                        SearchView(isSearchAndFollowFlowActive: .init(get: {
+                                            viewModel.isSearchAndFollowLinkActive
+                                        }, set: { new in
+                                            viewModel.setSearchAndFollow(using: new)
+                                        }))
                                     }
                                 }
                         }
--- a/AirCasting/CreateSessionViews/ChooseSessionType/ChooseSessionTypeView.swift
+++ b/AirCasting/CreateSessionViews/ChooseSessionType/ChooseSessionTypeView.swift
@@ -98,7 +98,7 @@ struct ChooseSessionTypeView: View {
                         viewModel.setSearchAndFollow(using: new)
                     })) {
                         CreatingSessionFlowRootView {
-                            SearchView(isSearchAndFollowFlowActive: .init(get: {
+                            SearchView(isSearchAndFollowLinkActive: .init(get: {
                                 viewModel.isSearchAndFollowLinkActive
                             }, set: { new in
                                 viewModel.setSearchAndFollow(using: new)
@@ -208,7 +208,7 @@ struct ChooseSessionTypeView: View {
                                     viewModel.setSearchAndFollow(using: new)
                                 })) {
                                     CreatingSessionFlowRootView {
-                                        SearchView(isSearchAndFollowFlowActive: .init(get: {
+                                        SearchView(isSearchAndFollowLinkActive: .init(get: {
                                             viewModel.isSearchAndFollowLinkActive
                                         }, set: { new in
                                             viewModel.setSearchAndFollow(using: new)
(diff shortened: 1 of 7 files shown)


anna1901 (commented):
Doing it this way would result on putting the dashboard view *on top* of the search and follow views. That's not what we would want to happen. We want to dismiss the views that are on top of dashboard and move the app back there that way.

Here you can read about dismissing views: https://www.swiftbysundell.com/articles/dismissing-swiftui-modal-and-detail-views/

In the code there is already an `exploreSessionsButtonTapped` variable available so you can explore how it is used and probably use it in this `Finish` button logic as well.

AirCasting/SearchAndFollow/SearchMap/SearchMapViewModel.swift
```diff
@@ -34,7 +34,9 @@ class SearchMapViewModel: ObservableObject {
     @Published var cardPointerID: PointerValue = .noValue
     @Published var shouldCardsScroll: Bool = false
     private var currentPosition: GeoSquare?
-    
+    @Injected var persistenceController: PersistenceController
```
anna1901, line 37: We don't need those changes here, I think?

AirCasting/SearchAndFollow/SearchLocation/SearchView.swift
```diff
@@ -9,9 +9,11 @@ struct SearchView: View {
     @StateObject var viewModel: SearchViewModel
     @Environment(\.presentationMode) var presentationMode
     @EnvironmentObject private var exploreSessionsButton: ExploreSessionsButton
+    @Binding var isSearchAndFollowFlowActive: Bool
```
anna1901, line 12: Since we are not even using this binding in this view, I think it would be good to remove it from this view and instead pass it ti the viewModel in the initialiser and then retrieve it from the view model when creating the SearchMapView

anna1901 (approved):
Works well, good job 🙌🏻 I added just two comments about changes you may consider

AirCasting/SearchAndFollow/SearchMap/SearchMapView.swift
```diff
@@ -190,4 +198,22 @@ private extension SearchMapView {
...
     }
+    
+    var finishButton: some View {
+        Button {
+            isSearchAndFollowFlowActive = false
+            tabSelection.selection = .dashboard
+        } label: {
+            Text(Strings.SearchMapView.finishText)
+                .font(Fonts.muliHeading2.bold())
+                .padding(.trailing, 7)
```
jankrzempek, line 209: What would you say to align this well, according to the TextField below; maybe we could use `.padding(. trailing)` just after overlay?

AirCasting/SearchAndFollow/SearchMap/SearchMapView.swift
```diff
@@ -190,4 +198,22 @@ private extension SearchMapView {
...
+        }
+        .overlay(
+            Rectangle()
```
jankrzempek, line 212: There is `Capsule()` which you can use maybe instead and then delete .conrnerRadius? Just idea if not please ignore 😄

AirCasting/CreateSessionViews/ChooseSessionType/ChooseSessionTypeView.swift
```diff
@@ -204,7 +208,11 @@ struct ChooseSessionTypeView: View {
                                     viewModel.setSearchAndFollow(using: new)
                                 })) {
                                     CreatingSessionFlowRootView {
-                                        SearchView()
+                                        SearchView(isSearchAndFollowFlowActive: .init(get: {
+                                            viewModel.isSearchAndFollowLinkActive
```
jankrzempek, line 212: You are using many time `isSearchAndFollowFlowActive` and here you are passing  `viewModel.isSearchAndFollowLinkActive`. Could you please use one convention throughout the whole concept? (I would suggest `isSearchAndFollowLinkActive`) 😁

AirCasting/SearchAndFollow/SearchLocation/SearchViewModel.swift
```diff
@@ -29,7 +31,8 @@ class SearchViewModel: ObservableObject {
         return sensorType(from: selected.name)
     }
     
-    init() {
+    init(isSearchAndFollowFlowActive: Binding<Bool>) {
+        self.isSearchAndFollowFlowActive = isSearchAndFollowFlowActive
```
jankrzempek, line 35: I am not sure about this one, because then you need to import the SwiftUI to the VM which is quite not the concept that we are following. Using the SwiftUI in the AC app, we needed to use this 'hack' for dismissing, but it is totally a UI thing. Don't you think it should stay on the UI layer? Don't have strong opinion as I am not the master 😟

jankrzempek (commented):
Working like a charm 🥳 (few comments only)

anna1901 (approved):
Awesome!

AirCasting/SearchAndFollow/SearchMap/SearchMapView.swift
```diff
@@ -190,4 +198,22 @@ private extension SearchMapView {
...
+                .font(Fonts.muliHeading2.bold())
+                .padding(.trailing, 7)
+        }
+        .overlay(
+            Capsule()
+                .frame(width: 85, height: 35)
+                .foregroundColor(.accentColor)
+                .opacity(0.1)
+        )
+        .padding(.trailing, 10)
```
jankrzempek, line 217: The TextField which is below has a padding of 20. I would go here for 20 too, but make it please as you see it better

jankrzempek (approved):
Awesome helps a lot in this feature 😎

PR #695, opened 2022-05-05
Save external session to the database

anna1901:
https://trello.com/c/URrh3Lwc

* creating new external session entity
* save session to the database


anna1901:
[Saving session to database](https://trello.com/c/URrh3Lwc/599-saving-session-to-database)

AirCasting/SearchAndFollow/SearchMap/SearchMapViewModel.swift
```diff
@@ -113,26 +113,55 @@ class SearchMapViewModel: ObservableObject {
...
     
+    private func timeAsDate(_ time: String) -> Date {
+        let formatter = DateFormatters.SearchAndFollow.timeFormatter
+        let date = formatter.date(from: time)
+        guard let d = date else {
+            Log.error("Failed to convert start time received from API to date")
+            return DateBuilder.getFakeUTCDate()
+        }
+        return d
+    }
```
jankrzempek, line 163: This can go maybe to some extension?

jankrzempek (commented):
Looks like a giant hard work 🙌🏻

jankrzempek (commented):
I've tested the Following part as the user so works great. I was able to follow and then go back and try again and it seems to be working fine and remeber my state 🥳

AirCasting/SearchAndFollow/CompleteScreen/Contollers/SearchAndFollowCompleteScreenController.swift
```diff
@@ -0,0 +1,65 @@
+// Created by Lunar on 04/05/2022.
+//
+
+import Foundation
+import Resolver
+
+protocol SearchAndFollowCompleteScreenController {
```
pawelgil, line 7: well, the `Controller` realm is really reserved for high-level stuff, domain only. Here we have a mix:
* `download` suggests networking, not really a high level scenario (`retrieve` would fit tho)
* Usage of `StreamWithMeasurement` which is a service layer struct

AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift
```diff
@@ -138,60 +100,69 @@ class CompleteScreenViewModel: ObservableObject {
...
         exitRoute()
     }
     
-    private func getMeasurementsAndDisplayData(_ thresholds: ThresholdsValue) {
-        let streams = [String(session.streamID)] // In the future this will be changed for Airbeam sessions, cause we will need to get other streams ids from backend
-        downloadMeasurements(streams: streams) { [weak self] result in
+    private func getMeasurementsAndDisplayData() {
+        let streams = session.stream.map(\.id)
+        
+        controller.downloadMeasurements(streams: streams) { [weak self] result in
```
pawelgil, line 136: continuing with `Controller` purpose - a lot of this code here looks like application logic and it seems like it can be packed into a `Controller`. See how it first uses `downloadMeasurements` here and than after result is present it again calls the `Controller` with `createExternalSession`. Maybe it would be possible to refactor this VM so that most of this work gets done inside some single high level function inside the `Controller`? It will probably require some rethinking and restructuring here, but it would clear the field here a lot and make this ViewModel uncluttered (cuz right now a lot of code here is not really view-related and can be a struggle to get thru if someone just wanted to quickly get an idea how the UI works)

AirCasting/SearchAndFollow/CompleteScreen/Contollers/SearchAndFollowCompleteScreenController.swift
```diff
@@ -0,0 +1,65 @@
...
+                         sensorName: stream.sensorName,
+                         sensorPackageName: stream.sensorPackageName,
+                         thresholdsValues: stream.thresholdsValues,
+                         measurements: measurements.map {.init(value: $0.value, time: DateBuilder.getDateWithTimeIntervalSince1970(Double($0.time/1000)), latitude: $0.latitude, longitude: $0.longitude)})
+        })
+    }
+    
+    func downloadMeasurements(streams: [Int], completion: @escaping (Result<[StreamWithMeasurements], Error>) -> Void) {
+        guard !streams.isEmpty else {
+            completion(.failure(CompletionScreenError.noStreams))
```
pawelgil, line 42: should it stay as failure or maybe `.success` with empty array?

AirCasting/SearchAndFollow/CompleteScreen/Contollers/SearchAndFollowCompleteScreenController.swift
```diff
@@ -0,0 +1,65 @@
...
+            return
+        }
+        var results: [Result<StreamWithMeasurements, Error>] = []
```
pawelgil, line 45: as I already pointed it to Janek - just as a sidenote to raise awareness - this `results` is being accessed from multiple threads for writing. It means we have a data race possibility - probably the most common threading issue. Here it's very unlikely that this will happen because of the fact that it will get wriiten to as a result of networking calls which are very very slow in terms of CPU speeds. But worth noting

AirCasting/SearchAndFollow/CompleteScreen/Contollers/SearchAndFollowCompleteScreenController.swift
```diff
@@ -0,0 +1,65 @@
...
+            group.enter()
+            self.service.downloadStreamWithMeasurements(id: streamId, measurementsLimit: measurementsLimit) { results.append($0); group.leave() }
+        }
+        group.notify(queue: .global()) {
+            var allDownstreams = [StreamWithMeasurements]()
+            for result in results {
+                do {
+                    allDownstreams.append(try result.get())
+                } catch {
+                    completion(.failure(error))
```
pawelgil, line 58: note that this will only bubble up the first error encountered. Might be completely fine here

AirCasting/SearchAndFollow/CompleteScreen/Contollers/SearchAndFollowCompleteScreenController.swift
```diff
@@ -0,0 +1,65 @@
...
+        var results: [Result<StreamWithMeasurements, Error>] = []
+        let measurementsLimit = 60*24 // We want measurements from 24 hours
+        let group = DispatchGroup()
+        streams.forEach { streamId in
+            group.enter()
+            self.service.downloadStreamWithMeasurements(id: streamId, measurementsLimit: measurementsLimit) { results.append($0); group.leave() }
+        }
+        group.notify(queue: .global()) {
+            var allDownstreams = [StreamWithMeasurements]()
+            for result in results {
```
pawelgil, line 54: this works, but can be done in a more compact way using `.map`:
```Swift
do { try completion(.success(results.map { try $0.get() })) }
catch { completion(.failure(error)) }
```

also this looks like a possible extension to Collection for `Element == Result<T, E>`

AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift
```diff
@@ -138,60 +100,73 @@ class CompleteScreenViewModel: ObservableObject {
         exitRoute()
     }
     
+    func confirmationButtonPressed() {
+        following()
+        saveToDb()
+    }
+    
+    private func following() {
```
pawelgil, line 108: let's change the name to something else. Like `setFollowing()`?

AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift
```diff
@@ -138,60 +100,73 @@ class CompleteScreenViewModel: ObservableObject {
...
+    }
+    
+    private func saveToDb() {
+        guard let externalSessionWithStreams = externalSessionWithStreams else {
+            assertionFailure("Confirmation button pressed when there was no session with streams")
+            return
+        }
+        
+        do {
+            try externalSessionsStore.createExternalSession(session: externalSessionWithStreams)
```
pawelgil, line 121: since we have a `Controller` in place, the VM should not really handle storage/db/other services directly

AirCasting/SearchAndFollow/SearchMap/SearchMapViewModel.swift
pawelgil: side note: it's a bit weird we have to use this anyway. Why don't we use a proper formatter for `JSONDecoder` on a service level and produce `Date` instead of `String`? Take a look at [docs](https://developer.apple.com/documentation/foundation/jsondecoder/2895216-datedecodingstrategy)

pawelgil (changes requested):
Looks good! I'd make some code quality changes

AirCasting/SearchAndFollow/CompleteScreen/Contollers/SearchAndFollowCompleteScreenController.swift
anna1901: So should it be called service?

AirCasting/CoreData/ExternalSessionsStore.swift
```diff
@@ -0,0 +1,87 @@
...
+            throw CreatingExternalSessionError.sessionAlreadyExists
+        }
+        
+        let sessionEntity = newSessionEntity()
+        updateSessionsParams(sessionEntity, session: session)
+        session.streams.forEach { stream in
+            addStream(stream, to: sessionEntity)
+        }
+        
+        // TODO: If thresholds don't exist in the db already, create the thresholdsEntity
```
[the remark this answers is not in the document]
anna1901: Yeah, but it is something that should be done in the next ticket, in a very near future. Should it still be added to issues? I thought we are using those TODOs and issues for marking code that needs to be refactored "at some point"

PR #696, opened 2022-05-05
Fix location reload on S&F feature

jankrzempek:
Execute code only when needed, after map idle is detected. Deleted what was not needed for map and search reload implementation - dismissView function for modal sheet.

+ The code for reloading dots on the map and refreshing the views was executing many times, now it should be improved.

https://trello.com/c/czdkWCr0/642-sf-map-after-entering-new-location-map-pans-to-new-location-but-results-arent-updated


pawelgil (approved):
LGTM, but I won't be able to test it now, so not assigning

PR #697, opened 2022-05-10
Thresholds refreshing color bug fix

SzymonGorowski:
https://trello.com/c/BsnTaVEi/644-frame-and-marked-spot-on-the-map-doesnt-refresh


AirCasting/SessionViews/SingleMeasurementView.swift
```diff
@@ -7,7 +7,7 @@ import Resolver
 
 struct SingleMeasurementView: View {
     @ObservedObject var stream: MeasurementStreamEntity
-    var threshold: SensorThreshold?
+    @ObservedObject var threshold: SensorThreshold
```
jankrzempek, line 10: 2/2 You have deleted the optional and now it cannot be nil. When you look below in this code, there is a possibility that it can be (`if let threshold = threshold, measurementPresentationStyle == .showValues `{), we should stay with that approach as I believe it was done intentionally by someone

AirCasting/SessionViews/ABMeasurementsView.swift
```diff
@@ -110,7 +117,7 @@ struct _ABMeasurementsView: View {
                 streamsToShow.count != 1 ? Spacer() : nil
                 ForEach(streamsToShow.filter({ !$0.gotDeleted }), id : \.self) { stream in
                     SingleMeasurementView(stream: stream,
-                                          threshold: nil,
+                                          threshold: .init(value: nil),
```
jankrzempek, line 120: Here you don't need to define the value as nil. The reason is the init so:

```
 init(value: SensorThreshold? = nil) {
         self.value = value
     }
```
You have a default value that is the same as the one you are assigning here.

AirCasting/SessionViews/SingleMeasurementView.swift
```diff
@@ -5,9 +5,17 @@ import SwiftUI
...
 import Resolver
 
+class SingleMeasurementViewThreshold: ObservableObject {
+    var value: SensorThreshold? = nil
+    
+    init(value: SensorThreshold? = nil) {
+        self.value = value
+    }
+}
+
```
jankrzempek, line 15: You are double nil'ing here. First, you define the value as optional nil. After that, we will call the init and assing /default value/ from init to our `var value` again. If you need more explanation please let me know 😄

jankrzempek (commented):
Minor improvements in the comments

anna1901 (approved):

AirCasting/SessionViews/ABMeasurementsView.swift
```diff
@@ -6,11 +6,19 @@ enum MeasurementPresentationStyle {
     case hideValues
 }
 
+class ABMeasurementsViewThreshold: ObservableObject {
```
pawelgil, line 9: Why this class was needed instead of changing view properties to a `@State`? (applies to the `SingleMeasurementViewThreshold` too)

pawelgil (approved):
It's peculiar why those new observable classes are needed, but let's not hold back, so +2

PR #698, opened 2022-05-10
S&F Chart: rounded values and proper timestamp

jankrzempek:
Measurements value are now rounded, and chart shows the proper time.
https://trello.com/c/IbhLI7o8/618-implement-averaging-in-the-chart

_Based on:_
-values are different
-same values do not share the same vertical position
-timestamps are off


AirCasting/SearchAndFollow/CompleteScreen/Views/SearchAndFollowChartViewModel.swift
```diff
@@ -21,8 +21,19 @@ class SearchAndFollowChartViewModel: ObservableObject {
...
+        // Result of this, is the chart that is the same; as on the session card.
+        guard let firstElement = updatedMeasurements.reversed().first?.time else { return (nil, nil) }
+        let hoursForRemoval = Calendar.current.component(.hour, from: firstElement)
+        for (index, item) in updatedMeasurements.enumerated().reversed() {
+            if Calendar.current.component(.hour, from: item.time) == hoursForRemoval {
+                updatedMeasurements.remove(at: index)
+            }
+        }
+        
+        for measurement in updatedMeasurements.reversed() {
```
anna1901, line 36: What do you think about adding this check `if Calendar.current.component(.hour, from: item.time) == hoursForRemoval` below this line inside this loop, and if it is true then continuing to the next measurement with `continue`? It will remove this additional loop through all of the measurements.

anna1901 (changes requested):

AirCasting/SearchAndFollow/CompleteScreen/Views/SearchAndFollowChartViewModel.swift
jankrzempek: Really like the idea. Changed!

anna1901: I meant that we can do it like this:
Screenshot 2022-05-12 at 10 22 44

However, this only works well for Airbeam sessions, because for OpenAQ data we often don't have measurement from most recent hour and for example now, when there is 4:00 in New York, the lates measurement for one session is from 1:00. So we actually want to compare the hour of the measurement to the current time.

anna1901 (changes requested):

AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift
```diff
@@ -172,13 +172,23 @@ class CompleteScreenViewModel: ObservableObject {
...
                 }
             }
         }
     }
     
+    private func componentsSeparation(name: String) -> String {
+        if name.contains(":") {
+            let value = name.components(separatedBy: ":").first!
+            return value.components(separatedBy: CharacterSet.decimalDigits).joined()
+        } else {
```
pawelgil, line 186: else is not needed because you're returning from `if`

AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift
```diff
@@ -172,13 +172,23 @@ class CompleteScreenViewModel: ObservableObject {
                     if let stream = self.externalSessionWithStreams!.streams.first {
                         self.selectedStream = stream.id
                         self.selectedStreamUnitSymbol = stream.unitSymbol
-                        (self.chartStartTime, self.chartEndTime) = self.chartViewModel.generateEntries(with: stream.measurements.map({ SearchAndFollowChartViewModel.ChartMeasurement(value: $0.value, time: $0.time) }), thresholds: stream.thresholdsValues)
+                        (self.chartStartTime, self.chartEndTime) = self.chartViewModel.generateEntries(with: stream.measurements.map({ SearchAndFollowChartViewModel.ChartMeasurement(value: $0.value, time: $0.time) }), thresholds: stream.thresholdsValues, basedOn: self.componentsSeparation(name: stream.sensorName))
```
pawelgil, line 175: this callsite is not clear to me. I don't know what `basedOn` is. The reason for that is that `componentsSeparation` doesn't make sens in context of `basedOn`. Maybe rename `componentsSeparation` to something telling us that it returns a proper sensorname?

AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift
```diff
@@ -172,13 +172,23 @@ class CompleteScreenViewModel: ObservableObject {
...
             }
         }
     }
     
+    private func componentsSeparation(name: String) -> String {
+        if name.contains(":") {
+            let value = name.components(separatedBy: ":").first!
+            return value.components(separatedBy: CharacterSet.decimalDigits).joined()
+        } else {
+            let value = name.components(separatedBy: "-").first!
```
pawelgil, line 187: that's a bit dangerous. I believe that `stream.sensorName` is provided externally (so by the API or a user or smth we don't fully control). This is why we cannot make that assumption that if it doesn't contain `:` it will 100% sure contain `-`. I think we need to be prepared to fail here. Please implement a fallback or error handling

AirCasting/SearchAndFollow/CompleteScreen/Views/SearchAndFollowChartViewModel.swift
```diff
@@ -9,20 +9,47 @@ class SearchAndFollowChartViewModel: ObservableObject {
...
         var times: [Date] = []
         var buffer: [ChartMeasurement] = []
+        var updatedMeasurements = measurements
+        
+        let sensor = getEntriesSensor(using: sensorName)
+        if sensor != .OpenAQ {
+            guard let firstElement = updatedMeasurements.reversed().first?.time else { return (nil, nil) }
+            
+            switch sensor {
+            case .AirBeam:
```
pawelgil, line 39: I think we should be using polymorphism instead of passing `sensorName` in and doing that by switching thru possibilities in here.

In other words:
Change this code so that if we ever add another sensor type this whole file won't need to be modified at all (this is called open-closed principle and should be applied here)

pawelgil (changes requested):

AirCasting/SearchAndFollow/CompleteScreen/Views/ChartSensor.swift
```diff
@@ -0,0 +1,69 @@
+// Created by Lunar on 23/05/2022.
+//
+
+import Foundation
+import SwiftUI
+
+protocol ChartSensor {
```
pawelgil, line 7: Not sure this name fits well. Maybe let's rename to `ChartMeasurementsFilter`?

and the `func clean` to func `func filter`?

What do you think?

AirCasting/SearchAndFollow/CompleteScreen/Views/SearchAndFollowChartViewModel.swift
```diff
@@ -5,24 +5,29 @@ import Foundation
...
+        let value: Double
+        let color: Color
+    }
+    
     @Published var entries: [ChartDot] = []
     
     let numberOfEntries = 9
     
-    func generateEntries(with measurements: [ChartMeasurement], thresholds: ThresholdsValue) -> (Date?, Date?) {
+    func generateEntries(with measurements: [ChartMeasurement], thresholds: ThresholdsValue, basedOn sensor: ChartSensor) -> (Date?, Date?) {
```
pawelgil, line 23: `basedOn` -> `using` ?

Maybe also make it optional, so the filter is not required?

pawelgil (changes requested):
Pretty good, let's polish things up a bit

AirCasting/SearchAndFollow/CompleteScreen/Views/ChartSensor.swift
```diff
@@ -0,0 +1,69 @@
...
+            Log.info("Missing sensor name in the chart VM")
+            return
+        }
+    }
+    
+    private func hoursForRemoval(using element: Date) -> Int { Calendar.current.component(.hour, from: element) }
+    
+    private func clearPAData(using updatedMeasurements: inout [MeasurementType], hourToRemove: Date) {
+        for (index, item) in updatedMeasurements.enumerated().reversed() {
+            if Calendar.current.component(.hour, from: item.time) == hoursForRemoval(using: hourToRemove) && Calendar.current.component(.minute, from: item.time) != 00 {
```
anna1901, line 45: Why we don't want to remove the measurement from the full hour?

anna1901 (changes requested):
I don't like that we have this iteration through the whole measurements array to remove those from last hour. I think it is an overkill, especially for purple air measurements, when we could be iterating through all 1440 measurements to remove just the last one.The solution I was thinking about was to, for example, add a code like this

```
if sensor.name == "Airbeam3" {
                // skip measurement if it's from the last hour
            } else if sensor.name == "PurpleAir" {
                // skip measurement if it's from the last hour
            }
```

to generateEntries method. What do you think about that? Have you considered that kind of solution?

AirCasting/SearchAndFollow/CompleteScreen/Views/SearchAndFollowChartViewModel.swift
```diff
@@ -4,50 +4,27 @@
...
-            case .PurpleAir:
-                // PA stands for PurpleAir
-                clearPAData(using: &updatedMeasurements, hourToRemove: firstElement)
-            case .undefined:
-                Log.info("Missing sensor name in the chart VM")
-            default:
-                return (nil, nil)
-            }
-        }
+        sensor.clean(measurements: &updatedMeasurements)
```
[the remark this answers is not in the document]
jankrzempek: Why those should be avoided, I am just wondering?

AirCasting/SearchAndFollow/CompleteScreen/Views/SearchAndFollowChartViewModel.swift
jankrzempek: I think that it is a lot of work needed to make it done. For now, when we are not filtering we are just going to return those measurements that we have passed. I would like to leave it at least for now as the time goes by. What would u say?

AirCasting/SearchAndFollow/CompleteScreen/Views/ChartSensor.swift
jankrzempek: As I remember we have decided that using PurpleAir we want to have ex. those from 9:00 and exclude only those from above

anna1901 (changes requested):
I can't test because of an issue with xcode but I think there is a mistake to fix.

PR #706, opened 2022-05-13
Save external sessions thresholds

anna1901:
https://trello.com/c/qQGFM77X


anna1901:
[Save thresholds values on session following if they aren't already stored in the db](https://trello.com/c/qQGFM77X/648-save-thresholds-values-on-session-following-if-they-arent-already-stored-in-the-db)

PR #708, opened 2022-05-15
MicrophoneManager refactor

pawelgil:


pawelgil:
possible regression of [Interrupted audio](https://trello.com/c/7M33g3AR/334-microphone-doesnt-resume-recording-when-other-app-was-using-a-microphone-and-then-we-came-back-to-the-app) - tested, doesn't seem to occur

AirCasting/MicrophoneSession/LevelMeasurementController/DecibelMeasurementSaveable.swift
```diff
@@ -0,0 +1,85 @@
...
+                                       thresholdVeryHigh: 100,
+                                       thresholdHigh: 80,
+                                       thresholdMedium: 70,
+                                       thresholdLow: 60,
+                                       thresholdVeryLow: 20)
+        
+        persistence.accessStorage { [weak self] storage in
+            do {
+                guard let self = self else { return }
+                _ = try storage.createSessionAndMeasurementStream(self.session, stream)
```
jankrzempek, line 78: Maybe this `storage.createSessionAndMeasurementStream(self.session, stream)` could not give us anything back so we could delete `_ =`? It is only used here as I can see

AirCasting/MicrophoneSession/LevelMeasurementController/DecibelSampler.swift
```diff
@@ -0,0 +1,37 @@
...
+        do {
+            try microphone.stopRecording()
+        } catch {
+            Log.error("Couldn't stop recording: \(error)")
+        }
+    }
+    
+    enum DecibelSamplerError: Error {
+        case couldntGetMicrophoneLevel
+    }
```
jankrzempek, line 36: Shouldn't enum be at the top?

AirCastingTests/Logic/MicrophoneSessions/LevelMeasurementControllerTests.swift
```diff
@@ -0,0 +1,164 @@
...
+        }
+        
+        func schedule(every: TimeInterval, closure: @escaping () -> Void) -> AnyObject {
+            self.closure = closure
+            return Token()
+        }
+        
+        func stop(token: AnyObject) {
+            
+        }
```
jankrzempek, line 133: Could you maybe make it one liner?

jankrzempek (approved):
Tested, looks awesome 🎉

AirCasting/MicrophoneSession/LevelMeasurementController/DecibelMeasurementSaveable.swift
```diff
@@ -0,0 +1,85 @@
+// Created by Lunar on 15/05/2022.
+//
+
+import Foundation
+import Resolver
+
+class DecibelMeasurementSaveable: MeasurementSaveable {
```
SzymonGorowski, line 7: I think this class could be `final`

AirCasting/MicrophoneSession/LevelMeasurementController/DecibelSampler.swift
```diff
@@ -0,0 +1,37 @@
+// Created by Lunar on 15/05/2022.
+//
+
+import Foundation
+
+class DecibelSampler: LevelSampler {
```
SzymonGorowski, line 6: I think this class could be `final`

SzymonGorowski (approved):
👏

AirCasting/MicrophoneSession/LevelMeasurementController/Interfaces/LevelSampler.swift
```diff
@@ -0,0 +1,15 @@
...
+/// Whenever a sampler disconnects this error type is returned
+struct LevelSamplerDisconnectedError: Error { }
+
+/// Interface for objects capable of taking a single measurement
+protocol LevelSampler {
+    associatedtype Measurement
+    /// Takes a single measurement of the `Measurement` type
+    /// - Parameter completion: completion closure called when the measurement is taken or an error is produced
+    func sample(completion: (Result<Measurement, Error>) -> Void)
+}
```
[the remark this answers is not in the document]
pawelgil: As always - it depends. We should always have in mind the perspective of future developers (most likely ourselves in the short term) which will be reading this file. The natural way of reading code is top-down, so we want all prerequisites and assertions on the very top (usually). I think this error type is something we really want to emphasize as it's pretty important for the rest of this module.

So in this form I think the error should be higher than the protocol. But that's a good find! As it's kind of smelly. I'll change it so that the protocol returns a typed error instead of a generic (`Error`).

AirCasting/MicrophoneSession/LevelMeasurementController/DecibelSampler.swift
pawelgil: There is no rule - we should judge ourselves what will be the best for this files readability. I think you're right here, I'll move it.

AirCasting/MicrophoneSession/LevelMeasurementController/DecibelSampler.swift
pawelgil: indeed

AirCasting/MicrophoneSession/LevelMeasurementController/DecibelMeasurementSaveable.swift
pawelgil: indeed

AirCasting/Utils/Microphone/AVMicrophone.swift
```diff
@@ -0,0 +1,96 @@
...
+        self.recorder = try AVAudioRecorder(url: URL(fileURLWithPath: "/dev/null", isDirectory: true), settings: AVMicrophone.recordSettings)
+        self.interruptionHandler = .init(observer: self, audioSession: audioSession)
+    }
+    
+    func getCurrentDecibelLevel() -> Double? {
+        recorder.updateMeters()
+        let power = recorder.averagePower(forChannel: 0)
+        var decibels = Double(power + 90.0)
+        (decibels < 0) ? decibels = 0 : nil
+        return decibels
```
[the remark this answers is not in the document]
pawelgil: Oh, great catch!

I'm thinking what would be correct on a domain level. Can microphone read `no value`? What if we try to get current level when the microphone is stopped? Should we `nil`? Should we throw? Should we crash? Or just pretend nothing's happening and return `0`?

I'll think about it, if you have any thoughts please share!

AppDelegate+Injection.swift
```diff
@@ -238,3 +245,13 @@ extension Resolver: ResolverRegistering {
         func isFeatureOn(_ feature: FeatureFlag) -> Bool? { true }
     }
 }
+
+// TODO: Remove when we move to Sessionable?
```
pawelgil, line 249: note to self: investigate 🕵️‍♂️

AirCasting/Utils/Microphone/AVMicrophone.swift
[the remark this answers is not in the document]
jankrzempek: The only thought I have right now is not to use `0` when the mic is stopped. I think it would mislead us, as the 0 is a numeric value that could potentially happen during the standard process of recording (possibly not, but who knows)

AirCasting/MicrophoneSession/LevelMeasurementController/LevelMeasurementController.swift
```diff
@@ -0,0 +1,46 @@
+// Created by Lunar on 15/05/2022.
+//
+
+import Foundation
+import CoreLocation
+import Resolver
```
anna1901, line 6: I think we don't need Resover here

pawelgil:
We need to retest this. @jankrzempek can you help with that?

jankrzempek:
> We need to retest this. @jankrzempek can you help with that?

Sure, doing!

jankrzempek (changes requested):
It needs LocationService for Injection and as I can see there is no default implementation for this protocol. Attaching error when trying to record mobile mic session 👾
Zrzut ekranu 2022-07-26 o 13 05 33

pawelgil:
@jankrzempek missing dependency was added, should not be crashing anymore

jankrzempek (approved):
Pimped up 🤌🏻

AirCasting/Settings/SettingsViewModel.swift
```diff
@@ -17,6 +17,7 @@ class SettingsViewModel: ObservableObject {
     @Injected private var urlProvider: URLProvider
     @Injected private var locationAuthorization: LocationAuthorization
     @Injected private var bluetoothHandler: BluetoothHandler
+    @Injected private var microphone: Microphone
```
anna1901, line 20: Why?

anna1901 (changes requested):
Recording is not resuming correctly after an interruption (like incoming phone call).

AirCasting/MicrophoneSession/MicrophoneManager.swift
```diff
@@ -1,217 +1,22 @@
...
-        do {
-            try recorder = AVAudioRecorder(url: URL(fileURLWithPath: "/dev/null", isDirectory: true), settings: MicrophoneManager.recordSettings)
-        } catch {
-            Log.error("Could not create AVAudioRecorder: \(error.localizedDescription)")
-        }
-    }
+final class MicrophoneManager {
+    @Injected private var microphone: Microphone
+    
+    private var controller: Any?
```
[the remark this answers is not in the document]
pawelgil: the only reason we need this `var` is to keep the controller alive, no need to provide a type really.

AirCasting/MicrophoneSession/LevelMeasurementController/DecibelMeasurementSaveable.swift
```diff
@@ -0,0 +1,85 @@
...
+    
+    func saveMeasurement(_ value: Double, completion: @escaping (Result<Void, Error>) -> Void) {
+        guard databasePrepared else { return }
+        do {
+            let location = session.locationless ? nil : try locationService.getCurrentLocation()
+            
+            persistence.accessStorage { [weak self] storage in
+                do {
+                    guard let session = self?.session else { return }
+                    guard let streamID = try storage.existingMeasurementStream(session.uuid, name: Constants.SensorName.microphone) else {
```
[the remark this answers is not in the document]
pawelgil: we could, but this won't optimize anything really. CoreData is not an SQL database (it only uses it as a persistent store), it doesn't fetch the stream from database here, no need to cache it at all.

AirCasting/MicrophoneSession/LevelMeasurementController/DecibelMeasurementSaveable.swift
```diff
@@ -0,0 +1,85 @@
...
+                try storage.updateSessionStatus(.DISCONNECTED, for: self.session.uuid)
+                Log.info("[DEBUG] Disconnected \(self.session.name ?? "Unknown") [\(self.session.uuid)]")
+            } catch {
+                Log.error("Couldn't disconnect \(self.session.name ?? "Unknown") [\(self.session.uuid)]! \(error.localizedDescription)")
+            }
+        }
+    }
+    
+    private func prepareData(completion: @escaping (Result<Void, Error>) -> Void) {
+        let stream = MeasurementStream(id: nil,
```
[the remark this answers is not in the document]
pawelgil: I can't, this was moved as-is from the original implementation

AirCasting/MicrophoneSession/LevelMeasurementController/DecibelMeasurementSaveable.swift
```diff
@@ -0,0 +1,85 @@
...
+                                       sensorName: Constants.SensorName.microphone,
+                                       sensorPackageName: "Builtin",
```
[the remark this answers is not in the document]
pawelgil: Indeed, or even an enum perhaps 🤔. We can add such task to trello since it's out of scope of this PR

AirCasting/MicrophoneSession/LevelMeasurementController/Adapters/LocationServiceAdapter.swift
```diff
@@ -0,0 +1,23 @@
...
+    init(tracker: LocationTracker) {
+        self.tracker = tracker
+        tracker.start()
+    }
+    
+    deinit {
+        tracker.stop()
+    }
+    
+    func getCurrentLocation() throws -> CLLocationCoordinate2D? {
```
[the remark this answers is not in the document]
pawelgil: it's not, but the `LocationService` requires this

AirCasting/Map/LocationTracker.swift
```diff
@@ -45,6 +45,7 @@ final class CoreLocationTracker: NSObject, LocationTracker, LocationAuthorizatio
             locationManager.startUpdatingLocation()
         }
         locationStartReference += 1
+        Log.info("Started location tracking (refcount: \(locationStartReference))")
```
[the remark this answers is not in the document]
pawelgil: It shouldn't flood us, and I think it might be helpful to see. In particular the `refcount` here can be a good debugging tool (we can track which actions triggered location tracking and if each of them has a corresponding stopTracking call)

pawelgil:
@anna1901 the issue should be resolved now

AirCasting/Utils/Microphone/AVMicrophone.swift
```diff
@@ -65,7 +65,7 @@ extension AVMicrophone: AVSessionInterruptionObserver {
     }
     
     func onAVSessionInteruptionEnd(shouldResume: Bool) {
-        Log.info("audio recorder interruption ended (should resume? \(shouldResume ? "yes." : "no.")")
+        Log.info("audio recorder interruption ended (should resume? \(shouldResume ? "yes." : "no.)")")
```
anna1901, line 68: One closing bracket missing or two brackets too much :)

anna1901 (approved):
Cool!! 🚀

PR #711, opened 2022-05-17
Swiping between dashboard view tabs

SzymonGorowski:
https://trello.com/c/rLm4dgXf/385-dashboard-swipe-to-navigate-between-tabs

https://trello.com/c/97LZ5vcz/303-scrolling-divide-list-into-separated-lists

This PR solves the two above tickets.


AirCasting/Dashboard/DashboardView.swift
```diff
@@ -14,15 +14,16 @@ import Resolver
...
-    @EnvironmentObject var selectedSection: SelectSection
+    // TODO: - We can remove it in a future
+    @EnvironmentObject var selectedSection_: SelectSection
     @EnvironmentObject var reorderButton: ReorderButton
     @EnvironmentObject var searchAndFollowButton: SearchAndFollowButton
     @State var isRefreshing: Bool = false
+    @State var selectedSection: SelectedSection = .following
     @Injected private var sessionSynchronizer: SessionSynchronizer
+    @Injected private var persistenceController: PersistenceController
     @InjectedObject private var featureFlagsViewModel: FeatureFlagsViewModel
```
pawelgil, line 25: Is that still needed?

AirCasting/Dashboard/SessionListView.swift
```diff
@@ -0,0 +1,89 @@
...
+import CoreData
+
+struct SessionsListView: View {
+    @FetchRequest<SensorThreshold>(sortDescriptors: [.init(key: "sensorName", ascending: true)]) private var thresholds
+    @StateObject private var coreDataHook: CoreDataHook
+    @Binding var isRefreshing: Bool
+    @InjectedObject private var featureFlagsViewModel: FeatureFlagsViewModel
+    
+    private let listCoordinateSpaceName = "listCoordinateSpace"
+    let selectedSection: SelectedSection
```
pawelgil, line 20: I think this can be `private`?

pawelgil (changes requested):
Minor improvements can be done

AirCasting/Dashboard/DashboardView.swift
SzymonGorowski: I think it is needed because we use it for init SessionListView.

Screenshot 2022-06-10 at 07 41 41

AirCasting/Dashboard/DashboardView.swift
```diff
@@ -14,15 +14,16 @@ import Resolver
 struct DashboardView: View {
     @StateObject var coreDataHook: CoreDataHook
     @FetchRequest<SensorThreshold>(sortDescriptors: [.init(key: "sensorName", ascending: true)]) var thresholds
-    @EnvironmentObject var selectedSection: SelectSection
+    // TODO: - We can remove it in a future
```
anna1901, line 17: Could you be more specific in this comment? When exactly can this be removed?

SzymonGorowski: Pawel told me that in the future we can rethink the way how it's working and change it.

anna1901: In what future? If we will not do it in the next few days, it would be good to change this comment, so that it states specifically what can be removed and when. What needs to happen for this to be removed? We already have some cryptic warning messages that we don't know what they mean.

anna1901 (approved):
Works great! 🎉

AirCasting/Dashboard/DashboardView.swift
```diff
@@ -56,10 +62,12 @@ struct DashboardView: View {
             }
         }
         .navigationBarTitle(Strings.DashboardView.dashboardText)
-        .onChange(of: selectedSection.selectedSection) { selectedSection in
-            self.selectedSection.selectedSection = selectedSection
-            try! coreDataHook.setup(selectedSection: self.selectedSection.selectedSection)
+        .onChange(of: selectedSection_.selectedSection) { selectedSection in
+            self.selectedSection_.selectedSection = selectedSection
```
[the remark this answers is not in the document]
SzymonGorowski: I think it's not necessary. I checked and it's working without it.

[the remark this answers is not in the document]
SzymonGorowski: I think it's not necessary. I checked and it's working without it.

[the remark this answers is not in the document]
anna1901: I think this `selectedSection_` is used in other views, to set the selected section to following or active when, for example, you finish creating session. Have you checked if that still works correctly? So this needs to stay in my opinion. It's an old code that is used in the app, so we need to be very careful with removing it.

pawelgil (approved):

AirCasting/Dashboard/DashboardView.swift
```diff
@@ -14,15 +14,15 @@ import Resolver
 struct DashboardView: View {
     @StateObject var coreDataHook: CoreDataHook
     @FetchRequest<SensorThreshold>(sortDescriptors: [.init(key: "sensorName", ascending: true)]) var thresholds
-    @EnvironmentObject var selectedSection: SelectSection
+    // TODO: - We can rethink the way how "Select Section" works and change it
+    @EnvironmentObject var selectedSection_: SelectSection
```
anna1901, line 18: I would rather name the new selectedSection a "selectedSectionForSwipingImplementation" or something like that because it is a confusing why there are two

PR #713, opened 2022-05-19
Sessionable protocol

anna1901:
We are adding Sessionable protocol to hide SessionEntity and ExternalSessionEntity behind it
https://trello.com/c/DVPSiTp7


AirCasting/CoreData/ExternalSessionEntity+CoreDataProperties.swift
```diff
@@ -49,3 +66,89 @@ extension ExternalSessionEntity {
...
 }
+
+extension ExternalSessionEntity {
+
+    var pm1Stream: MeasurementStreamEntity? {
+        let match = measurementStreams?.first(where: { (stream) -> Bool in
+            (stream as? MeasurementStreamEntity)?.sensorName == "AirBeam3-PM1" ||
+                (stream as? MeasurementStreamEntity)?.sensorName == "AirBeam2-PM1" ||
+                    (stream as? MeasurementStreamEntity)?.sensorName == "AirBeam1-PM1"
+        })
```
jankrzempek, line 77: Why we are not using the same approach as below to find those that `.contains("PM1)`?

AirCasting/Dashboard/Reordering/ReorderingDashboardViewModel.swift
```diff
@@ -20,6 +20,7 @@ class ReorderingDashboardViewModel: ObservableObject {
     func finish() {
         measurementStreamStorage.accessStorage { storage in
             self.sessions.reversed().enumerated().forEach { index, session in
+                // TODO: implement new logic for saving new sessions order
```
jankrzempek, line 23: Maybe we could add also a warning here that it can be easily find later?

AirCasting/CoreData/ExternalSessionEntity+CoreDataProperties.swift
anna1901: Because PM10 also contains PM1

jankrzempek (approved):

AirCasting/CoreData/ExternalSessionEntity+CoreDataProperties.swift
```diff
@@ -49,3 +66,89 @@ extension ExternalSessionEntity {
...
+        let pm1Stream = match as? MeasurementStreamEntity
+        return pm1Stream
+    }
+
+    var pm2Stream: MeasurementStreamEntity? {
+        let match = measurementStreams?.first(where: { (stream) -> Bool in
```
pawelgil, line 83: we should extract the boy of this `var`s into a function that takes a string as a param because all of those vars are exactly the same except the string.

```Swift
var pm2Stream: MeasurementStreamEntity? { streamWithSensorName(containing: "PM2.5") }
```

AirCasting/CoreData/Storages/UIStorage.swift
```diff
@@ -9,8 +9,8 @@ protocol UIStorage {
 }
 
 protocol UIStorageContextUpdate {
-    func cardStateToggle(for sessionUUID: SessionUUID) throws
-    func changeStream(for sessionUUID: SessionUUID, stream: String) throws
+    func cardStateToggle(for sessionUUID: SessionUUID, isSessionExternal: Bool) throws
```
pawelgil, line 12: I don't like that we need to pass `sessoinExternal` in here. It should be handled internally IMO. `SessionUUID` should be unique across all sessions

AirCasting/CoreData/Storages/UIStorage.swift
```diff
@@ -9,8 +9,8 @@ protocol UIStorage {
...
+    func changeStream(for sessionUUID: SessionUUID, stream: String, isSessionExternal: Bool) throws
```
pawelgil, line 13: I don't like that we need to pass `sessoinExternal` in here. It should be handled internally IMO. `SessionUUID` should be unique across all sessions

AirCasting/Services/RemoveOldMeasurementsService.swift
```diff
@@ -16,14 +16,13 @@ final class RemoveOldMeasurementsService {
     
     /// In fixed sessions we need to remove measurements older than 24h, but we treat 24 not like a date, but as a sum of measurements.
     /// We know that we have 60 measurements per hour, so we take 1440 last measurements for each stream, and we remove older than the first of them.
-    func removeOldestMeasurements(in context: NSManagedObjectContext, from sessionUUID: SessionUUID, of type: DownloadMeasurementsService.Session.SessionType) throws {
-        switch type {
-        case .regular:
+    func removeOldestMeasurements(in context: NSManagedObjectContext, from sessionUUID: SessionUUID, which isExternal: Bool) throws {
```
pawelgil, line 19: passing flag params like `isXYZ` to functions is usually a codesmell, maybe we can avoid that in this PR?

AirCasting/UIState/UIStateHandler.swift
```diff
@@ -5,27 +5,27 @@ import Foundation
 import Resolver
 
 protocol SessionCardUIStateHandler {
-    func toggleCardExpanded(sessionUUID: SessionUUID)
-    func changeSelectedStream(sessionUUID: SessionUUID, newStream: String)
+    func toggleCardExpanded(sessionUUID: SessionUUID, isSessionExternal: Bool)
+    func changeSelectedStream(sessionUUID: SessionUUID, newStream: String, isSessionExternal: Bool)
```
pawelgil, line 9: same as in other places: we should avoid flag arguments

pawelgil (changes requested):

PR #714, opened 2022-05-20
Five streams searchfollow

jankrzempek:
https://trello.com/c/li6uFGrE/654-show-all-5-streams-for-airbeam-sessions


AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift
```diff
@@ -141,6 +142,43 @@ class CompleteScreenViewModel: ObservableObject {
         exitRoute()
     }
     
+    private func getAndDisplayData() {
+        if session.stream.first!.sensorName.contains("AirBeam") {
+            do {
+                try streamsDownloader.downloadStreams(with: session.id, for: .airBeam3) { result in
```
anna1901, line 148: Are you sure this should always be `.airbeam3`? Isn't it possible for the user to follow an Airbeam2 session?

AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreen.swift
```diff
@@ -39,7 +39,10 @@ struct CompleteScreen: View {
             if viewModel.isMapSelected {
                 SearchCompleteScreenMapView(longitude: viewModel.sessionLongitude, latitude: viewModel.sessionLatitude)
             } else {
-                SearchAndFollowChartView(viewModel: viewModel.chartViewModel)
+                SearchAndFollowChartView(viewModel: viewModel.chartViewModel, streamID: $viewModel.selectedStream)
```
anna1901, line 42: Why do we need that binding for streamID? And do we need to add this `onStreamChange` method at all? Maybe we could just call `generateEntries` function on chartViewModel from the ComleteScreenViewModel instead?

AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift
```diff
@@ -141,6 +142,43 @@ class CompleteScreenViewModel: ObservableObject {
...
+                    switch result {
+                    case .success(let success):
+                        // TODO: - FIX Thresholds and Improve this part of code
```
anna1901, line 151: Is this PR ready for review with this Thresholds logic that still needs to be added? I think we may need to think of another approach for this task, like calling a different end point, if we are not getting threshold values from this one.

AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift
```diff
@@ -141,6 +142,43 @@ class CompleteScreenViewModel: ObservableObject {
...
+                                                                                                                 low: Int32(30),
+                                                                                                                 medium: Int32(40),
+                                                                                                                 high: Int32(50),
+                                                                                                                 veryHigh: Int32(120)))})
+                        self.session.stream = []
+                        partialSession.forEach { self.session.stream.append($0) }
+                        Log.info("Completed downloading missing streams.")
+                        self.getMeasurementsAndDisplayData()
+                    case .failure(let failure):
+                        Log.info("Something want wrong when downoading missing streams.")
```
anna1901, line 171: Typos in "want" and "downoading"

AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift
```diff
@@ -141,6 +142,43 @@ class CompleteScreenViewModel: ObservableObject {
...
+                    }
+                }
+            } catch {
+                Log.info("Something want wrong when downoading missing streams.")
```
anna1901, line 175: Could we have a different message here, so that we would know what went wrong when reading logs?

AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift
```diff
@@ -141,6 +142,43 @@ class CompleteScreenViewModel: ObservableObject {
         exitRoute()
     }
     
+    private func getAndDisplayData() {
+        if session.stream.first!.sensorName.contains("AirBeam") {
+            do {
+                try streamsDownloader.downloadStreams(with: session.id, for: .airBeam3) { result in
+                    switch result {
+                    case .success(let success):
```
anna1901, line 150: Could you consider changing a name of the variable to `downloadedStreams` or something like that instead of `success`? `success` doesn't really say much about what this variable contains

AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift
```diff
@@ -141,6 +142,43 @@ class CompleteScreenViewModel: ObservableObject {
...
+                        // TODO: - FIX Thresholds and Improve this part of code
+                        // TODO: Compare strings and get threshold value, example - comment below
+                        // let vl = MeasurementStream.init(sensorName: .pm1, sensorPackageName: "").thresholdVeryLow
+                        let partialSession = success.map({ PartialExternalSession.Stream(id: $0.streamId,
```
anna1901, line 154: Is this variable a partial session or session streams? 🤔

AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift
```diff
@@ -141,6 +142,43 @@ class CompleteScreenViewModel: ObservableObject {
...
     
+    private func getAndDisplayData() {
+        if session.stream.first!.sensorName.contains("AirBeam") {
+            do {
+                try streamsDownloader.downloadStreams(with: session.id, for: .airBeam3) { result in
+                    switch result {
+                    case .success(let success):
+                        // TODO: - FIX Thresholds and Improve this part of code
+                        // TODO: Compare strings and get threshold value, example - comment below
+                        // let vl = MeasurementStream.init(sensorName: .pm1, sensorPackageName: "").thresholdVeryLow
```
anna1901, line 153: I don't fully understand this comment but we shouldn't read threshold values from database at this stage. We should use the threshold values that we get from backend

pawelgil (changes requested):

AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift
```diff
@@ -170,15 +208,30 @@ class CompleteScreenViewModel: ObservableObject {
...
-                        (self.chartStartTime, self.chartEndTime) = self.chartViewModel.generateEntries(with: stream.measurements.map({ SearchAndFollowChartViewModel.ChartMeasurement(value: $0.value, time: $0.time) }), thresholds: stream.thresholdsValues)
+                        self.assignValues(with: stream)
+                        self.defineChartRange(with: stream)
                     }
                 }
             }
         }
     }
     
+    func streamChanged() {
```
anna1901, line 219: Is this function called anywhere?

AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreen.swift
anna1901: You could add a code
```
if let stream = externalSessionWithStreams?.streams.first(where:  { $0.id == id }) {
            defineChartRange(with: stream)
        }
 ``` 

in `selectedStream(with id: Int)` function for example

Comment (anna1901) [reply AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreen.swift:42]: And add `entries = []` in generateEntries method

**Review by anna1901 — CHANGES_REQUESTED** [review]

File: AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift (hunk 5 above)
Comment (jankrzempek) [reply AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift:175]: I've added error.localizedDescription so far

File: AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift (hunk 9 above)
Comment (jankrzempek) [reply AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift:219]: on the view, yes

File: AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreen.swift (hunk 2 above)
Comment (jankrzempek) [reply AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreen.swift:42]: Wow, yes. thaaaanks!

File: AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift [hunk 10]
```diff
@@ -141,6 +145,51 @@ class CompleteScreenViewModel: ObservableObject {
         exitRoute()
     }
     
+    private func refresh() {
+        guard ((session.stream.first?.sensorName.contains("AirBeam")) != nil) else { getMeasurementsAndDisplayData(); return }
+        var currentSensor = AirBeamStreamPrefix.airBeam3
+        if session.stream.first!.sensorName.contains("AirBeam2") { currentSensor = .airBeam2 }
```
anna1901, line 151: Is it possible, that there would be no stream for this session? Maybe we could add a guard before to check if session.stream.first is not nil?

AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift
```diff
@@ -141,6 +145,51 @@ class CompleteScreenViewModel: ObservableObject {
...
+        do {
+            try streamsDownloader.downloadStreams(with: session.id, for: currentSensor) { result in
+                switch result {
+                case .success(let downloadedStreams):
+                    
+                    // TODO: - FIX Thresholds, get those values from backend not from our hardcoded struct
+                    #warning("🚨 FIX Thresholds 🚨")
```
anna1901, line 158: I like how explicit this is 😄

AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift
```diff
@@ -141,6 +145,51 @@ class CompleteScreenViewModel: ObservableObject {
...
+                    let streamHardcodedData = [MeasurementStream(sensorName: .f, sensorPackageName: ""),
+                                               MeasurementStream(sensorName: .pm1, sensorPackageName: ""),
+                                               MeasurementStream(sensorName: .pm10, sensorPackageName: ""),
+                                               MeasurementStream(sensorName: .pm2_5, sensorPackageName: ""),
+                                               MeasurementStream(sensorName: .rh, sensorPackageName: "")]
+                    
+                    let sessionStream = downloadedStreams.map({ stream -> PartialExternalSession.Stream in
```
anna1901, line 165: Thanks, this name is much better. I would change it to plural though, as it is holding multiple streams

AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift
```diff
@@ -141,6 +145,51 @@ class CompleteScreenViewModel: ObservableObject {
...
+                    self.session.stream = []
+                    sessionStream.forEach { self.session.stream.append($0) }
+                    Log.info("Completed downloading missing streams.")
+                    self.getMeasurementsAndDisplayData()
+                case .failure(let error):
+                    Log.error("Something went wrong when downoading missing streams. \(error.localizedDescription)")
+                }
+            }
+        } catch {
+            Log.error("Something went wrong when downoading missing streams. \(error.localizedDescription)")
```
anna1901, line 188: Typo in `downoading`

AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift
anna1901: Now I think it's not anymore

anna1901 (commented):
I can't test for now because of some technical issues. I will test in a moment and if all goes well I'll give an approve :)

AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift
```diff
@@ -94,6 +95,9 @@ class CompleteScreenViewModel: ObservableObject {
     }
     
     func selectedStream(with id: Int) {
+        if let stream = externalSessionWithStreams?.streams.first(where: { $0.id == id }) {
```
anna1901, line 98: The symbol below the chart doesn't change. I think this method should look something like this:
```
if let stream = externalSessionWithStreams?.streams.first(where: { $0.id == id }) {
            assignValues(with: stream)
            defineChartRange(with: stream)
        }
```

anna1901 (approved):
Please just fix this symbol below the chart. Otherwise, works great!! Good work 🚀

PR #715, opened 2022-05-23
Proper predicate passed to corresponding fetch.

jankrzempek:


anna1901 (approved):
Ah, yes, my bad. I spotted that and fixed in the sessionable protocol branch too, but we can merge it separately :)

pawelgil (approved):

PR #717, opened 2022-05-26
Reordering now includes external sessions too

pawelgil:


AirCasting/CoreData/UIStateHolderEntity+CoreDataClass.swift
```diff
@@ -0,0 +1,23 @@
...
+//
+//  Created by Pawel Gil on 26/05/2022.
+//
+//
+
+import Foundation
+import CoreData
+
+@objc(UIStateHolderEntity)
+public class UIStateHolderEntity: NSManagedObject {
```
anna1901, line 13: This can be removed?

AirCasting/CoreData/UIStateHolderEntity+CoreDataProperties.swift
```diff
@@ -0,0 +1,23 @@
...
+//
+//  Created by Pawel Gil on 26/05/2022.
+//
+//
+
+import Foundation
+import CoreData
+
+
+extension UIStateHolderEntity {
```
anna1901, line 13: Same as this one?

AirCasting/CoreData/Storages/UIStorage.swift
```diff
@@ -66,6 +66,49 @@ final class HiddenCoreDataUIStorage: UIStorageContextUpdate {
...
+    }
+
+    func giveHighestOrder(to sessionUUID: SessionUUID) {
+        do {
+            guard let session = try context.existingSessionable(uuid: sessionUUID) else {
+                Log.error("Trying to update order for nonexistent session [\(sessionUUID)]")
+                return
+            }
+            let highestOrder = try context.getHighestRowOrder()
+            session.userInterface?.rowOrder = (highestOrder ?? 0) + 1
```
anna1901, line 92: We need to add `createUIStateIfNeededForSessionable(session)` above, because userInterface gets deleted when a session is unfollowed (probably something we want to change) and so adding this highest order doesn't work for sessions that had previously been followed

anna1901 (approved):
Some small changes need to be done, but otherwise works well 🚀✅

PR #718, opened 2022-05-27
Add map to external session card

anna1901:
https://trello.com/c/8GRFhdYe


pawelgil (approved):
Didn't find any mistakes. I feel that we should at some point take a look at making those Cards more composable, so we can reuse the same code for all, yet being able to specify different components for different sesisons/tabs.

PR #723, opened 2022-05-30
Adding loading indicator to the create account view

SzymonGorowski:
https://trello.com/c/J31pMYJM/665-adding-loading-indicator-to-the-create-account-view

### DIFF
--- a/AirCasting/AuthViews/CreateAccountView.swift
+++ b/AirCasting/AuthViews/CreateAccountView.swift
@@ -24,15 +24,14 @@ struct CreateAccountView: View {
     @State private var isUsernameBlank = false
     @State private var presentedError: AuthorizationError?
     @State private var alert: AlertInfo?
-    @State private var isActive: Bool
+    @State private var isLoading = false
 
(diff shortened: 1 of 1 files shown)


AirCasting/AuthViews/CreateAccountView.swift
```diff
@@ -24,12 +24,22 @@ struct CreateAccountView: View {
     @State private var isUsernameBlank = false
     @State private var presentedError: AuthorizationError?
     @State private var alert: AlertInfo?
-    
-    init(completion: @escaping () -> Void) {
+    @State private var isActive: Bool
```
anna1901, line 27: I'm not sure it's the right variable name. What exactly is active? I this a better name would be: `isLoading`, because we are setting it to true when data is loading, right? Also, I think it could be set to false here instead of a value being passed to the initializer like with the rest of @State variables.

anna1901 (approved):
Wow, so fast. Good job. 😄 🚀 I added just a few suggestions for improvements.

PR #726, opened 2022-05-31
Handle card expand on tap

jankrzempek:
https://trello.com/c/ZvbFcgXl/667-following-tap-parameter-card-should-expand-on-parameter-tap


SzymonGorowski (approved):
Works as expected! 🚀

PR #728, opened 2022-06-01
Graph - Search&Follow

jankrzempek:
Apologies for formatting code in this PR, I won't do it anymore from now.

The thing:
This code is not perfect and I know that. I have copied the view files and adjusted them to the external session. I would like to move forward with it and not change the whole concept, at least for now. In other PR I would want to rethink the concept and adjust the code so that we won't be in need to copy and paste the whole view.


SzymonGorowski (commented):
Works great, good job! 🥇 🚀

AirCasting/Services/RemoveOldMeasurementsService.swift
```diff
@@ -21,20 +20,41 @@ final class RemoveOldMeasurementsService {
...
-            try removeMeasurementsSurplus(context: context, streams: session.allStreams)
+            try removeMeasurementsSurplus(context: context, streams: session.allStreams, provider: session.provider)
         }
     }
     
-    private func removeMeasurementsSurplus(context: NSManagedObjectContext, streams: [MeasurementStreamEntity]) throws {
+    // PurpleAir — measurements from there are not consistent, so we need to do more 'manual work' 🔧
+    // OpenAQ — measurements from there are not consistent, so we need to do more 'manual work' 🔧
+    // 'manual work' - taking the newest measurement and by having its time we are going back exactly 24 hours
+    // AirBeam - measurement every 1 min (60 times/1h): 1440 measurements (per 24 hours)
```
pawelgil, line 30: what happened here is a call for refactor of this class:
1. Create a protocol for this class:
2. Create multiple implementations - one for each provider
3. Register them in `Resolver` and inject based on the argument which would be a `SessionUUID`
4. Inject wherever it's needed

pawelgil (commented):
CR is not finished, publishing comments because my internet got downed by the thunderstorm

AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift
```diff
@@ -227,33 +238,33 @@ class CompleteScreenViewModel: ObservableObject {
...
     private func defineChartRange(with stream: ExternalSessionWithStreamsAndMeasurements.Stream) {
-    guard let separatedSensorName = self.componentsSeparation(name: stream.sensorName) else {
-        Log.error("No sensor name can be extracted from current stream.sensorName")
-        return
-    }
-            (self.chartStartTime, self.chartEndTime) = self.chartViewModel.generateEntries(with: stream.measurements.map({ SearchAndFollowChartViewModel.ChartMeasurement(value: $0.value, time: $0.time) }), thresholds: stream.thresholdsValues, using: ChartMeasurementsFilterDefault(name: separatedSensorName))
+        guard let separatedSensorName = self.componentsSeparation(name: stream.sensorName) else {
+            Log.error("No sensor name can be extracted from current stream.sensorName")
+            return
+        }
```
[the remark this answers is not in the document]
jankrzempek: I know now, I mentioned in the PR description - sorry for that

AirCasting/Services/RemoveOldMeasurementsService.swift
jankrzempek: I am not sure about 4th one. I will try to do it and if you could please check then

AirCasting.xcodeproj/project.pbxproj
```diff
@@ -457,7 +457,12 @@
 		DDF62601269C779800F23489 /* UIApplication+AppInfo.swift.swift in Sources */ = {isa = PBXBuildFile; fileRef = DDF62600269C779800F23489 /* UIApplication+AppInfo.swift.swift */; };
 		DDF7599A268332E900D906BA /* Strings.swift in Sources */ = {isa = PBXBuildFile; fileRef = DDF75999268332E900D906BA /* Strings.swift */; };
 		EE969E7E1981C41852FB7745 /* Pods_AirCastingTests.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 3A54057326D27D23E9906704 /* Pods_AirCastingTests.framework */; };
+		FA028B6C28520C0700BB66A0 /* RemoveOldMeasurements.swift in Sources */ = {isa = PBXBuildFile; fileRef = FA028B6B28520C0700BB66A0 /* RemoveOldMeasurements.swift */; };
+		FA028B6E28520C1000BB66A0 /* OpenAQMeasurementsService.swift in Sources */ = {isa = PBXBuildFile; fileRef = FA028B6D28520C1000BB66A0 /* OpenAQMeasurementsService.swift */; };
```
anna1901, line 461: OpenAQRemoveOldMeasurementsService?

AirCasting.xcodeproj/project.pbxproj
```diff
@@ -457,7 +457,12 @@
...
+		FA028B7028520C1800BB66A0 /* PurpleAirMeasurementsService.swift in Sources */ = {isa = PBXBuildFile; fileRef = FA028B6F28520C1800BB66A0 /* PurpleAirMeasurementsService.swift */; };
```
anna1901, line 462: Same as above

AirCasting/APICommunicator/StreamDownloader.swift
```diff
@@ -7,7 +7,7 @@ import Resolver
 struct StreamWithMeasurements: Decodable {
     let title: String
     let username: String
-    let measurements: [StreamWithMeasurements.Measurements]
+    var measurements: [StreamWithMeasurements.Measurements]
```
anna1901, line 10: Why is this changed?

AirCasting/Graph/External/ExternalGraph.swift
```diff
@@ -0,0 +1,170 @@
...
+        context.coordinator.entries = entries
+        context.coordinator.currentThreshold = ThresholdWitness(sensorThreshold: thresholds)
+        context.coordinator.totalNumberOfMeasurements = entries.count
+        context.coordinator.currentMeasurementsNumber = entries.count
+        context.coordinator.stream = stream
+    }
+    
+    private func simplifyGraphline(entries: [ChartDataEntry], uiView: AirCastingGraph) {
+        
+        let counter: Int = calculateSeeingPointsNumber(entries: entries, uiView: uiView)
```
anna1901, line 89: SeenPoints? VisiblePoints? I don't know what "SeeingPoint" could be :D

AirCasting/Graph/External/ExternalGraph.swift
```diff
@@ -0,0 +1,170 @@
...
+            let previous = Calendar.current.nextDate(after: midnight,
+                                                    matching: components,
+                                                    matchingPolicy: Calendar.MatchingPolicy.previousTimePreservingSmallerComponents,
+                                                    repeatedTimePolicy: Calendar.RepeatedTimePolicy.first,
+                                                     direction: .backward)
+            
+            if let midnightPoint = previous?.currentUTCTimeZoneDate.timeIntervalSince1970 {
+                midnightPoints.append(midnightPoint)
+            }
+            midnight = previous!
```
anna1901, line 131: Maybe we should add some guard let for this previous unwrapping?

AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift
```diff
@@ -52,6 +52,12 @@ class CompleteScreenViewModel: ObservableObject {
     let sessionEndTime: Date
     let sensorType: String
     let exitRoute: () -> Void
+    
+    private var sessionAlreadyFollowed: Bool {
```
anna1901, line 56: We don't need to add this back I think? Please make sure you are not adding some old things from develop (which have been modified on purpose in previous PRs) by accident.

AirCasting/Services/OldMeasurementRemovers/PurpleAirMeasurementsService.swift
```diff
@@ -0,0 +1,28 @@
...
+    
+    func removeOldestMeasurements(in context: NSManagedObjectContext, from sessionUUID: SessionUUID) throws {
+        let session = try context.existingExternalSession(uuid: sessionUUID)
+        try removeMeasurementsSurplus(context: context, streams: session.allStreams)
+    }
+    
+    private func removeMeasurementsSurplus(context: NSManagedObjectContext, streams: [MeasurementStreamEntity]) throws {
+        for stream in streams {
+            guard let measurementsCount = stream.measurements?.count else { continue }
+            timeBasedRemover(context: context, measurementsCount: measurementsCount, stream: stream)
```
anna1901, line 16: Do we need to pass this measurementsCount variable at all?

anna1901 (commented):
Did you consider changing the algorithm for removing old measurements of AirBeam sessions and just always apply the timeBasedRemover? Then we wouldn't even have to check the provider.

AirCasting/APICommunicator/StreamDownloader.swift
jankrzempek: Zrzut ekranu 2022-06-13 o 09 30 03
Other way I am not able to assign those filter measurements here (CompleteScreenViewModel)

AirCasting/Graph/External/ExternalGraph.swift
jankrzempek: Yeah, Visible sounds way better thx!

AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift
jankrzempek: Sorry for that

AirCasting/Graph/External/ExternalGraph.swift
```diff
@@ -0,0 +1,168 @@
...
+    
+    @InjectedObject private var userSettings: UserSettings
+    @ObservedObject var stream: MeasurementStreamEntity
+    @ObservedObject var thresholds: SensorThreshold
+    private var rangeChangeAction: OnChange?
+    
+    var isAutozoomEnabled: Bool
+    let simplifiedGraphEntryThreshold = 1000
+    
+    init(stream: MeasurementStreamEntity, thresholds: SensorThreshold, isAutozoomEnabled: Bool) {
```
anna1901, line 17: I'm asking out of curiosity: since this view doesn't get sessionEntity as an argument, why did you create a new view for external sessions? MeasurementsStreams are the same for both types of sessions.

AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift
```diff
@@ -231,8 +231,16 @@ class CompleteScreenViewModel: ObservableObject {
                     return
                 }
                 
+                // Special filter for Sensors as those measurements are not consistent
+                var streamWithMeasureementsCopy = downloadedStreamsWithMeasurements
+                guard let measurements = streamWithMeasureementsCopy.first?.measurements,
+                      let lastMeasurement = measurements.last else { return }
+                let twentyFourHours = 86400000 // 24 hours in miliseconds: 60 * 60 * 24
+                let beginingOfCorrectPeriod = lastMeasurement.time - twentyFourHours
+                streamWithMeasureementsCopy[0].measurements = (measurements.filter({ $0.time >= beginingOfCorrectPeriod }))
```
anna1901, line 240: Why are you doing this only for the first stream?

AirCasting/Services/OldMeasurementRemovers/DefaultRemoveOldMeasurementsService.swift
```diff
@@ -0,0 +1,36 @@
...
+    }
+    
+    private func removeMeasurementsSurplus(context: NSManagedObjectContext, streams: [MeasurementStreamEntity]) throws {
+        for stream in streams {
+            timeBasedRemover(context: context, stream: stream)
+        }
+    }
+    
+    private func timeBasedRemover(context: NSManagedObjectContext, stream: MeasurementStreamEntity) {
+        let lastMeasurement = stream.allMeasurements?.last
```
anna1901, line 29: How about adding guard let here so that we won't have to force unwrap it later?

AirCasting/Services/OldMeasurementRemovers/DefaultRemoveOldMeasurementsService.swift
```diff
@@ -0,0 +1,36 @@
...
+        let twentyFour = 86400000 // 24 hours in miliseconds: 60 * 60 * 24
+        let beginingOfCorrectPeriod = lastMeasurement!.time.milliseconds - twentyFour
+        stream.allMeasurements?.reversed().forEach({ measurement in
```
anna1901, line 32: Maybe instead of iterating through the whole array of measurements, we could use a fetch request to fetch measurements with date earlier that 24h ago and then deleting each of them? Have you thought about that kind of solution?

AirCasting/Graph/External/ExternalGraph.swift
jankrzempek: Sure, but I think that having the same instance of Graph for two things (external and internal sessions) can give us less possibility for flexible modification in the future. I mean by that; the external session is mostly the same in our app, but at the same time way different then internal one -  so splitting it into different ones can help us in the future I believe

AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift
jankrzempek: My bad, I shouldn't

AirCasting/Services/OldMeasurementRemovers/DefaultRemoveOldMeasurementsService.swift
jankrzempek: of course, Thanks!!

AirCasting/Services/OldMeasurementRemovers/DefaultRemoveOldMeasurementsService.swift
jankrzempek: I don't quite get it, could you please elaborate?

AirCasting/Graph/External/ExternalGraph.swift
```diff
@@ -0,0 +1,168 @@
...
+        context.coordinator.currentThreshold = ThresholdWitness(sensorThreshold: thresholds)
+        context.coordinator.currentMeasurementsNumber = calculateVisiblePointsNumber(entries: entries, uiView: uiView)
+        context.coordinator.entries = entries
+        context.coordinator.lastMeasurementTime = stream.allMeasurements?.last?.time
+        context.coordinator.stream = stream
+        return uiView
+    }
+    
+    func updateUIView(_ uiView: AirCastingGraph, context: Context) {
+
```
SzymonGorowski, line 53: I think you can remove the blank space here.

AirCasting/Graph/External/ExternalGraph.swift
```diff
@@ -0,0 +1,168 @@
...
+        return newGraph
+    }
+    
+    func makeUIView(context: Context) -> AirCastingGraph {
+        let uiView = AirCastingGraph(onDateRangeChange: { newRange in
+            rangeChangeAction?(newRange)
+        })
+        try? uiView.updateWithThreshold(thresholdValues: thresholds.rawThresholdsBinding.wrappedValue)
+
+        let entries = stream.allMeasurements?.sorted(by: { $0.time < $1.time }).compactMap({ measurement -> ChartDataEntry? in
```
SzymonGorowski, line 35: You can also use `.sorted { $0.time < $1.time }` a little bit shorter version. Also in the line 71.

AirCasting/Graph/External/ExternalGraph.swift
```diff
@@ -0,0 +1,168 @@
...
+        
+        context.coordinator.entries = entries
+        context.coordinator.currentThreshold = ThresholdWitness(sensorThreshold: thresholds)
+        context.coordinator.currentMeasurementsNumber = entries.count
+        context.coordinator.lastMeasurementTime = stream.allMeasurements?.last?.time
+        context.coordinator.stream = stream
+    }
+    
+    private func simplifyGraphline(entries: [ChartDataEntry], uiView: AirCastingGraph) {
+
```
SzymonGorowski, line 87: Same as above.

SzymonGorowski (approved):
Works great! Well done 🚀 🥇

AirCasting/SessionViews/ExternalSessionsViews/ExternalSessionCard.swift
```diff
@@ -26,43 +26,51 @@ struct ExternalSessionCard: View {
...
                 mapStatsDataSource?.dataSource.stream = newStream
                 uiStateHandler.changeSelectedStream(sessionUUID: session.uuid, newStream: newStream?.sensorName ?? "")
+                graphStatsDataSource.stream = newStream
+                graphStatsDataSource.dataSource.stream = newStream
             })
             .onChange(of: isMapButtonActive) { _ in
                 reorderButton.setHidden(if: isMapButtonActive)
                 searchAndFollowButton.setHidden(if: isMapButtonActive)
             }
     }
```
SzymonGorowski, line 72: You can also add here the below code to resolve bugs with appearing icons when the user will swipe from the graph.

`.onChange(of: isGraphButtonActive) { _ in
            reorderButton.setHidden(if: isGraphButtonActive)
            searchAndFollowButton.setHidden(if: isGraphButtonActive)
        }`

AirCasting/Services/OldMeasurementRemovers/DefaultRemoveOldMeasurementsService.swift
```diff
@@ -0,0 +1,40 @@
...
+        for stream in streams {
+            timeBasedRemover(context: context, stream: stream)
+        }
+    }
+    
+    private func timeBasedRemover(context: NSManagedObjectContext, stream: MeasurementStreamEntity) {
+        guard let lastMeasurement = stream.allMeasurements?.last else { Log.error("No last measurement when trying to remove those from > 24h"); return }
+        let twentyFour = 86400000 // 24 hours in miliseconds: 60 * 60 * 24
+        let beginingOfCorrectPeriod = lastMeasurement.time.milliseconds - twentyFour
+        stream.allMeasurements?.forEach({ measurement in
```
anna1901, line 36: I meant that instead of going through all of the 1400 measurements and checking the time we could let core data handle it with a fetch request like this:

```
private func timeBasedRemover(context: NSManagedObjectContext, stream: MeasurementStreamEntity) {
        guard let lastMeasurementTime = stream.allMeasurements?.last?.time else { Log.error("No last measurement when trying to remove those from > 24h"); return }
        let threshold = lastMeasurementTime.twentyFourHoursBefore
        
        let req: NSFetchRequest<MeasurementEntity> = NSFetchRequest(entityName: "MeasurementEntity")
        req.predicate = NSPredicate(format: "time < %@ AND measurementStream == %@", threshold as CVarArg, stream)
        if let measurements = try? context.fetch(req) {
            measurements.forEach { Log.info("Removing measurement: \($0)"); context.delete($0) }
        }
    }
```

(I added `twentyFourHoursBefore` to a Date extension)
I'm not an expert on performance in iOS but I think this could be faster.

AirCasting/Graph/External/ExternalGraph.swift
jankrzempek: I cannot agree here with you. If we could have some views done in a more generic way then we could always apply some initial version of it and after that extend it according to the needs (of let's say) current implementation. What I mean by that is: having an ex. `CoreGraphView` and then inheriting from it in ex `ExternalGraphView` and `InternalGraphView` would allow us to explicitly state what is common and change it if the future would not affect anything else (possibly). In our case, the internal graph has notes when the second does not (maybe also it has some other changes which I do not remember). Having one graph here means always the same in both places.

At least for me, it is the same with having different structs for APIs or data which we are saving. We tend to make a few structs for the same data (even if those structs are sometimes exactly the same) based on the "file" so that in the future we could protect ourselves from changing most of the code; the change only would be done in the file where we really need it (most of the time, because passing data and so on comes here, but most of the time...)

AirCasting/Graph/External/ExternalGraph.swift
jankrzempek: I would stay with `by` prefix as it sound better to me as long as it cool with you?

jankrzempek: I would stay with `by` prefix as it sound better to me as long as it cool with you?

AirCasting/SessionViews/ExternalSessionsViews/ExternalSessionCard.swift
jankrzempek: Thank you 💙

AirCasting/Graph/External/ExternalGraph.swift
anna1901: Thanks for a detailed response. Sure, I forgot about the notes. :) In that case, I think it is good to keep them separately. But as you mention, it would be cool to have a parent class with all of the graph stylings.

AirCasting/Graph/External/ExternalGraph.swift
SzymonGorowski: Of course, it was just a suggestion.

AirCasting/Services/OldMeasurementRemovers/DefaultRemoveOldMeasurementsService.swift
```diff
@@ -0,0 +1,54 @@
...
+            try! context.save()
+            Log.info(success)
+        case .failure(let error):
+            Log.error(error.localizedDescription)
+        }
+    }
+}
+
+extension Date {
+    var twentyFourHoursBefore: Double {
```
anna1901, line 50: I think we can actually return it as a date, and then we wouldn't have to change it back to Date in deleteMeasurements function. It would look like this:
```
var twentyFourHoursBefore: Date {
        self - (24 * 60 * 60)
    }
```

anna1901: Also, I think it can be added in the Date+Extensions file

AirCasting/CoreData/NSManagedObjectContext+Utils.swift
```diff
@@ -13,6 +13,21 @@ extension NSManagedObjectContext {
...
     
+    enum MeasurementsDeletion: Error {
+        case errorWhenFetchingRequest(_: String)
+    }
+    
+    func deleteMeasurements(thresholdInMiliseconds: Double, stream: MeasurementStreamEntity) -> Result<String, Error> {
+        let req: NSFetchRequest<MeasurementEntity> = NSFetchRequest(entityName: "MeasurementEntity")
+        req.predicate = NSPredicate(format: "time < %@ AND measurementStream == %@",
+                                    NSDate(timeIntervalSince1970: thresholdInMiliseconds/1000), stream)
+        guard let measurements = try? self.fetch(req) else {
```
anna1901, line 24: What do you think about not returning Result here, but instead making it a throwing function?

AirCasting/Services/OldMeasurementRemovers/DefaultRemoveOldMeasurementsService.swift
```diff
@@ -0,0 +1,54 @@
...
+        }
+    }
+    
+    private func timeBasedRemover(context: NSManagedObjectContext, stream: MeasurementStreamEntity) {
+        guard let lastMeasurementTime = stream.allMeasurements?.last?.time else { Log.error("No last measurement when trying to remove those from > 24h"); return }
+        let threshold = lastMeasurementTime.twentyFourHoursBefore
+        let result = context.deleteMeasurements(thresholdInMiliseconds: threshold, stream: stream)
+        switch result {
+        case .success(let success):
+            try! context.save()
```
anna1901, line 41: What do you think about moving context.save() into the `deleteMeasurements` function? Then we could actually move both fetching and saving into separate queue (with context.perform() ). At least I think so. And if this method was a throwing function instead of returning a result, I think it would shorten the code:

```
do {
  context.deleteMeasurements ...
} catch {
  Log.error(error)
}

**Review by anna1901 — COMMENTED** [review]
Thank you for incorporating my suggestions. 😊 I added some additional comments and I'm going to test later (after you decide to add the changes or not).

File: AirCasting/Services/OldMeasurementRemovers/DefaultRemoveOldMeasurementsService.swift (hunk 19 above)
Comment (jankrzempek) [reply AirCasting/Services/OldMeasurementRemovers/DefaultRemoveOldMeasurementsService.swift:50]: I moved this extension to the file but I prefer to stay with miliseconds approach so then we are able to easly and in clear way use NSDate: ` NSDate(timeIntervalSince1970: thresholdInMiliseconds/1000)`

Comment (anna1901) [reply AirCasting/Services/OldMeasurementRemovers/DefaultRemoveOldMeasurementsService.swift]: Then I'd suggest adding inMiliseconds to the name for consistency, but I'll leave it up to you 🙂

**Review by anna1901 — APPROVED** [review]
simulator_screenshot_EB1D0DFF-9B08-404A-B615-02989F5ACA1A
There are still some discrepancies regarding the times on the graph that might be worth investigating (end date is not equal to most recent measurement time I think). But overall great job!🚀  This was a massive ticket with a lot to consider. I'm leaving those comments for you consideration.

File: AirCasting/Extensions/Date+Extensions.swift [hunk 22]
```diff
@@ -61,4 +61,11 @@ extension Date {
         calendar.timeZone = TimeZone.utc
         return calendar.startOfDay(for: self).timeIntervalSince1970 + (23 * 60 * 60 + 3540 + 59)
     }
+    
+    var twentyFourHoursBefore: Double {
+        let twentyFourHours = 86400000 // 24 hours in miliseconds: 60 * 60 * 24
+        return Double(self.milliseconds - twentyFourHours)
```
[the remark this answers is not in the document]
jankrzempek: Deep one, like the idea - changing to seconds right now

AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift
```diff
@@ -202,7 +202,17 @@ class CompleteScreenViewModel: ObservableObject {
     
     private func createExternalSession(with sortedStreams: [MeasurementsDownloaderResultModel.Stream]) {
         DispatchQueue.main.async {
-            self.externalSessionWithStreams = self.service.createExternalSession(from: self.session, with: sortedStreams)
+            var streamsWithMeasurements = sortedStreams
```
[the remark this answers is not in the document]
jankrzempek: Great idea, applied!

PR #737, opened 2022-06-06
Set card state when following a session

anna1901:
https://trello.com/c/ewqWXysF

### DIFF
--- a/AirCasting/SearchAndFollow/CompleteScreen/Services/SearchAndFollowCompleteScreenService.swift
+++ b/AirCasting/SearchAndFollow/CompleteScreen/Services/SearchAndFollowCompleteScreenService.swift
@@ -64,6 +64,11 @@ struct DefaultSearchAndFollowCompleteScreenService: SearchAndFollowCompleteScree
             case .success:
                 uiStore.accessStorage { store in
                     store.giveHighestOrder(to: session.uuid)
+                    do {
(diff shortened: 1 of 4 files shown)


AirCasting/SearchAndFollow/CompleteScreen/Services/SearchAndFollowCompleteScreenService.swift
```diff
@@ -64,6 +64,11 @@ struct DefaultSearchAndFollowCompleteScreenService: SearchAndFollowCompleteScree
             case .success:
                 uiStore.accessStorage { store in
                     store.giveHighestOrder(to: session.uuid)
+                    do {
+                        try store.cardStateToggle(for: session.uuid)
+                    } catch {
+                        Log.error("Changing card state failed: \(error)")
```
pawelgil, line 70: don't we want `completion(.failure(error))` here?

pawelgil (approved):

AirCasting/SearchAndFollow/CompleteScreen/Services/SearchAndFollowCompleteScreenService.swift
anna1901: I think this uiState thing is not essential so I wouldn't block following a session only because we couldn't save the card state

PR #738, opened 2022-06-07
Add more space in the search and follow chart

anna1901:
https://trello.com/c/KQmMQHpj

It's a very basic version of the solution. We can work on it further later on.


jankrzempek (approved):
🥳 The one thing that I saw while testing is `-0` thing which I am not sure about. It's maybe not related at all but just letting you know
IMG_0024

PR #743, opened 2022-06-08
Fix adding indoor argument when creating cellular sessions

anna1901:
There was a small bug for creating indoor cellular sessions. We were not passing this argument when creating a session struct and it was set to false by default. This PR fixes that.


SzymonGorowski (approved):
🥇

jankrzempek (approved):
👏

pawelgil (approved):

PR #751, opened 2022-06-10
Changing shadow on the cards

SzymonGorowski:
https://trello.com/c/DWKJchFj/370-dashboard-cards-make-them-look-like-android

### DIFF
--- a/AirCasting/Dashboard/DashboardView.swift
+++ b/AirCasting/Dashboard/DashboardView.swift
@@ -155,6 +155,7 @@ struct DashboardView: View {
                         }
                     }
                 }
+                .padding(5)
             }
             .coordinateSpace(name: dashboardCoordinateSpaceName)
         }
--- a/AirCasting/Dashboard/DashboardView.swift
+++ b/AirCasting/Dashboard/DashboardView.swift
@@ -155,11 +155,10 @@ struct DashboardView: View {
                         }
                     }
                 }
-                .padding(5)
+                .padding(.horizontal)
             }
             .coordinateSpace(name: dashboardCoordinateSpaceName)
         }
-        .padding(.horizontal)
         .frame(maxWidth: .infinity)
         .background(Color.aircastingGray.opacity(0.05))
     }
--- a/AirCasting/SearchAndFollow/SearchMap/BottomCard.swift
+++ b/AirCasting/SearchAndFollow/SearchMap/BottomCard.swift
@@ -45,7 +45,7 @@ struct BottomCardView: View {
         .background(
             Group {
                 Color.white
-                    .shadow(color: .bottomCardShadow, radius: 9, x: 0, y: 1)
+                    .cardShadow()
             }
         )
         .cornerRadius(8)
--- a/AirCasting/SearchAndFollow/SearchMap/BottomCard.swift
+++ b/AirCasting/SearchAndFollow/SearchMap/BottomCard.swift
@@ -45,7 +45,7 @@ struct BottomCardView: View {
         .background(
             Group {
                 Color.white
-                    .shadow(color: .sessionCardShadow, radius: 9, x: 0, y: 1)
+                    .shadow(color: .bottomCardShadow, radius: 9, x: 0, y: 1)
             }
         )
         .cornerRadius(8)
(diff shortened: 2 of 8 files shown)


Modules/AirCastingStyling/Sources/AirCastingStyling/Color.swift
```diff
@@ -48,7 +48,11 @@ extension Color {
...
     }
     
-    public static var sessionCardShadow: Color {
+    public static var bottomCardShadow: Color {
         return Color(red: 205/255, green: 209/255, blue: 214/255, opacity: 0.36)
     }
+    
+    public static var sessionCardShadow: Color {
+        return Color(red: 205/255, green: 209/255, blue: 214/255)
+    }
```
jankrzempek, line 57: Maybe we could have something like 
```
public static var sessionCardShadow: sessionCardShadow2 {
    return sessionCardShadow2.values
}
}

public enum sessionCardShadow2 {
case values

var color: Color {
    switch self {
    default: return .red
    }
}

var radius: Int {
    switch self {
    default: return 0
    }
}
}
```
and then in every place which needs to use that, we could have `.shadow(color: .color, radius: .radius, x: .x, y: .y))` I am not sure if it should be enum or struct or sth else. In the end it is just a suggestion, please skip it if you don't want to consider 😎

jankrzempek (approved):
Looking like Android (🫣), very well completed!

AirCasting/SearchAndFollow/SearchMap/BottomCard.swift
```diff
@@ -45,7 +45,7 @@ struct BottomCardView: View {
         .background(
             Group {
                 Color.white
-                    .shadow(color: .sessionCardShadow, radius: 9, x: 0, y: 1)
+                    .shadow(color: .bottomCardShadow, radius: 9, x: 0, y: 1)
```
jankrzempek, line 48: Here you u`re using the old-style version approach, please change to the new one 😄

jankrzempek (changes requested):

Modules/AirCastingStyling/Sources/AirCastingStyling/CardShadow.swift
```diff
@@ -0,0 +1,22 @@
+//
+//  CardShadow.swift
+//  
+//
+//  Created by lunar  on 13/06/2022.
+//
+
+import SwiftUI
+
+public struct CardShadow: ViewModifier {
```
anna1901, line 10: 👏🏻 Nice! Thanks for separating it like that!

Modules/AirCastingStyling/Sources/AirCastingStyling/CardShadow.swift
```diff
@@ -0,0 +1,22 @@
...
+    
+    public func body(content: Content) -> some View {
+        content
+            .shadow(color: .sessionCardShadow, radius: 3, x: 0, y: 2)
```
anna1901, line 14: I'm not sure why you decided to set x value to 0. If it was to make shadow symethrical, then you can actually add two `.shadow` modifiers, and this way add it on both sides. Like this:

```
.shadow(color: .sessionCardShadow, radius: 3, x: 1, y: 1)
.shadow(color: .sessionCardShadow, radius: 3, x: -1, y: -1)
```

AirCasting/Dashboard/DashboardView.swift
```diff
@@ -155,6 +155,7 @@ struct DashboardView: View {
                         }
                     }
                 }
+                .padding(5)
```
anna1901, line 158: I think cards got too narrow this way. I realize that you probably did it so that the shadow on the sides would be visible, but you can achieve it also by deleting `.padding(.horizontal)` from ZStack and adding it here instead.

AirCasting/SearchAndFollow/SearchMap/BottomCard.swift
jankrzempek: Yeah, looks good I think. Please change this if possible

jankrzempek (approved):

anna1901 (approved):
Awesome 🚀

PR #753, opened 2022-06-13
Adding alert at signout asking the user about  deleting locationless sessions

SzymonGorowski:
https://trello.com/c/wVeUS8Fg/526-on-logout-let-the-user-know-that-they-will-loose-the-locationless-sessions

### DIFF
--- a/AirCasting/CoreData/NSManagedObjectContext+Utils.swift
+++ b/AirCasting/CoreData/NSManagedObjectContext+Utils.swift
@@ -104,12 +104,12 @@ extension NSManagedObjectContext {
         return new
     }
     
-    func checkForLocationlessSessions() throws -> Bool  {
+    func anyLocationlessSessionsPresent() throws -> Bool  {
         let fetchRequest: NSFetchRequest<SessionEntity> = SessionEntity.fetchRequest()
         fetchRequest.predicate = NSPredicate(format: "locationless = %d", true)
 
         let results = try self.fetch(fetchRequest)
-        if !results.isEmpty { return true } else { return false }
+        return !results.isEmpty
     }
 }
 
--- a/AirCasting/CoreData/NSManagedObjectContext+Utils.swift
+++ b/AirCasting/CoreData/NSManagedObjectContext+Utils.swift
@@ -103,6 +103,14 @@ extension NSManagedObjectContext {
         new.setValue(uuid.rawValue, forKey: "uuid")
         return new
     }
+    
+    func checkForLocationlessSessions() throws -> Bool  {
+        let fetchRequest: NSFetchRequest<SessionEntity> = SessionEntity.fetchRequest()
+        fetchRequest.predicate = NSPredicate(format: "locationless = %d", true)
+
+        let results = try self.fetch(fetchRequest)
+        if !results.isEmpty { return true } else { return false }
+    }
 }
 
 extension NSManagedObjectContext {
--- a/AirCasting/CoreData/NSManagedObjectContext+Utils.swift
+++ b/AirCasting/CoreData/NSManagedObjectContext+Utils.swift
@@ -103,14 +103,6 @@ extension NSManagedObjectContext {
         new.setValue(uuid.rawValue, forKey: "uuid")
         return new
     }
-    
-    func anyLocationlessSessionsPresent() throws -> Bool  {
-        let fetchRequest: NSFetchRequest<SessionEntity> = SessionEntity.fetchRequest()
-        fetchRequest.predicate = NSPredicate(format: "locationless = %d", true)
-
-        let results = try self.fetch(fetchRequest)
-        return !results.isEmpty
-    }
 }
 
 extension NSManagedObjectContext {
--- a/AirCasting/CoreData/Storages/MeasurementStreamStorage.swift
+++ b/AirCasting/CoreData/Storages/MeasurementStreamStorage.swift
@@ -289,6 +289,15 @@ final class HiddenCoreDataMeasurementStreamStorage: MeasurementStreamStorageCont
         }
     }
     
+    func checkForLocationlessSessions() throws -> Bool {
+        do {
+            return try context.checkForLocationlessSessions()
+        } catch {
+            Log.error("Error when checking for locationless sessions: \(error.localizedDescription)")
+        }
+        return false
+    }
+    
     // MARK: - Notes
 
     enum NoteStorageError: Swift.Error, LocalizedError {
--- a/AirCasting/CoreData/Storages/MeasurementStreamStorage.swift
+++ b/AirCasting/CoreData/Storages/MeasurementStreamStorage.swift
@@ -289,15 +289,6 @@ final class HiddenCoreDataMeasurementStreamStorage: MeasurementStreamStorageCont
         }
     }
     
-    func anyLocationlessSessionsPresent() throws -> Bool {
-        do {
-            return try context.anyLocationlessSessionsPresent()
-        } catch {
-            Log.error("Error when checking for locationless sessions: \(error.localizedDescription)")
-            return false
-        }
-    }
-    
     // MARK: - Notes
 
     enum NoteStorageError: Swift.Error, LocalizedError {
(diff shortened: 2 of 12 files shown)


AirCasting/CoreData/Storages/MeasurementStreamStorage.swift
```diff
@@ -289,6 +289,15 @@ final class HiddenCoreDataMeasurementStreamStorage: MeasurementStreamStorageCont
...
     }
     
+    func checkForLocationlessSessions() throws -> Bool {
+        do {
+            return try context.checkForLocationlessSessions()
+        } catch {
+            Log.error("Error when checking for locationless sessions: \(error.localizedDescription)")
+        }
+        return false
+    }
```
jankrzempek, line 299: Here maybe it would be cleaner to use return false inside catch as it should be the only reason when we should use literal `false` I think

jankrzempek (approved):
Looks great for me, works even better 🚀

AirCasting/Settings/MyAccountViewSignOut.swift
```diff
@@ -41,18 +42,19 @@ private extension MyAccountViewSignOut {
...
         Button(action: {
-            guard networkChecker.connectionAvailable else {
-                alert = InAppAlerts.unableToLogOutAlert()
-                return
-            }
-            userState.currentState = .loggingOut
-            do {
-                try logoutController.logout {
-                    userState.currentState = .idle
+            measurementStreamStorage.accessStorage { storage in
```
anna1901, line 45: Wouldn't it make sense to check for the connection before checking for locationless sessions? Because if the connection is unavailable we can't move forward with logging out anyway.

AirCasting/CoreData/NSManagedObjectContext+Utils.swift
```diff
@@ -103,6 +103,14 @@ extension NSManagedObjectContext {
         new.setValue(uuid.rawValue, forKey: "uuid")
         return new
     }
+    
+    func checkForLocationlessSessions() throws -> Bool  {
```
pawelgil, line 107: I don't quite dig this adding domain specific functions to `NSManagedObjectContext` using extensions. I think this should be handled solely by our tailored persistence/storage classes.

2nd thing here is that I don't think name of this functions matches return type:
- Hey, check for locationless sessions please
- Ok I checked that for you and `yes`.
:)

Maybe `anyLocationlessSessionsPresent` or something?

AirCasting/CoreData/NSManagedObjectContext+Utils.swift
```diff
@@ -103,6 +103,14 @@ extension NSManagedObjectContext {
...
+        let fetchRequest: NSFetchRequest<SessionEntity> = SessionEntity.fetchRequest()
+        fetchRequest.predicate = NSPredicate(format: "locationless = %d", true)
+
+        let results = try self.fetch(fetchRequest)
+        if !results.isEmpty { return true } else { return false }
```
pawelgil, line 112: how about just `return !results.isEmpty`?

AirCasting/Settings/MyAccountViewSignOut.swift
```diff
@@ -12,6 +12,7 @@ struct MyAccountViewSignOut: View {
     @Injected private var networkChecker: NetworkChecker
     @Injected private var logoutController: LogoutController
     @Injected private var deleteController: DeleteAccountController
+    @Injected private var measurementStreamStorage: MeasurementStreamStorage
```
pawelgil, line 15: please move storage logic to `LogoutController` (it will perhaps require some thinking because of the alert requirement, but please try. Ping me if problematic)

AirCasting/Settings/MyAccountViewSignOut.swift
```diff
@@ -94,4 +101,17 @@ private extension MyAccountViewSignOut {
         .padding(.bottom, 20)
         .padding()
     }
+    
+    private func logoutUser() {
+        userState.currentState = .loggingOut
```
pawelgil, line 106: I think we can move setting userState from here to the `LogoutController` too. This kind of logic should not be present in Views

pawelgil (changes requested):

AirCasting/Utils/Strings.swift
```diff
@@ -917,6 +917,13 @@ struct Strings {
                                                                    comment: "")
         static let accountDeletionSuccessMessage: String = NSLocalizedString("Your account has been deleted.",
                                                                    comment: "")
+        static let logoutWarningTitle: String = NSLocalizedString("Are you sure to sign out?",
+                                                                   comment: "")
+        static let logoutWarningMessage: String = NSLocalizedString("You will lose the session recorder with disabled location.",
+                                                                   comment: "")
+        static let logoutWarningConfirmButton: String = NSLocalizedString("Confirm",
+                                                                          comment: "")
+        static let fetchingDataFailedMessage: String = NSLocalizedString("Something went wrong fetching the data. Please try again later.", comment: "")
```
anna1901, line 926: Can we remove "fetching the data" from this message?

anna1901 (approved):
Screenshot 2022-06-16 at 21 35 51
Please change the text of the alert, cause it has mistakes. The header should read: "Are you sure you want to sign out?" . Or if we want shorter version, maybe we can say just "Are you sure?" in the header. Then in the text: "You will lose sessions recorded with disabled location". Unless it says "the session" when it's just one session and "sessions" when there are more, could be cool as well. 😄

AirCasting/Settings/SettingsMyAccountView.swift
```diff
@@ -0,0 +1,63 @@
+// Created by Lunar on 18/06/2021.
+//
+
+import SwiftUI
+import AirCastingStyling
+import Resolver
+
+struct SettingsMyAccountView<VM : SettingsMyAccountViewModel>: View {
```
jankrzempek, line 8: Please delete space, should be `<VM: Setting...`

AirCasting/Settings/SettingsMyAccountView.swift
```diff
@@ -0,0 +1,63 @@
...
+            }
+        }
+        .navigationTitle(Strings.Commons.myAccount)
+        .alert(item: $viewModel.alert, content: { $0.makeAlert() })
+    }
+}
+
+private extension SettingsMyAccountView {
+    var logInLabel: some View {
+        Text(Strings.SignOutSettings.logged + "\(KeychainStorage(service:  Bundle.main.bundleIdentifier!).getUsername())")
```
jankrzempek, line 31: I think you have two spaces `(service:  Bun`

AirCasting/Settings/SettingsMyAccountView.swift
```diff
@@ -0,0 +1,63 @@
...
+            Group {
+                HStack {
+                    Text(Strings.SignOutSettings.signOut)
+                    Spacer()
+                    Image(systemName: "chevron.right")
+                }
+                .padding(.horizontal)
+            }
+        }.buttonStyle(BlueButtonStyle())
+            .padding()
```
jankrzempek, line 50: Please be consistent and use modifier starting from new line

jankrzempek (approved):
Working fine for me, the code is dope 🫡

AirCasting/Settings/SettingsMyAccountViewModel.swift
```diff
@@ -0,0 +1,80 @@
...
+                    return
+                }
+                self.changeUserState(to: .deletingAccount)
+                self.deleteController.deleteAccount { result in
+                    self.changeUserState(to: .idle)
+                    switch result {
+                    case .success(_): break
+                    case .failure(let error):
+                        self.showAlert(InAppAlerts.failedDeletingAccount())
+                        assertionFailure("Failed to delete account: \(error)")
```
pawelgil, line 50: we shouldn't do assertion here. Asserts are designed for defining expectations (asserting ;) ), usually guarding ourselves from programming errors, like here:
```Swift
func divide(_ number: Int, by divider: Int) -> Double {
    assert(divider != 0)
    return ...
}
```

this `deleteButtonTapped()` code in here is a runtime failure condition that is not caused by programming error, it can just happen due to outside conditions. In this case I think an error log would be sufficient

pawelgil (approved):

PR #761, opened 2022-06-20
Threshold external - refactor

jankrzempek:


AirCasting/SearchAndFollow/AirBeamMeasurements/AirBeamMeasurementsDownloader.swift
```diff
@@ -4,7 +4,7 @@ import SwiftUI
 import Resolver
 
 protocol AirBeamMeasurementsDownloader {
-    func downloadStreams(with sessionID: Int, for airbeam: AirBeamStreamPrefix, completion: @escaping (Result<[MeasurementsDownloaderResultModel], Error>) -> Void) throws
```
SzymonGorowski, line 7: I think now `AirBeamStreamPrefix ` is not used anywhere.

AirCasting/SearchAndFollow/AirBeamMeasurements/AirBeamStreamSuffixes.swift
```diff
@@ -19,4 +19,14 @@ enum AirBeamStreamSuffixes: CaseIterable {
         case .RH: return "rh"
         }
     }
+
```
SzymonGorowski, line 22: I think `rawName` is unused currently.

SzymonGorowski (changes requested):
Good job 🏅 , but Ozon for OpenAQ doesn't show values. Except that everything works as expected.

SzymonGorowski (approved):
That was a fast fix 🚀 it works!

AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift
```diff
@@ -215,39 +200,26 @@ class CompleteScreenViewModel: ObservableObject {
...
-                }
+    private func createExternalSession(with sortedStreams: [MeasurementsDownloaderResultModel.Stream]) {
+        DispatchQueue.main.async {
+            self.externalSessionWithStreams = self.service.createExternalSession(from: self.session, with: sortedStreams)
+            
+            self.sessionStreams = .ready(self.externalSessionWithStreams!.streams.map {
+                .init(id: $0.id,
+                      sensorName: Self.getSensorName($0.sensorName),
+                      sensorUnit: $0.unitSymbol,
+                      lastMeasurementValue: $0.measurements.last?.value ?? 0,
```
anna1901, line 211: I know this code was already here but I think we should add some guard that measurements are not empty and if they are we should indicate that in the view instead of showing 0. Maybe it would be a good idea to create a ticket for that?

AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift
```diff
@@ -215,39 +200,26 @@ class CompleteScreenViewModel: ObservableObject {
...
+                      measurements: $0.measurements.map({.init(value: $0.value,
+                                                               time: $0.time,
+                                                               latitude: $0.latitude,
+                                                               longitude: $0.longitude)}),
+                      thresholds: $0.thresholdsValues)
+            })
+            
+            if let stream = self.externalSessionWithStreams!.streams.first {
+                self.assignValues(with: stream)
+                self.defineChartRange(with: stream)
```
anna1901, line 222: I think a more proper name for this function would be `generateChartEntires`.

PR #762, opened 2022-06-20
Edge case in Thresholds Settings

SzymonGorowski:
https://trello.com/c/LqRBfmZh/168-edge-case-handle-situations-when-thresholdveryhigh-thresholdverylow-0

### DIFF
--- a/AirCasting/SessionViews/ThresholdSettings/ThresholdsSettingsViewModel.swift
+++ b/AirCasting/SessionViews/ThresholdSettings/ThresholdsSettingsViewModel.swift
@@ -14,7 +14,8 @@ class ThresholdSettingsViewModel: ObservableObject {
     let initialThresholds: ThresholdsValue
     var threshold: SensorThreshold
     private let formatter: ThresholdFormatter
-    
+    @Published var alert: AlertInfo?
+
     init(initialThresholds: ThresholdsValue, threshold: SensorThreshold) {
         self.initialThresholds = initialThresholds
         self.threshold = threshold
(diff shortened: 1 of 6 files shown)


jankrzempek (approved):
Seems fine for me!  Great work 🥳

AirCasting/SessionViews/ThresholdSettings/ThresholdsSettingsViewModel.swift
```diff
@@ -23,15 +24,27 @@ class ThresholdSettingsViewModel: ObservableObject {
...
+    func updateToNewThresholds(completion: @escaping ((Result<ThresholdsValue, Error>) -> Void)) {
+        enum ThresholdValuesError: Error {
+            case veryLowEqualsVeryHigh
+        }
+        
+        guard thresholdVeryLow != thresholdVeryHigh else {
+            DispatchQueue.main.async {
+                self.alert = InAppAlerts.thresholdsValuesSettingsWarning()
+            }
+            completion(.failure(ThresholdValuesError.veryLowEqualsVeryHigh))
```
pawelgil, line 36: don't we want this to be run on main thread too?

pawelgil (approved):
This is an interesting case overall. Notice how we check for the `veryHigh > veryLow` condition in multiple places which is not ideal, because if anyone creates new code that presents thresholds in the future, they will probably not take this into account and introduce a bug into the code.

This is a good candidate for refactor so that we'd create a special type for thresholds which itself would guarantee data correction

anna1901 (approved):

PR #766, opened 2022-06-27
Log out - UI improved

jankrzempek:
https://trello.com/c/rX5gdP7j/420-do-logout-sync-in-background


AirCasting/AuthViews/CreateAccountView.swift
```diff
@@ -23,7 +23,10 @@ struct CreateAccountView: View {
     @State private var isUsernameBlank = false
     @State private var alert: AlertInfo?
     @State private var isLoading = false
-
+
```
SzymonGorowski, line 26: I think this space can be removed.

SzymonGorowski (changes requested):
I was unable to create an account as per the attached screenshot.

Screenshot 2022-06-28 at 10 35 21

I think you have to make some changes with `func checkIfUserInputIsCorrect()` in `CreateAccountView` 
 
 Everything else works as expected.
 
One idea from me. What do you think to add some animation when the indicator disappears and the button changes state from disabled to enabled? I know that we do not have a lot of animations in the app but I think it will look better. Instead, we will have a smooth refresh of the UI.

https://user-images.githubusercontent.com/82096200/176134629-33daf32f-3dec-45c8-bad6-382aa29ebacd.mp4

AirCasting/RootAppView.swift
```diff
@@ -13,22 +13,34 @@ struct RootAppView: View {
 
     @InjectedObject var userAuthenticationSession: UserAuthenticationSession
     @InjectedObject var lifeTimeEventsProvider: LifeTimeEventsProvider
+    @InjectedObject private var userState: UserState
+    @StateObject var SignInPersistanceObserved = SignInPersistance.shared
```
SzymonGorowski, line 17: I think you can use `private` for every property.

SzymonGorowski (approved):
Works as expected 🚀 !

AirCasting/AuthViews/CreateAccountView.swift
```diff
@@ -15,15 +15,14 @@ struct CreateAccountView: View {
...
-    @State private var username: String = ""
-    @State private var password: String = ""
     @State private var isPasswordCorrect = true
     @State private var isEmailCorrect = true
     @State private var isUsernameBlank = false
     @State private var alert: AlertInfo?
     @State private var isLoading = false
-
+    @State private var linkActive = false
+    @StateObject var SignInPersistanceObserved = SignInPersistance.shared
```
anna1901, line 24: Variable name should start with small letter I think

AirCasting/RootAppView.swift
```diff
@@ -11,24 +11,36 @@ import Resolver
...
+                    if SignInPersistanceObserved.signInActive {
+                        SignInView(completion: { self.lifeTimeEventsProvider.hasEverLoggedIn = true }).environmentObject(lifeTimeEventsProvider)
+                    } else {
+                        CreateAccountView(completion: { self.lifeTimeEventsProvider.hasEverLoggedIn = true }).environmentObject(lifeTimeEventsProvider)
+                    }
+                }
             }
         }
         .environment(\.managedObjectContext, Resolver.resolve(PersistenceController.self).viewContext) //TODO: Where is this used??
+        .onAppWentToBackground {
```
anna1901, line 41: Why did you add this here?

anna1901 (changes requested):
Great changes. I added some suggestions. Unfortunately this new UI element is cut on smaller screens (see attached screenshot). 
Simulator Screen Shot - iPhone 8 - 2022-07-01 at 15 29 36

AirCasting/RootAppView.swift
jankrzempek: So that we can prevent any credentials leak/steal thing. Without it, when you go to the background just for a while, you are able to go back and see those credentials as saved - it was not the case here. We should keep them on the screen (saved) only for the sake of allowing the user not to type them again when view refreshes - in any other case it needs to be deleted for safety

AirCasting/SignInPersistance.swift
```diff
@@ -0,0 +1,23 @@
+import Foundation
+
+class SignInPersistance: ObservableObject {
+    @Published var username: String = ""
+    @Published var password: String = ""
+    @Published var email: String = ""
+    @Published var signInActive: Bool = false
```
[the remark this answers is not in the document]
jankrzempek: Sure, maybe it will be better 🫡

AirCasting/SignInPersistance.swift
```diff
@@ -0,0 +1,23 @@
...
+    
+    public static let shared = SignInPersistance()
+    
+    func clearCredentials() {
+        username = ""
+        email = ""
+        password = ""
+    }
+
+    func clearDataWithCredentials() {
```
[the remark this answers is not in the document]
jankrzempek: I'm not so much into reset here, in some case I think `signInActive` represent data. I wil try to change it somehow, please let me know what do u think

AirCasting/AuthViews/SignInView.swift
```diff
@@ -160,14 +170,25 @@ private extension SignInView {
         .buttonStyle(BlueTextButtonStyle())
     }
     
-    var signupButton: some View {
+    var signUpLink: some View {
```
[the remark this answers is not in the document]
jankrzempek: I don't think we can do it here right now (at least we would need to change a big part of logic). We are not sure if the `CreateAccountView` will be below `SignInView` in the hierarchy right now; so dismissing that could create some unknown issues 🫣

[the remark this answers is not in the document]
anna1901: @jankrzempek is there any scenario, when the CreateAccountView is not below the SignInView?

anna1901 (approved):
👏🏻

PR #768, opened 2022-06-28
Adding info labels to EditSessionView

SzymonGorowski:
https://trello.com/c/vFLVu4mA/713-adding-labels-to-editsessionview

### DIFF
--- a/AirCasting/SessionViews/EditView.swift
+++ b/AirCasting/SessionViews/EditView.swift
@@ -36,18 +36,20 @@ struct EditView<VM: EditViewModel>: View {
             titleLabel
                 .padding(.bottom, 20)
             VStack(alignment: .leading) {
-                Text(Strings.EditSession.sessionNameLabel)
+                createLabel(with: Strings.EditSession.sessionNameLabel)
                 createTextfield(placeholder: Strings.EditSession.sessionNamePlaceholder,
                                 binding: $editSessionViewModel.sessionName)
+                .font(.body)
                 if editSessionViewModel.shouldShowError {
                     errorMessage(text: Strings.EditSession.erorr)
                 }
             }
             .padding(.bottom)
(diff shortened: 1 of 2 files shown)


AirCasting/SessionViews/EditView.swift
```diff
@@ -77,4 +82,10 @@ struct EditView<VM: EditViewModel>: View {
             presentationMode.wrappedValue.dismiss()
         }.buttonStyle(BlueTextButtonStyle())
     }
+    
+    private var customSpacer: some View {
+        Rectangle()
+            .fill(Color(UIColor.systemBackground))
+            .frame(height: 10)
+    }
```
jankrzempek, line 90: Why not just a padding instead?

SzymonGorowski: This is what happen when I added .`padding()`

Screenshot 2022-06-28 at 12 57 15

jankrzempek: 😂 I was going to say nothing

anna1901: @SzymonGorowski padding or spacer? 🤔

SzymonGorowski: @jankrzempek & @anna1901 thank you for digging 🤦  I realized that I put the `.padding()` in the wrong place.

anna1901 (changes requested):
Please change the fonts size according to designs. For labels we have to guess what would be good, but maybe there are some labels in the designs that can serve as an inspiration.

If you group labels with textFields and error messages to separate variables and put the in a VStack with some spacing (like this: `VStack(spacing: 20)`), wouldn't it remove the necessity for creating this custom spacer?

anna1901 (approved):

PR #772, opened 2022-06-28
SelectDevice View UI changes

SzymonGorowski:
https://trello.com/c/97cN2ttT/716-selectdevice-view-adding-indicator-to-the-buttons


anna1901 (changes requested):
Since those labels are almost identical it seems like they could be changed to function that takes a string and returns a label view. But I'm leaving that for your consideration.

anna1901 (approved):
Sorry, I see that those indicators are not centered on the design. So I'm leaving that for your consideration also.

jankrzempek (approved):
Cooooool!

PR #784, opened 2022-07-05
Progress View improved - SD Sync flow

jankrzempek:
Progress View inside the AB sync flows equally divided to look well visually.

**Where:**
This PR only covers the ProgressBar inside the `Sync data from AB`.

**What:**
ProgressBar is divided into 7 parts. The value corresponding to each screen was assigned to it.

https://trello.com/c/CJ4oNJ95/728-improved-progressbar-sync-data-from-ab

### DIFF
--- a/AirCasting/SDSync/ViewsFlow/SDSyncRootView/SDSyncRootView.swift
+++ b/AirCasting/SDSync/ViewsFlow/SDSyncRootView/SDSyncRootView.swift
@@ -10,7 +10,7 @@ struct SDSyncRootView: View {
     
     var body: some View {
         VStack(spacing: 40) {
-            ProgressView(value: 0.1)
+            ProgressView(value: 0.142)
             Spacer()
             ZStack(alignment: Alignment(horizontal: .trailing, vertical: .bottom), content: {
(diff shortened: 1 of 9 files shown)


SzymonGorowski (approved):
The code looks good. 👍  I am unable to test it.

AirCasting/SDSync/ViewsFlow/SDSyncRootView/SDSyncRootView.swift
```diff
@@ -10,7 +10,7 @@ struct SDSyncRootView: View {
     
     var body: some View {
         VStack(spacing: 40) {
-            ProgressView(value: 0.1)
+            ProgressView(value: 0.142)
```
anna1901, line 13: I'm not sure how you ended up with those values, but what do you think about changing them to something like `1.0/6` (where 6 is the number of screens in this flow and 1 is the number of the particular screen)? It would be more clear for a person reading the code in my opinion.

anna1901 (changes requested):

anna1901 (approved):

PR #785, opened 2022-07-05
Images in the SD Sync process alligned and rezised.

jankrzempek:
Aim here was to make those images of AB closer in size so that they match. Also: CTRL + I here 😄

https://trello.com/c/0mqDppDM/729-match-ab-images-in-sd-sync


SzymonGorowski (approved):
The code looks great, but I am unable to test it. 🎖️

anna1901 (approved):
👏🏻

PR #792, opened 2022-07-08
Fixing cut off text on UnplugAB View

SzymonGorowski:
https://trello.com/c/n8RqGgos/735-sd-sync-text-cut-off


anna1901 (commented):
I think the text is too small this way and it looks weird. Please make the images smaller or move the button down on each screen (eg. by removing spacer and adding padding instead)

PR #795, opened 2022-07-11
Setting section which allows to switch between 12/24 hour format.

jankrzempek:
Applied 12h format to DateFormatters which we currently use. After that conditionally decide which one is proper at the given moment. The default state for this setting ON after installation.

https://trello.com/c/PcNTJiAE/734-settings-use-24-hour-format-toggled-on-by-default

### DIFF
--- a/AirCasting/AuthViews/UserSettings.swift
+++ b/AirCasting/AuthViews/UserSettings.swift
@@ -13,6 +13,7 @@ class UserSettings: ObservableObject {
     @Injected private var featureFlagProvider: FeatureFlagProvider
     private let convertToCelsiusKey = Constants.UserDefaultsKeys.convertToCelsius
     private let satteliteMapKey = Constants.UserDefaultsKeys.satelliteMapKey
+    private let twentyFourHourFormatKey = Constants.UserDefaultsKeys.twentyFourHoursFormatKey
 
     var contributingToCrowdMap: Bool {
         get {
@@ -68,6 +69,17 @@ class UserSettings: ObservableObject {
             objectWillChange.send()
         }
     }
+    
+    var twentyFourHour: Bool {
+        get {
+            userDefaults.bool(forKey: twentyFourHourFormatKey)
+        }
+        set {
+            userDefaults.setValue(newValue, forKey: twentyFourHourFormatKey)
+            Log.info("Changed twenty four hour format to \(twentyFourHour ? "ON" : "OFF")")
+            objectWillChange.send()
+        }
+    }
 
     init(userDefaults: UserDefaults = .standard) {
         self.userDefaults = userDefaults
@@ -78,5 +90,6 @@ class UserSettings: ObservableObject {
         disableMapping = isFeatureFlagOn ? userDefaults.bool(forKey: locationlessKey) : false
         convertToCelsius = userDefaults.bool(forKey: convertToCelsiusKey)
         satteliteMap = userDefaults.bool(forKey: satteliteMapKey)
+        twentyFourHour = userDefaults.valueExists(forKey: twentyFourHourFormatKey) ? userDefaults.bool(forKey: twentyFourHourFormatKey) : true
     }
 }
--- a/AirCasting/Graph/GraphRenderers.swift
+++ b/AirCasting/Graph/GraphRenderers.swift
@@ -9,9 +9,8 @@ class TimeAxisRenderer: XAxisRenderer {
     @InjectedObject private var userSettings: UserSettings
     var rightMarginOffset: CGFloat = 50
     var dateFormatter: DateFormatter {
-        if userSettings.twentyFourHour { return DateFormatters.TimeAxisRenderer.shortUTCDateFormatter }
-        rightMarginOffset = 60
-        return DateFormatters.TimeAxisRenderer.shortUTCDate12hFormatter
+        if !userSettings.twentyFourHour { rightMarginOffset = 70 }
+        return DateFormatters.TimeAxisRenderer.shared.shortUTCDateFormatter
     }
     
     override func drawLabels(context: CGContext, pos: CGFloat, anchor: CGPoint) {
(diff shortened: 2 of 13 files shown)


AirCasting/SearchAndFollow/CompleteScreen/Views/Header/StaticSessionHeader.swift
```diff
@@ -53,7 +55,10 @@ private extension StaticSessionHeader {
     }
 
     func adaptTimeAndDate() -> Text {
-        let formatter = DateFormatters.SessionCartView.utcDateIntervalFormatter
+        var formatter: DateIntervalFormatter {
+            if userSettings.twentyFourHour { return DateFormatters.SessionCardView.utcDateIntervalFormatter }
+            return DateFormatters.SessionCardView.utcDateInterval12hFormatter
+        }
```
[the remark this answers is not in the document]
jankrzempek: I was thinking about that and wasn't convinced, because of the fact that we would have 'kind of tricky' `if else` in our `Formatters`. Those should be simple and not have additional logic beneath which user is not aware of at first use -- _just a comment of my mind process_. I will try to apply your approach ASAP!

AirCasting/AuthViews/UserSettings.swift
```diff
@@ -68,6 +69,17 @@ class UserSettings: ObservableObject {
...
+    var twentyFourHour: Bool {
+        get {
+            userDefaults.bool(forKey: twentyFourHourFormatKey)
+        }
+        set {
+            userDefaults.setValue(newValue, forKey: twentyFourHourFormatKey)
+            Log.info("Changed twenty four hour format to \(twentyFourHour ? "ON" : "OFF")")
+            objectWillChange.send()
+        }
+    }
```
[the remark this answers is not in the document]
jankrzempek: In the meantime when everyone is asleep... 🫡

AirCasting/SearchAndFollow/CompleteScreen/Views/Header/StaticSessionHeader.swift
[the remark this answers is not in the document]
pawelgil: > I was thinking about that and wasn't convinced, because of the fact that we would have 'kind of tricky' `if else` in our `Formatters`. Those should be simple and not have additional logic beneath which user is not aware of at first use -- _just a comment of my mind process_. I will try to apply your approach ASAP!

That is true if we treat `DateFormatters` as a kind of "static library" (so to say). If it bothers you to put logic in there you can introduce another layer that will be used (Injected) here which will contain logic. It would be a bit cleaner, I agree. Both approaches are good for this project tho, you can pick whatever you feel like. As long as the logic is moved away from views :)

AirCasting/Graph/GraphRenderers.swift
```diff
@@ -9,9 +9,8 @@ class TimeAxisRenderer: XAxisRenderer {
     @InjectedObject private var userSettings: UserSettings
     var rightMarginOffset: CGFloat = 50
     var dateFormatter: DateFormatter {
-        if userSettings.twentyFourHour { return DateFormatters.TimeAxisRenderer.shortUTCDateFormatter }
-        rightMarginOffset = 60
-        return DateFormatters.TimeAxisRenderer.shortUTCDate12hFormatter
+        if !userSettings.twentyFourHour { rightMarginOffset = 70 }
```
pawelgil, line 12: it would be awesome to not depend on userSettings here for this offset. Not sure if achievable, feel free to disregard

AirCasting/Utils/DataFormatters.swift
```diff
@@ -2,9 +2,47 @@
 //
 
 import Foundation
+import Resolver
 
 enum DateFormatters {
     
+    struct SessionCardView {
+        @InjectedObject private var userSettings: UserSettings
+        static let shared = SessionCardView()
```
pawelgil, line 11: this is ok, but for singletons we additionally want to disable the possibility to create `SessionCardView`. Right now there is no problem with doing that in the rest of the codebase (just `SessionCardView()`). To do so we make the `init` private:

```Swift
private init() { }
```

AirCasting/Utils/DataFormatters.swift
```diff
@@ -2,9 +2,47 @@
...
+        
+        var pollutionChartDateFormatter: DateFormatter {
+            let df = DateFormatter()
```
pawelgil, line 14: Applies to all methods in `SessionCardView` and `TimeAxisRenderer`

Creating date formatters is a heavy task, we should keep them static and only choose which one to use based on `userSettings`, not create them each time from scratch

pawelgil (changes requested):

AirCasting/Utils/DataFormatters.swift
```diff
@@ -2,9 +2,82 @@
...
+        }
+        
+        static let pollutionChartDateFormatter24: DateFormatter = {
+            let df = DateFormatter()
+            df.dateFormat = "HH:mm"
+            df.timeZone = TimeZone(abbreviation: "UTC")
+            return df
+        }()
+        
+        static let pollutionChartDateFormatter12: DateFormatter = {
```
pawelgil, line 30: can be private (?)

AirCasting/Utils/DataFormatters.swift
```diff
@@ -2,9 +2,82 @@
...
+        @InjectedObject private var userSettings: UserSettings
+        static let shared = TimeAxisRenderer()
+        
+        private init() { }
+        
+        var shortUTCDateFormatter: DateFormatter {
+            userSettings.twentyFourHour ? Self.shortUTCDateFormatter24 : Self.shortUTCDateFormatter12
+        }
+        
+        static let shortUTCDateFormatter12: DateFormatter = {
```
pawelgil, line 64: can be private (?)

pawelgil (approved):

PR #800, opened 2022-07-12
Set default value for selected stream as PM2.5.

jankrzempek:
Aim here is to always show PM2.5 stream as the selected one after initial loading - everywhere (external, internal, S&F). 

https://trello.com/c/USttWn47/738-make-pm25-default-for-both-external-sessions-and-internal

### DIFF
--- a/AirCasting/CoreData/ExternalSessionEntity+CoreDataProperties.swift
+++ b/AirCasting/CoreData/ExternalSessionEntity+CoreDataProperties.swift
@@ -45,13 +45,6 @@ extension ExternalSessionEntity {
         
         return abStreams.isEmpty ? allStreams : abStreams
     }
-    
-    func defaultStreamSelection() -> MeasurementStreamEntity? {
-        allStreams.first { stream in
-            guard let name = stream.sensorName else { return false }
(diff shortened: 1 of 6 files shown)


AirCasting/CoreData/ExternalSessionEntity+CoreDataProperties.swift
```diff
@@ -45,6 +45,13 @@ extension ExternalSessionEntity {
         
         return abStreams.isEmpty ? allStreams : abStreams
     }
+    
+    func defaultStreamSelection() -> MeasurementStreamEntity? {
```
pawelgil, line 49: I would put it in an extension which lives near the View file that this function is used in. This is a model file, it should not contain view-related stuff.

Applies to 2 other instances too (`SessionEntity` and `ExternalSessionWithBlaBla`)

pawelgil (approved):

AirCasting/CoreData/ExternalSessionEntity+CoreDataProperties.swift
jankrzempek: Sure, sounds right

PR #802, opened 2022-07-12
SD Sync saving performance (crash) fix

pawelgil:
https://trello.com/c/VeJ8oOaL/707-sd-sync-fails-when-sd-card-contains-lots-of-data


AirCasting/ABConnector/SDSyncController.swift
```diff
@@ -68,13 +69,16 @@ class SDSyncController {
...
+                            Log.verbose("[TEST] Files verified.")
                             self.handle(files: verifiedFiles, sensorName: sensorName, completion: completion)
                         case .failure(let error):
+                            Log.verbose("[TEST] File verification failed. \(error.localizedDescription)")
                             Log.error(error.localizedDescription)
                             completion(false)
                         }
                     }
                 case .failure:
+                    Log.verbose("[TEST] Failed to download SD card data from AB")
```
pawelgil, line 81: can we make this into Log.error with proper logging?

Also I think `[TEST]` logs can be removed (including other files). We can review them and leave those that we want to have in logs (switching them to `Log.info`)

AirCasting/SDSync/MobileSessions/SDCardMobileSessionsSavingService.swift
```diff
@@ -71,39 +65,53 @@ class SDCardMobileSessionsSavingService: SDCardMobileSessionssSaver {
...
+        measurementStreamStorage.accessStorage { storage in
+            do {
+                var existingStreamID = try storage.existingMeasurementStream(sdStream.sessionUUID, name: sdStream.name.rawValue)
+                if existingStreamID == nil {
+                    let measurementStream = self.createMeasurementStream(for: sdStream.name, sensorPackageName: deviceID)
+                    existingStreamID = try storage.saveMeasurementStream(measurementStream, for: sdStream.sessionUUID)
+                }
+                try storage.addMeasurements(measurements, toStreamWithID: existingStreamID!)
+            } catch {
+                Log.info("Saving measurements failed: \(error)")
```
pawelgil, line 88: can be moved to `Log.error`

AirCasting/SDSync/MobileSessions/SDCardMobileSessionsSavingService.swift
```diff
@@ -119,4 +127,12 @@ class SDCardMobileSessionsSavingService: SDCardMobileSessionssSaver {
     private func createMeasurementStream(for sensorName: MeasurementStreamSensorName, sensorPackageName: String) -> MeasurementStream {
         MeasurementStream(sensorName: sensorName, sensorPackageName: sensorPackageName)
     }
+    
+    private func getExistingSession(sessionUUID: SessionUUID) -> SessionEntity? {
+        var session: SessionEntity?
+        measurementStreamStorage.accessStorage { storage in
+            session = try? storage.getExistingSession(with: sessionUUID)
+        }
+        return session
```
pawelgil, line 136: It's not safe to return CoreData entities outside of it's NSManagedObjectContext queue

AirCastingTests/Regressions/SDSync/SDSyncRegressions.swift
```diff
@@ -0,0 +1,130 @@
...
+        Resolver.test.register { self.fixedServices as MeasurementUpdatingService }
+        Resolver.test.register { self.persistence as PersistenceController }
+    }
+    
+    //
+    // Regression source:
+    // https://trello.com/c/VeJ8oOaL
+    // SD sync fails when SD card contains lots of data
+    //
+    func test_whenAirbeamHasLotsOfMeasurementsToSync_theAppCrashes() throws {
```
pawelgil, line 42: this only worked once for me and it's sitting in the same target as our other tests, so it will delay tests execution by 1m40s... Not sure what should we do with that, but it shouldn't go into development branch. Maybe comment out?

pawelgil (commented):
We need to resolve some stuff before merging in. I wasn't able to do a deep CD because of the code complexity, we should perform some tests

AirCasting/SDSync/MobileSessions/SDCardMobileSessionsSavingService.swift
jankrzempek: Got it, changes everything to `completion`. Not sure if it can be than easier and I understand it is not the prettiest solution :.

PR #808, opened 2022-07-14
Add missing RELEASE preprocessor flags to project settings

pawelgil:
This should resolve problems with no logs and errors in crashlytics


AirCasting.xcodeproj/project.pbxproj
```diff
@@ -3321,6 +3321,7 @@
 				ENABLE_STRICT_OBJC_MSGSEND = YES;
 				GCC_C_LANGUAGE_STANDARD = gnu11;
 				GCC_NO_COMMON_BLOCKS = YES;
+				GCC_PREPROCESSOR_DEFINITIONS = "RELEASE=1";
```
jankrzempek, line 3324: Out of curiosity, why not only "RELEASE" but "RELEASE=1"?

jankrzempek (approved):
🥳

SzymonGorowski (approved):
👍

AirCasting.xcodeproj/project.pbxproj
pawelgil: @jankrzempek sorry for late response, I still don't get emails from github (mistery...)
In my opinion this is a good practice when dealing with GCC preprocessor. The reason is that if we do `-DRELEASE=1` we'll be able to do a classic `#ifdef RELEASE`, but also we make sure `#if RELEASE` will too be included

PR #810, opened 2022-07-18
On finish button tap - direct user to following section.

jankrzempek:
https://trello.com/c/F36PtkM9/759-181-7-sf-map-tapping-finish-navigates-to-the-last-tab-user-was-on-should-always-navigate-to-the-following-tab


anna1901 (approved):
Cool! works well :)

PR #813, opened 2022-07-18
Block BT connection while recording/connected to AB

jankrzempek:
We should only focus here on blocking session which are mobile active.

https://trello.com/c/95zNudeL/411-block-connecting-to-ab-when-you-have-an-active-mobile-session-on-this-ab


SzymonGorowski (approved):
🎉

SzymonGorowski (changes requested):
Sorry for the previous approval. I made it before I tried to test it. I am unable to create a mobile session even if I don't record anything as per attached screenshot.

Screenshot 2022-07-19 at 09 52 48

AirCasting/Utils/Connection/BluetoothConnection.swift
```diff
@@ -0,0 +1,53 @@
...
+import Resolver
+
+final class BluetoothConnectionProtector: Connectable, SessionsFetchable {
+    
+    enum BluetoothConnectionProtectorError: Error {
+        case alreadyConnected
+        case readError(Error)
+    }
+    
+    @Injected private var persistenceController: PersistenceController
```
pawelgil, line 12: if you opt to use `SessionsFetchable`, you won't need this (we will be moving to a higher abstraction that hides all that CoreData complexity)

AirCasting/Utils/Connection/BluetoothConnection.swift
```diff
@@ -0,0 +1,53 @@
...
+                                    SessionStatus.DISCONNECTED.rawValue,
+                                    SessionStatus.NEW.rawValue)
+        fetchSessions(constrained: .predicate(predicate)) { predicated in
+            switch predicated {
+            case .success(let sessions):
+                if sessions.count > 0 {
+                    result = .success(())
+                } else {
+                    result = .failure(BluetoothConnectionProtectorError.alreadyConnected)
+                }
```
pawelgil, line 46: why not `result = sessions.count > 0 ? .success(()) : .faiulre(...)`?

Also - isn't that flipped? It will return "yes, we can connect" if it **finds** ongoing sessions, won't it?

AirCasting/Utils/Connection/Connectable.swift
```diff
@@ -0,0 +1,5 @@
+import Foundation
+
+protocol Connectable {
```
pawelgil, line 3: I like the name, but not for this one. This protocol is not for connecting per se, just for checking if a connection can be made. I think we need a better suiting name... Can't think of any good one atm tho, sorry 😅

AppDelegate+Injection.swift
```diff
@@ -161,6 +161,9 @@ extension Resolver: ResolverRegistering {
         main.register { DefaultDeleteAccountController() as DeleteAccountController }
         main.register { DefaultRemoveDataController() as RemoveDataController }
         
+        // MARK: TEST
```
pawelgil, line 164: `TEST` needs to change

pawelgil (changes requested):

AirCasting/CreateSessionViews/SelectDeviceView.swift
```diff
@@ -19,6 +19,7 @@ struct SelectDeviceView: View {
     @EnvironmentObject private var sessionContext: CreateSessionContext
     @InjectedObject private var bluetoothManager: BluetoothManager
     @InjectedObject private var microphoneManager: MicrophoneManager
+    @Injected private var bluetoothConnectionProtector: Connectable
     @Binding var creatingSessionFlowContinues : Bool
     @Binding var sdSyncContinues : Bool
     @State private var showAlert = false
```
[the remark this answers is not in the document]
jankrzempek: Zrzut ekranu 2022-07-19 o 12 58 15
Here at the bottom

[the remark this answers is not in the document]
SzymonGorowski: I meant `showAlert` 😄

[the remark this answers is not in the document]
jankrzempek: 😅😅😅

AirCasting/Utils/Connection/BluetoothConnection.swift
```diff
@@ -0,0 +1,53 @@
...
+final class BluetoothConnectionProtector: Connectable, SessionsFetchable {
+    
+    enum BluetoothConnectionProtectorError: Error {
+        case alreadyConnected
+        case readError(Error)
+    }
+    
+    @Injected private var persistenceController: PersistenceController
+    
+    func fetchSessions(constrained: Database.Constraint, completion: @escaping (Result<[Database.Session], Error>) -> Void) {
```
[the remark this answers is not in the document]
jankrzempek: It is better not to say why

AirCasting/Utils/Connection/BluetoothConnection.swift
jankrzempek: I will, but please if you could comment it as much as possible - it is still a bit tough for me

SzymonGorowski (approved):
Looks and works great 👍

AirCasting/Utils/Connection/BluetoothConnection.swift
```diff
@@ -0,0 +1,30 @@
...
+    enum BluetoothConnectionProtectorError: Error {
+        case alreadyConnected
+        case readError(Error)
+    }
+    
+    @Injected private var database: SessionsFetchable
+    
+    func isAirBeamAvailableForNewConnection(completion: @escaping (Result<Void, Error>) -> Void) {
+        let predicate = NSPredicate(format: "deviceType == %@ AND type == %@ AND (status == %li || status == %li || status == %li)",
+                                    "1",
```
anna1901, line 16: don't we have something like AirBeam.rawValue?

anna1901 (changes requested):
Would this allow the user to create multiple standalone sessions? I think not. Also, I think this ticket is about blocking creating multiple session with the same specific airbeam, and not blocking creating more than one mobile bluetooth session and I think that's what the code it doing.

jankrzempek:
> Would this allow the user to create multiple standalone sessions? I think not. Also, I think this ticket is about blocking creating multiple session with the same specific airbeam, and not blocking creating more than one mobile bluetooth session and I think that's what the code it doing.

I agree, those standalone it something that I still need to take care of. Will do it in a moment.
In addition to blocking only connection to the AirBeam which is ongoing, I cannot agree. First is, that we would need to handle that as late as connecting to AirBeam happens. It's the later part of the process and that could frustrate the user that he was not informed earlier. The most important here is the connection itself: having the ability to connect more than one AB to our phone can lead to unknown bugs. For now, only AirPlay supports two device connections at once. In our case we would need to be connected to both and receive data from them - I think it could be too much.

AirCasting/Utils/Connection/BluetoothConnection.swift
jankrzempek: Yes we do, IDK why it does not work but it does not and I was trying hard 😿

AirCasting/CreateSessionViews/SelectDeviceView.swift
```diff
@@ -61,6 +61,9 @@ struct SelectDeviceView: View {
             selected = 0
             #warning("Handle that mobileWasTapped is somehow public")
             emptyDashboardButtonTapped.mobileWasTapped = false
+            isBTButtonActive(completion: { value in
```
anna1901, line 64: How about `isBTButtonActive({ isBTButtonAvailable = $0 })? However, I don't like this buttonActive part of the name and buttonAvailable either. Maybe the name of the fuction could be isBTConnectionAvailable and variable can be called isBTButtonEnabled (or disabled)?

AirCasting/Utils/Connection/BluetoothConnection.swift
```diff
@@ -0,0 +1,29 @@
...
+final class BluetoothConnectionProtector: ConnectionProtectable {
+    
+    enum BluetoothConnectionProtectorError: Error {
+        case alreadyConnected
+        case readError(Error)
+    }
+    
+    @Injected private var database: SessionsFetchable
+    
+    func isAirBeamAvailableForNewConnection(completion: @escaping (Result<Void, Error>) -> Void) {
```
anna1901, line 14: Are we really checking if AirBeam is available here? I think the name is misleading cause we are only checking if there is an active Airbeam session

AirCastingTests/BluetoothProtectorTests.swift
```diff
@@ -0,0 +1,199 @@
+import XCTest
+import Resolver
+import Foundation
+@testable import AirCasting
+
+class BluetoothProtectorTests: ACTestCase {
+    
+    func test_whenAskedToPredicateStatuses_excludesUnsupported() throws {
```
anna1901, line 8: I don't get the name. 😢

AirCastingTests/BluetoothProtectorTests.swift
```diff
@@ -0,0 +1,199 @@
...
+        @objc let type: String
+        @objc let status: Int
+        
+        init(deviceType: String, type: String, status: Int) {
+            self.deviceType = deviceType
+            self.type = type
+            self.status = status
+        }
+        
+        static func createExpected(status: Int) -> FakePredicateBlueprint {
```
anna1901, line 193: What does `createExpected` mean in this context?

AirCastingTests/BluetoothProtectorTests.swift
```diff
@@ -0,0 +1,199 @@
...
+        sut.isAirBeamAvailableForNewConnection(completion: { _ in })
+        
+        XCTAssertEqual(database.constrainedCalls.count, 1)
+        let constraint = try XCTUnwrap(database.constrainedCalls.first)
+        guard case let .predicate(predicate) = constraint else { XCTFail(); return }
+        
+        let correctStatuses: [FakePredicateBlueprint] = [.createExpected(status: 0),
+                                                  .createExpected(status: 2),
+                                                  .createExpected(status: -1)]
+        // Status: 2 should be excluded
```
anna1901, line 22: So those aren't only "correctStatuses" anymore, right?

AirCastingTests/BluetoothProtectorTests.swift
```diff
@@ -0,0 +1,199 @@
...
+        let invalidStatuses: [FakePredicateBlueprint] = [.createExpected(status: -2),
+                                                         .createExpected(status: 3),
+                                                         .createExpected(status: 5)]
+        
+        XCTAssertEqual(0, (invalidStatuses as NSArray).filtered(using: predicate).count)
+    }
+    
+    func test_whenAskedToPredicateDeviceTypes_excludesUnsupported() throws {
+        let database = AirBeamDatabaseSpy()
+        Resolver.test.register { database as SessionsFetchable }
```
anna1901, line 34: I think those two lines could be moved to setup func

AirCastingTests/BluetoothProtectorTests.swift
```diff
@@ -0,0 +1,199 @@
...
+        let sut = BluetoothConnectionProtector()
```
anna1901, line 35: I think this can be a class var, as it is used in each test

AirCastingTests/BluetoothProtectorTests.swift
```diff
@@ -0,0 +1,199 @@
...
+        
+        sut.isAirBeamAvailableForNewConnection(completion: { _ in })
+        
+        XCTAssertEqual(database.constrainedCalls.count, 1)
+        let constraint = try XCTUnwrap(database.constrainedCalls.first)
+        guard case let .predicate(predicate) = constraint else { XCTFail(); return }
+        
+        let invalidSessionType: [FakePredicateBlueprint] = [.init(deviceType: "1", type: SessionType.fixed.rawValue, status: 1),
+                                                            .init(deviceType: "1", type: "", status: 1),
+                                                            .init(deviceType: "1", type: "MonileSession", status: 1)]
```
anna1901, line 63: 😅

AirCastingTests/BluetoothProtectorTests.swift
```diff
@@ -0,0 +1,199 @@
...
+        
+        XCTAssertEqual(0, (invalidSessionType as NSArray).filtered(using: predicate).count)
+    }
+    
+    func test_failure_whenSessionIsActive() {
```
anna1901, line 68: I think the name format should be test_when_then to be consistent

AirCastingTests/BluetoothProtectorTests.swift
```diff
@@ -0,0 +1,199 @@
...
+        }
+        let resultError = TestError()
+        let database = AirBeamDatabseStub(toReturn: .failure(resultError))
+        Resolver.test.register { database as SessionsFetchable }
+        let sut = BluetoothConnectionProtector()
+        
+        sut.isAirBeamAvailableForNewConnection { result in
+            switch result {
+            case .success(_):
+                break // This is expected path
```
anna1901, line 122: Is it? 🕵🏻‍♀️

anna1901 (changes requested):
I hate to break this to you, but there is another part of work that needs to be done. We shouldn't allow users to record mobile session on a device that is recording a session in standalone mode. I think that's what is included in this ticket.

mheimbinder:
This task is only about blocking connecting to a second AirBeam when you
are currently connected to an AirBeam of the same model (1,2,3). It should
NOT:

1) block connecting to multiple BT devices, as long as they have different
names, i.e. connecting to 1 AB2 and 1 AB3 simultaneously should be allowed;

2) block connecting to a second AB3 when the first AB3 is not connected
because it's been placed in standalone mode, i.e. you should be able to
configure AB3 #1 into standalone mode and then continue to place AB3 #2,
#3, #4 etc into standalone mode one after the other.

-Michael-

AirCasting/CreateSessionViews/SelectDeviceView.swift
```diff
@@ -16,12 +16,13 @@ struct SelectDeviceView: View {
     @State private var isTurnOnBluetoothLinkActive: Bool = false
     @State private var isPowerABLinkActive: Bool = false
     @State private var isMicLinkActive: Bool = false
+    @State private var isBTButtonAvailable: Bool = false
```
pawelgil, line 19: small digression: "is" prefix is not needed for this kind of properties, see: `btButtonAvailable: bool` says it all already. Ofc don't change it

AirCasting/CreateSessionViews/SelectDeviceView.swift
pawelgil: I agree, `isConnectionAvailable` is better, currently I can't really tell what's going on here, why is button active or inactive, changing func name to `isAirBeamConnectionAvailable` or something will make it super clear

mheimbinder:
Hi Michael,

My comment in this pull request is about this case:
Someone is recording a mobile session on a specific AB3. They put this session in standalone mode. They are trying to record another mobile AB session on this AB3.

We shouldn’t allow them to connect to that AB3 which is recording a standalone session. Is that right? Currently, an Airbeam which is in standalone mode is still listed as a device to which we can connect.

Ania

mheimbinder:
Hi Ania.  Thanks for following up.  You are correct!

-Michael-

--------------------
Michael Heimbinder (he/him/his)
Executive Director, HabitatMap
347.410.9499
www.habitatmap.org
www.aircasting.org
www.aircastingactions.org
www.takingspace.org
twitter.com/habitatmap

AirCasting/CreateSessionViews/SelectDeviceView.swift
```diff
@@ -84,6 +87,10 @@ struct SelectDeviceView: View {
             createLabel(title: Strings.SelectDeviceView.bluetoothLabel, subtitle: Strings.SelectDeviceView.bluetoothDevice)
         })
         .buttonStyle(WhiteSelectingButtonStyle(isSelected: selected == 1))
+        .disabled(!isBTButtonAvailable)
+        .onTapGesture {
+            alert = InAppAlerts.bluetoothSessionAlreadyRecordingAlert()
```
pawelgil, line 92: Will it do what I think it will do? When user taps a disabled button it will show an alert? Is that good UX? I'm not sure... I think very few users would actually tap on a disabled button. If we want to explain why the button is disabled we should add a label.

disregard if this is not the case here (IDK how `WhiteSelectingButtonStyle` looks like TBH)

AirCasting/Utils/Connection/BluetoothConnection.swift
```diff
@@ -0,0 +1,29 @@
+import SwiftUI
+import CoreData
+import Resolver
+
+final class BluetoothConnectionProtector: ConnectionProtectable {
```
pawelgil, line 5: rename the file plz to refect class name

AirCasting/Utils/Connection/BluetoothConnection.swift
pawelgil: @anna1901 I think this name is accurate. On the implementation level indeed we're just checking for existing database AB session. But we're operating on a higher abstraction level here. We don't want clients of this protocol to be aware of any of this, they just want to check if they can connect to AB, they don't care how it's checked behind the scenes.

Imagine a hypothetical scenario where we're asked to disable connecting to older AB models. With this protocol and this function name we'd be able to just add another class that conforms to `ConnectionProtectable`, combine it with this one inside resolver registration (using a [composite pattern](https://refactoring.guru/design-patterns/composite)) and that's it, no existing code would be modified, very clean and extensible.
Now if we named this function `checkForExistingABSessions` we'd lose all of that flexibility. Just because of the name 🤯

AirCastingTests/BluetoothProtectorTests.swift
pawelgil: agreed. This test needs to be broken doen into two. One for correct statuses and one for incorrect ones. Also better naming would be great. Try to put yourself on a position of someone who don't know the implementation details, just runs the tests to check if they messed something up. If a test fails, it should communicate what is broken just from reading the name (ideally, not always possible. But most of the time). Also - a single test should fail only for a single reason. This test will fail if either the predicate excludes correct status or it includes incorrect ones - this is why we need to split it here

AirCastingTests/BluetoothProtectorTests.swift
pawelgil: I think we use Spy/Stub classes to double for `SessionsFetchable`, so it wouldn't work because we're switching between the two. What we can do is to move those Stub/Spy to properties.

AirCastingTests/BluetoothProtectorTests.swift
pawelgil: We can, but we need to remember that we need to do `Resolver.test.register { ... }` first. I think making `sut` a `lazy var` would do the trick as long as we're registering dependencies first thing in each test

AirCastingTests/BluetoothProtectorTests.swift
```diff
@@ -0,0 +1,199 @@
...
+        guard case let .predicate(predicate) = constraint else { XCTFail(); return }
+        
+        let invalidDeviceType: [FakePredicateBlueprint] = [.init(deviceType: "-1", type: "MobileSession", status: 1),
+                                                           .init(deviceType: "0", type: "MobileSession", status: 1),
+                                                           .init(deviceType: "2", type: "MobileSession", status: 1)]
+        
+        XCTAssertEqual(0, (invalidDeviceType as NSArray).filtered(using: predicate).count)
+    }
+    
+    func test_whenAskedToPredicateSessionTypes_excludesUnsupported() throws {
```
pawelgil, line 50: Isn't this test repeating a part of the first one?

AirCastingTests/BluetoothProtectorTests.swift
pawelgil: good catch, we should `XCTFail()` here

pawelgil (changes requested):

AirCasting/Utils/Connection/BluetoothConnection.swift
jankrzempek: I came up with sth, we will see what you will say 😅

AirCasting/CreateSessionViews/SelectDeviceView.swift
jankrzempek: Thanks! It will be changed

AirCastingTests/BluetoothProtectorTests.swift
```diff
@@ -0,0 +1,267 @@
+import XCTest
+import Resolver
+import Foundation
+@testable import AirCasting
+
+class BluetoothProtectorTests: ACTestCase {
+    let databaseSpy = AirBeamDatabaseSpy()
+    lazy var sut = BluetoothConnectionProtector()
+    
+    func testWrongStatuses_whenAskedToPredicate_shouldResultWithZero() throws {
```
pawelgil, line 10: applies to all tests here:
Try to make names of tests less focused on what  the test is doing and more on what production code is doing and expectations. When naming tests imagine that someone only gets test report and so they only see the failing test name - it would be hard to immediately tell what failed here. Maybe something like:
`test_whenAsksDatabaseForExistingSessions_requestsOnlyMobileSessions` (not sure if that's true here, but hopefully you get the idea)

AirCasting/Utils/Connection/BluetoothConnectionProtector.swift
```diff
@@ -0,0 +1,32 @@
...
+    
+    func isAirBeamAvailableForNewConnection(peripheraUUID: String, completion: @escaping (Result<Void, Error>) -> Void) {
+        let defaultPredicate =  NSPredicate(format: "deviceType == %@ AND type == %@",
+                                            "1", SessionType.mobile.rawValue)
+        let p1 = NSPredicate(format: "status == %li || status == %li",
+                             SessionStatus.RECORDING.rawValue, SessionStatus.NEW.rawValue)
+        let p2 = NSPredicate(format: "status == %li AND peripheralUUID == %@",
+                             SessionStatus.DISCONNECTED.rawValue, peripheraUUID)
+        let predicate = NSCompoundPredicate(orPredicateWithSubpredicates: [p1, p2])
+        let predicateWithDefault = NSCompoundPredicate(andPredicateWithSubpredicates: [defaultPredicate, predicate])
```
pawelgil, line 22: nice! some advanced predicate action here! 🤩

are all cases covered in tests? 🕵️‍♂️

AirCasting/CoreData/SessionEntity.swift
```diff
@@ -18,7 +18,9 @@ public class SessionEntity: NSManagedObject, Identifiable {
     @nonobjc public class func fetchRequest() -> NSFetchRequest<SessionEntity> {
         return NSFetchRequest<SessionEntity>(entityName: "SessionEntity")
     }
-
+    
+    @NSManaged public var peripheralUUID: String?
```
pawelgil, line 22: That's a bit fishy since this property only makes sense for certain type of sessions (for a mic session it doesn't make any sense).

Currently I cannot tell you what the correct design would look like since there are moving parts here, you can try to come up with a solution or we can schedule a call to work on it together (we can also skip and just accept that technical debt, it's not a huge deal)

pawelgil (changes requested):
Looks nice, let's rename some tests to be more clear and perhaps rethink this peripheralUUID addition to SessionEntity

pawelgil (approved):

anna1901 (changes requested):

AirCasting/CoreData/Storages/SessionStorage.swift
```diff
@@ -1,40 +1,53 @@
...
+// Created by Lunar on 09/08/2022.
 //
-
-import Foundation
 import CoreData
 import Resolver
 
-final class SessionDataEraser: ObservableObject {
-    @Injected private var persistenceController: PersistenceController
+protocol SessionStorage {
```
anna1901, line 6: I think this should not be called just "SessionStorage" if it has just one function which is related to clearing bluetooth connection. The name of this protocol should be more specific

AirCasting/CreateSessionViews/ConnectingABView.swift
```diff
@@ -1,22 +1,23 @@
...
+import Resolver
 
 struct ConnectingABView<VM: AirbeamConnectionViewModel>: View {
     
-    @Environment(\.presentationMode) var presentationMode
     @StateObject var viewModel: VM
     @Binding var creatingSessionFlowContinues: Bool
     @State private var showNextScreen: Bool = false
-    @State private var presentAlert: Bool = false
+    @Injected private var mobilePeripheralSessionManager: MobilePeripheralSessionManager
```
anna1901, line 14: This is unneeded

anna1901 (approved):
Works well 🚀

PR #817, opened 2022-07-25
Explore sessions button - fix

jankrzempek:
When you tap on the explore session button and then go to `create session` screen - S&F screen will always appear.

https://trello.com/c/syJ6K83c/739-navigates-to-sf


SzymonGorowski (approved):
Works well. 👍

anna1901 (approved):
Good catch!

PR #818, opened 2022-07-25
Make image smaller so that user can access GetStarted button.

jankrzempek:
https://trello.com/c/DRsxTWfP/764-ipad-landscape-intro-wizard-first-screen-cut-off-cant-continue

🚩This is the temporary, not ideal fix. As soon as Gosia will be back, I will reach out to her to get images designed specifically for iPad.

### DIFF
--- a/AirCasting/Assets.xcassets/Bitmap.imageset/Contents.json
+++ b/AirCasting/Assets.xcassets/Bitmap.imageset/Contents.json
@@ -13,6 +13,14 @@
       "filename" : "Bitmap@3x.png",
       "idiom" : "universal",
       "scale" : "3x"
+    },
+    {
+      "idiom" : "ipad",
+      "scale" : "1x"
+    },
+    {
+      "idiom" : "ipad",
+      "scale" : "2x"
     }
   ],
   "info" : {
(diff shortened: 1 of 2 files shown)


SzymonGorowski (approved):
Approved 😄

anna1901 (changes requested):
I don't like that it is stretched like that. I think it would be better to change how this image is added to this view and add it as a background image instead

jankrzempek:
> I don't like that it is stretched like that. I think it would be better to change how this image is added to this view and add it as a background image instead

What do you mean by background, could you please elaborate or give an example?

anna1901 (commented):
Add image as a background image using `.background()`. I'm not sure how I can elaborate on this. We can have a call if I should provide you some more information.

AirCasting/Assets.xcassets/Bitmap.imageset/Contents.json
```diff
@@ -13,6 +13,14 @@
       "filename" : "Bitmap@3x.png",
       "idiom" : "universal",
       "scale" : "3x"
+    },
+    {
+      "idiom" : "ipad",
```
anna1901, line 18: This shouldn't be here, right?

AirCasting/Onboarding/GetStarted.swift
```diff
@@ -11,18 +11,25 @@ struct GetStarted: View {
...
+            logoImage
             descriptionText
             if featureFlagsViewModel.enabledFeatures.contains(.searchAndFollow) {
                 continueToAirNearYouScreenButton
             } else {
                 startButton
             }
-            Spacer()
-        }
+        }.background(
```
anna1901, line 22: new line?

PR #819, opened 2022-07-25
Changing selectedSection_ in DashboardView name to allow better understanding

SzymonGorowski:
https://trello.com/c/yciWEUJV/766-renaming-property-in-dashboardview-to-be-more-clear


AirCasting/Dashboard/DashboardView.swift
```diff
@@ -15,7 +15,7 @@ struct DashboardView: View {
     @StateObject var coreDataHook: CoreDataHook
     @FetchRequest<SensorThreshold>(sortDescriptors: [.init(key: "sensorName", ascending: true)]) var thresholds
     // TODO: - We can rethink the way how "Select Section" works and change it
-    @EnvironmentObject var selectedSection_: SelectSection
+    @EnvironmentObject var selectedSectionNeededForSwipingImplementation: SelectSection
```
anna1901, line 18: Isn't this selected section the one that was used previously? I think this other selectedSection was added with this feature, wasn't it?

jankrzempek, line 18: To be honest I am not sure why you have this property that holds `SelectSection` twice. Maybe you could stay with only this one. If so you could be able to delete a few things below.

About the naming, I personally would leave `Implementation` suffix. 
Only idea, so do it what you want, but maybe: `selectedSectionDesignedForSwiping`?

 If you will stay with only one property then the naming also can change 180 degrees I suppose.

AirCasting/Dashboard/DashboardView.swift
```diff
@@ -59,9 +59,9 @@ struct DashboardView: View {
         .navigationBarTitle(Strings.DashboardView.dashboardText)
         .navigationBarHidden(true)
         .onChange(of: selectedSection, perform: { newValue in
-            self.selectedSection_.selectedSection = newValue
+            self.selectedSectionNeededForSwipingImplementation.selectedSection = newValue
         })
-        .onChange(of: selectedSection_.selectedSection , perform: { newValue in
+        .onChange(of: selectedSectionNeededForSwipingImplementation.selectedSection , perform: { newValue in
             self.selectedSection = newValue
         })
```
jankrzempek, line 66: This onChange where one property overrides the other, so that they always match can be deleted I think. Please tell me 😄

jankrzempek (commented):
I wondered and have two comments, please tell me how you see it ✌️

AirCasting/Dashboard/DashboardView.swift
SzymonGorowski: Good catch Ania, thanks :)

AirCasting/CreateSessionViews/ConfirmCreatingSessionView.swift
```diff
@@ -140,9 +140,9 @@ extension ConfirmCreatingSessionView {
                 case .success:
                     self.creatingSessionFlowContinues = false
                     if sessionContext.sessionType == .mobile {
-                        selectedSection.selectedSection = SelectedSection.mobileActive
+                        selectedSection.section = DashboardSection.mobileActive
```
jankrzempek, line 143: You can just use `.mobileActive` here as the Type is inherited

AirCasting/CreateSessionViews/ConfirmCreatingSessionView.swift
```diff
@@ -140,9 +140,9 @@ extension ConfirmCreatingSessionView {
...
                     } else {
-                        selectedSection.selectedSection = SelectedSection.following
+                        selectedSection.section = DashboardSection.following
```
jankrzempek, line 145: You can just use `.following` here as the Type is inherited

AirCasting/SessionViews/SessionCardView.swift
```diff
@@ -49,12 +49,12 @@ struct SessionCardView: View {
 
     var shouldShowValues: MeasurementPresentationStyle {
         // We need to specify selectedSection to show values for fixed session only in following tab
-        let shouldHide = isCollapsed && ( (session.isFixed && selectedSection.selectedSection == SelectedSection.fixed) || session.isDormant)
+        let shouldHide = isCollapsed && ( (session.isFixed && selectedSection.section == DashboardSection.fixed) || session.isDormant)
```
jankrzempek, line 52: same here, only `.fixed`

AirCasting/SessionViews/SessionCardView.swift
```diff
@@ -49,12 +49,12 @@ struct SessionCardView: View {
...
         return shouldHide ? .hideValues : .showValues
     }
 
     var showChart: Bool {
-        (session.isMobile && session.isActive) || (session.isFixed && selectedSection.selectedSection == SelectedSection.following)
+        (session.isMobile && session.isActive) || (session.isFixed && selectedSection.section == DashboardSection.following)
```
jankrzempek, line 57: and here 🙌🏻

jankrzempek (approved):
Comments which you can consider, but despite that - awesome cleaning and prettying 🦾

anna1901 (changes requested):
Good work!
There is one more small thing to correct: when the user swipes to fixed section the picker should scroll to the fixed tab. Right now the "Fixed" text is cut off. There is also one small code quality suggestion

anna1901 (approved):

PR #827, opened 2022-07-27
Add test for sending request to share session

anna1901:
This is a simple test for checking if we are sending a right request for sharing session via email

### DIFF
--- a/AirCastingTests/API/ShareSessionFileTest.swift
+++ b/AirCastingTests/API/ShareSessionFileTest.swift
@@ -1,0 +1,27 @@
+// Created by Lunar on 27/07/2022.
+//
+
+import XCTest
+import Resolver
+@testable import AirCasting
+
+class ShareSessionFileTest: APIServiceTestCase {
+    lazy var sut = ShareSessionApi()
+
+    func test_createsCorrectRequest() throws {
+        var receivedRequest: URLRequest?
+        client.requestTaskStub = { request, completion in
+            receivedRequest = request
+            completion(.failure(URLError(.timedOut)), request)
+        }
+        
+        let email = "test@test.com"
+        let uuid = UUID().uuidString
+        sut.sendSession(email: email, uuid: uuid, completion: { _ in })
+
+        let request = try XCTUnwrap(receivedRequest)
+        XCTAssertEqual(request.url, URL(string: "http://aircasting.org/api/sessions/export_by_uuid.json?email=\(email)&uuid=\(uuid)")!)
+        XCTAssertEqual(request.httpMethod, "GET")
+    }
+
+}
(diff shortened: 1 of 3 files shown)


AirCastingTests/API/ShareSessionFileTest.swift
```diff
@@ -0,0 +1,27 @@
...
+        
+        let email = "test@test.com"
+        let uuid = UUID().uuidString
+        sut.sendSession(email: email, uuid: uuid, completion: { _ in })
+
+        let request = try XCTUnwrap(receivedRequest)
+        XCTAssertEqual(request.url, URL(string: "http://aircasting.org/api/sessions/export_by_uuid.json?email=\(email)&uuid=\(uuid)")!)
+        XCTAssertEqual(request.httpMethod, "GET")
+    }
+
```
jankrzempek, line 26: Can be deleted

AirCastingTests/API/ShareSessionFileTest.swift
```diff
@@ -0,0 +1,27 @@
...
+            completion(.failure(URLError(.timedOut)), request)
+        }
+        
+        let email = "test@test.com"
+        let uuid = UUID().uuidString
+        sut.sendSession(email: email, uuid: uuid, completion: { _ in })
+
+        let request = try XCTUnwrap(receivedRequest)
+        XCTAssertEqual(request.url, URL(string: "http://aircasting.org/api/sessions/export_by_uuid.json?email=\(email)&uuid=\(uuid)")!)
+        XCTAssertEqual(request.httpMethod, "GET")
```
pawelgil, line 24: this test is actually testing 2 different things:
URL creation
HTTP method sent

It's very minor here, but it could be split into 2.

The reason is that if we only looked at the tests report we'd only see something like:
```
[PASSED] SomeSuit.test_that_succeeded
[PASSED] SomeSuit.other_test_that_succeeded
[FAILED] ShareSessionFileTest.createCorrectRequset
```
So we wouldn't know why. If we split this into 2:
```
[PASSED] ShareSessionFileTest.whenSendsRequest_hasCorrectURL
[FAILED] ShareSessionFileTest.whenSendsRequest_usesGETMethod
```
Which tells us exactly to the point what is malfunctioning

AirCastingTests/API/ShareSessionFileTest.swift
```diff
@@ -0,0 +1,27 @@
+// Created by Lunar on 27/07/2022.
+//
+
+import XCTest
+import Resolver
```
jankrzempek, line 5: Can be deleted

jankrzempek (approved):
Works Well!

pawelgil (approved):

PR #832, opened 2022-08-01
Dormant sessions alert

anna1901:
https://trello.com/c/YBt6lMvu

### DIFF
--- a/AirCasting/APICommunicator/DormantStreamAlertAPI.swift
+++ b/AirCasting/APICommunicator/DormantStreamAlertAPI.swift
@@ -45,13 +45,13 @@ struct DormantStreamAlertAPI {
         
         do {
             request.httpBody = try encoder.encode(params)
-            _ = try authorization.authorise(request: &request)
+            try authorization.authorise(request: &request)
         } catch {
             Log.error("Failed to prepare request")
             completion(.failure(.wrongRequest))
         }
         
-        _ = apiClient.requestTask(for: request) { [validator] result, request in
+        apiClient.requestTask(for: request) { [validator] result, request in
             do {
                 switch result {
                 case .failure(let error): completion(.failure(.failedRequest(error)))
--- a/AirCasting/APICommunicator/DormantStreamAlertAPI.swift
+++ b/AirCasting/APICommunicator/DormantStreamAlertAPI.swift
@@ -5,6 +5,7 @@ import Foundation
 import Resolver
 
 enum DormantStreamAlertAPIError: Error {
+    case wrongRequest
     case failedRequest(Error)
 }
 
@@ -12,15 +13,43 @@ struct DormantStreamAlertAPI {
     @Injected private var urlProvider: URLProvider
     @Injected private var apiClient: APIClient
     @Injected private var validator: HTTPResponseValidator
+    @Injected private var authorization: RequestAuthorisationService
+    
+    private let encoder = JSONEncoder()
+    private let decoder = JSONDecoder()
+    
+    struct Params: Encodable {
+        private let data: Nested
+        
+        struct Nested: Encodable {
+            let sessionStoppedAlert: String
+        }
+        
+        init(value: Bool) {
+            self.data = Nested(sessionStoppedAlert: String(value))
+        }
+    }
+    
+    struct Output: Decodable {
+        let action: String
+    }
     
     func sendNewSetting(value: Bool, completion: @escaping (Result<Void, DormantStreamAlertAPIError>) -> Void) {
         let url = urlProvider.baseAppURL.appendingPathComponent("api/user/settings")
         
-        let params = ["session_stopped_alert": value]
+        let params = Params(value: value)
         var request = URLRequest(url: url)
         request.httpMethod = "POST"
-        request.httpBody = try? JSONSerialization.data(withJSONObject: params, options: [])
         request.addValue("application/json", forHTTPHeaderField: "Content-Type")
+        encoder.keyEncodingStrategy = .convertToSnakeCase
+        
+        do {
+            request.httpBody = try encoder.encode(params)
+            _ = try authorization.authorise(request: &request)
+        } catch {
+            Log.error("Failed to prepare request")
+            completion(.failure(.wrongRequest))
+        }
         
         _ = apiClient.requestTask(for: request) { [validator] result, request in
             do {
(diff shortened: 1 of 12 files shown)


AirCastingTests/API/DormantStreamAlertAPITest.swift
```diff
@@ -0,0 +1,62 @@
...
+            receivedRequest = request
+            completion(.failure(URLError(.timedOut)), request)
+        }
+        
+        sut.sendNewSetting(value: false, completion: { _ in })
+        let body = try XCTUnwrap(receivedRequest?.httpBody)
+        let expectedBody = "{\"data\":{\"session_stopped_alert\":\"false\"}}"
+        
+        XCTAssertEqual(String(data: body, encoding: .utf8), expectedBody)
+    }
```
jankrzempek, line 61: I am not any expert in those Tests, so just an idea. Do we need to check body content based on the the value which we are sending? Can't we just assume that the value is `.any` and then, aim to test only `body` itself - then we would be able to stay with only one out of those two tests. What do you think?

AirCastingTests/API/DormantStreamAlertAPITest.swift
```diff
@@ -0,0 +1,62 @@
...
+}
```
jankrzempek, line 62: Awesome those tests are!!! 👌🏻
Alt Text

jankrzempek (commented):
Works well, email is here!
Let's talk about those tests and then my work is done here 🫡

AirCasting/APICommunicator/DormantStreamAlertAPI.swift
```diff
@@ -0,0 +1,70 @@
...
+        
+        let params = Params(value: value)
+        var request = URLRequest(url: url)
+        request.httpMethod = "POST"
+        request.addValue("application/json", forHTTPHeaderField: "Content-Type")
+        encoder.keyEncodingStrategy = .convertToSnakeCase
+        
+        do {
+            request.httpBody = try encoder.encode(params)
+            _ = try authorization.authorise(request: &request)
```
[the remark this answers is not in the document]
anna1901: I get a warning if I don't add this. Do you know if there is a way around this?

AirCasting/APICommunicator/DormantStreamAlertAPI.swift
```diff
@@ -0,0 +1,70 @@
...
+            let sessionStoppedAlert: String
+        }
+        
+        init(value: Bool) {
+            self.data = Nested(sessionStoppedAlert: String(value))
+        }
+    }
+    
+    struct Output: Decodable {
+        let action: String
```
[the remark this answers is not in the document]
anna1901: "action" is what we get from backend so I'm not sure it makes sense to change this

AirCastingTests/API/DormantStreamAlertAPITest.swift
```diff
@@ -0,0 +1,62 @@
...
+            receivedRequest = request
+            completion(.failure(URLError(.timedOut)), request)
+        }
+        
+        sut.sendNewSetting(value: true, completion: { _ in })
+        let request = try XCTUnwrap(receivedRequest)
+        XCTAssertEqual(request.httpMethod, "POST")
+    }
+
+    func test_whenSendsRequestwithTrueValue_hasCorrectBody() throws {
```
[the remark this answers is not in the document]
anna1901: Where? Could you elaborate what it should look like and why?

AirCastingTests/API/DormantStreamAlertAPITest.swift
anna1901: But the body itself changes, so I'm not sure how could we change this to use `.any`? Could you elaborate?

AirCasting/APICommunicator/DormantStreamAlertAPI.swift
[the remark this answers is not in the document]
jankrzempek: I said it to you, becouse I was not getting any warning when deleting `_ =`. Please skip then, I think it would be a waste of time to think about this

AirCastingTests/API/DormantStreamAlertAPITest.swift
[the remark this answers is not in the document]
jankrzempek: naming here should be `test_whenSendsRequestWithTrueValue_hasCorrectBody` instead of `test_whenSendsRequestwithTrueValue_hasCorrectBody`

AirCastingTests/API/DormantStreamAlertAPITest.swift
jankrzempek: I can see now that I took the wrong way of thinking yesterday (thinking that maybe we could test what was not changed inside the body, but no, I think it better not to go this way) - sorry for the confusion.

jankrzempek (approved):
👌🏻

AirCasting/APICommunicator/DormantStreamAlertAPI.swift
```diff
@@ -0,0 +1,70 @@
...
+
+import Foundation
+import Resolver
+
+enum DormantStreamAlertAPIError: Error {
+    case wrongRequest
+    case failedRequest(Error)
+}
+
+struct DormantStreamAlertAPI {
```
pawelgil, line 12: Would be great to add protocol for this as it's a boundary object

pawelgil (changes requested):
Nice! Those tests are PRO! Let's add some more and protocolize the API layer

AirCastingTests/API/DormantStreamAlertAPITest.swift
```diff
@@ -5,58 +5,80 @@ import XCTest
...
+        
+        XCTAssertEqual(1, completionCheck)
+    }
+    
+    func test_whenSendsRequestWithSuccess_callsCompletionWithSuccess() throws {
+        client.requestTaskStub = { request, completion in
+            completion(.success((data: Data(#"{ "action": "success" }"#.utf8), response: HTTPURLResponse())), request)
+        }
+        
+        var completionCheck = 0
```
pawelgil, line 62: same

pawelgil (approved):
💯

anna1901:
[settings: dormant stream alert](https://trello.com/c/YBt6lMvu/749-settings-dormant-stream-alert)

PR #834, opened 2022-08-03
ExploreSession button bug - UITests

jankrzempek:
Covered explore session button flow in which SF screen could appear always on + button click.

To check please move `exploreSessionsButton.exploreSessionsButtonTapped = false` to viewInitialized.
Zrzut ekranu 2022-08-3 o 14 34 15


pawelgil (approved):
You folks are on 🔥 with the testing lately! So awesome! 💯

PR #835, opened 2022-08-03
Search and follow - UnitTests

jankrzempek:
https://trello.com/c/lAnLeTxO/666-write-tests-for-search-and-follow-feature

### DIFF
--- a/AirCasting/SearchAndFollow/SearchLocation/SearchViewModel.swift
+++ b/AirCasting/SearchAndFollow/SearchLocation/SearchViewModel.swift
@@ -40,7 +40,7 @@ class SearchViewModel: ObservableObject {
     func textFieldTapped() { isLocationPopupPresented.toggle() }
     
     func onParameterTap(with param: String) {
-        self.measurementTypes.forEach({ $0.isSelected = false })
+        deselectAll(measurementTypes)
         self.measurementTypes.first(where: { $0.name == param })?.isSelected = true
         switch measurementType(from: param) {
         case .particulateMatter: self.sensorTypes = [
@@ -56,9 +56,13 @@ class SearchViewModel: ObservableObject {
     }
     
     func onSensorTap(with name: String) {
-        self.sensorTypes.forEach({ $0.isSelected = false })
+        deselectAll(sensorTypes)
         self.sensorTypes.first(where: { $0.name == name })?.isSelected = true
         self.objectWillChange.send()
+    }
+    
+    func deselectAll(_ all: [SearchParameter]) {
+        all.forEach({ $0.isSelected = false })
     }
     
     func locationNameInteracted(with newLocationName: String) {
--- a/AirCasting/SearchAndFollow/SearchLocation/SearchViewModel.swift
+++ b/AirCasting/SearchAndFollow/SearchLocation/SearchViewModel.swift
@@ -61,10 +61,6 @@ class SearchViewModel: ObservableObject {
         self.objectWillChange.send()
     }
     
-    func deselectAll(_ all: [SearchParameter]) {
-        all.forEach({ $0.isSelected = false })
-    }
-    
     func locationNameInteracted(with newLocationName: String) {
         addressName = newLocationName
     }
@@ -79,6 +75,10 @@ class SearchViewModel: ObservableObject {
                 onEnd()
             }))
         }
+    }
+    
+    private func deselectAll(_ all: [SearchParameter]) {
+        all.forEach({ $0.isSelected = false })
     }
     
     private func measurementType(from: String) -> MeasurementType? {
--- a/AirCastingTests/SearchAndFollowTests/AirBeamMeasurementDownloaderTests.swift
+++ b/AirCastingTests/SearchAndFollowTests/AirBeamMeasurementDownloaderTests.swift
@@ -17,40 +17,45 @@ class AirBeamMeasurementDownloaderTests: ACTestCase {
     func test_requestingDataFromServer_usesCorrectPath() throws {
         sut.downloadStreams(with: Int.default, completion: { _ in })
         
+        let firstCallHistoryElement = try XCTUnwrap(clientSpy.callHistory.first, "After sut execution there should be first element in call history.")
         XCTAssert(clientSpy.callHistory.count == 1, "Number of requests should be equal to one, as only once call was requested.")
-        let url = try XCTUnwrap(URLComponents(string: clientSpy.callHistory.first!.description)?.url, "Problem when trying to get url.")
+        let url = try XCTUnwrap(URLComponents(string: firstCallHistoryElement.description)?.url, "Problem when trying to get url.")
         XCTAssertEqual(url.path, "/api/fixed/sessions/\(Int.default)/streams.json")
     }
     
-    func test_requestingDataFromServer_usesGetMethod() {
+    func test_requestingDataFromServer_usesGetMethod() throws {
         sut.downloadStreams(with: Int.default, completion: { _ in })
         
+        let firstCallHistoryElement = try XCTUnwrap(clientSpy.callHistory.first, "After sut execution there should be first element in call history.")
         XCTAssert(clientSpy.callHistory.count == 1, "Number of requests should be equal to one, as only once call was requested.")
-        XCTAssertEqual(clientSpy.callHistory.first!.httpMethod, "GET")
+        XCTAssertEqual(firstCallHistoryElement.httpMethod, "GET")
     }
     
     func test_requestingDataFromServer_usesGivenSesionUUID() throws {
         let sessionUUID = 123456-789
         sut.downloadStreams(with: sessionUUID, completion: { _ in })
         
+        let firstCallHistoryElement = try XCTUnwrap(clientSpy.callHistory.first, "After sut execution there should be first element in call history.")
         XCTAssert(clientSpy.callHistory.count == 1, "Number of requests should be equal to one, as only once call was requested.")
-        let url = try XCTUnwrap(URLComponents(string: clientSpy.callHistory.first!.description)?.url, "Problem when trying to get url.")
+        let url = try XCTUnwrap(URLComponents(string: firstCallHistoryElement.description)?.url, "Problem when trying to get url.")
         XCTAssertEqual(url.path, "/api/fixed/sessions/\(sessionUUID)/streams.json")
     }
     
     func test_requestingDataFromServer_usesCorrectNameForLimit() throws {
         sut.downloadStreams(with: Int.default, completion: { _ in })
         
+        let firstCallHistoryElement = try XCTUnwrap(clientSpy.callHistory.first, "After sut execution there should be first element in call history.")
         XCTAssert(clientSpy.callHistory.count == 1, "Number of requests should be equal to one, as only once call was requested.")
-        let queryItem = try XCTUnwrap(URLComponents(string: clientSpy.callHistory.first!.description)?.queryItems, "Problem when trying to get url.")
+        let queryItem = try XCTUnwrap(URLComponents(string: firstCallHistoryElement.description)?.queryItems, "Problem when trying to get url.")
         XCTAssert(queryItem.contains(where: { $0.name == "measurements_limit" }))
     }
     
     func test_requestingDataFromServer_givesCorrectMeasurementNumber() throws {
         sut.downloadStreams(with: Int.default, completion: { _ in })
         
+        let firstCallHistoryElement = try XCTUnwrap(clientSpy.callHistory.first, "After sut execution there should be first element in call history.")
         XCTAssert(clientSpy.callHistory.count == 1, "Number of requests should be equal to one, as only once call was requested.")
-        let queryItem = try XCTUnwrap(URLComponents(string: clientSpy.callHistory.first!.description)?.queryItems, "Problem when trying to get url.")
+        let queryItem = try XCTUnwrap(URLComponents(string: firstCallHistoryElement.description)?.queryItems, "Problem when trying to get url.")
         let measurementsValue = try XCTUnwrap(queryItem.first(where: { $0.name == "measurements_limit" }))
         XCTAssertEqual(measurementsValue.value, "1440")
     }
--- a/AirCastingTests/SearchAndFollowTests/AirBeamMeasurementDownloaderTests.swift
+++ b/AirCastingTests/SearchAndFollowTests/AirBeamMeasurementDownloaderTests.swift
@@ -1,0 +1,49 @@
+// Created by Lunar on 03/08/2022.
+//
+import XCTest
+import Resolver
+@testable import AirCasting
+
+class AirBeamMeasurementDownloaderTests: ACTestCase {
+    lazy var sut = AirBeamMeasurementsDownloaderDefault()
+    let clientSpy = APIClientSpy()
+    
+    func test_whenRequestingDataFromServer_shouldUseAssertedPath() {
+        Resolver.test.register { self.clientSpy as APIClient }
+        sut.downloadStreams(with: Int.default, completion: { _ in })
+        guard let url = URLComponents(string: clientSpy.request.description)?.url else { XCTFail("Problem when trying to get url."); return }
+        XCTAssertEqual(url.path, "/api/fixed/sessions/\(Int.default)/streams.json")
+    }
+    
+    func test_whenRequestingDataFromServer_shouldUseGetMethod() {
+        Resolver.test.register { self.clientSpy as APIClient }
+        sut.downloadStreams(with: Int.default, completion: { _ in })
+        XCTAssertEqual(clientSpy.request.httpMethod, "GET")
+    }
+    
+    func test_whenRequestingDataFromServer_shouldUseGivenSesionUUIDinsidePath() {
+        Resolver.test.register { self.clientSpy as APIClient }
+        let sessionUUID = 123456-789
+        sut.downloadStreams(with: sessionUUID, completion: { _ in })
+        guard let url = URLComponents(string: clientSpy.request.description)?.url else { XCTFail("Problem when trying to get url."); return }
+        XCTAssertEqual(url.path, "/api/fixed/sessions/\(sessionUUID)/streams.json")
+    }
+    
+    class APIClientSpy: APIClient {
+        
+        struct URLSessionCancellable: Cancellable {
+            weak var dataTask: URLSessionDataTask?
+            
+            func cancel() {
+                dataTask?.cancel()
+            }
+        }
+        
+        var request: URLRequest!
+        
+        func requestTask(for request: URLRequest, completion: @escaping (Result<(data: Data, response: HTTPURLResponse), Error>, URLRequest) -> Void) -> Cancellable {
+            self.request = request
+            return URLSessionCancellable()
+        }
+    }
+}
--- a/AirCastingTests/SearchAndFollowTests/AirBeamMeasurementDownloaderTests.swift
+++ b/AirCastingTests/SearchAndFollowTests/AirBeamMeasurementDownloaderTests.swift
@@ -6,44 +6,52 @@ import Resolver
 
 class AirBeamMeasurementDownloaderTests: ACTestCase {
     lazy var sut = AirBeamMeasurementsDownloaderDefault()
-    let clientSpy = APIClientSpy()
+    let clientSpy = APIClientMock()
     
-    func test_whenRequestingDataFromServer_shouldUseAssertedPath() {
+    override func setUp() {
+        super.setUp()
         Resolver.test.register { self.clientSpy as APIClient }
+        clientSpy.returning((data: Data.default, response: .success()))
+    }
+    
+    func test_requestingDataFromServer_usesCorrectPath() throws {
         sut.downloadStreams(with: Int.default, completion: { _ in })
-        guard let url = URLComponents(string: clientSpy.request.description)?.url else { XCTFail("Problem when trying to get url."); return }
+        
+        XCTAssert(clientSpy.callHistory.count == 1, "Number of requests should be equal to one, as only once call was requested.")
+        let url = try XCTUnwrap(URLComponents(string: clientSpy.callHistory.first!.description)?.url, "Problem when trying to get url.")
         XCTAssertEqual(url.path, "/api/fixed/sessions/\(Int.default)/streams.json")
     }
     
-    func test_whenRequestingDataFromServer_shouldUseGetMethod() {
-        Resolver.test.register { self.clientSpy as APIClient }
+    func test_requestingDataFromServer_usesGetMethod() {
         sut.downloadStreams(with: Int.default, completion: { _ in })
-        XCTAssertEqual(clientSpy.request.httpMethod, "GET")
+        
+        XCTAssert(clientSpy.callHistory.count == 1, "Number of requests should be equal to one, as only once call was requested.")
+        XCTAssertEqual(clientSpy.callHistory.first!.httpMethod, "GET")
     }
     
-    func test_whenRequestingDataFromServer_shouldUseGivenSesionUUIDinsidePath() {
-        Resolver.test.register { self.clientSpy as APIClient }
+    func test_requestingDataFromServer_usesGivenSesionUUID() throws {
         let sessionUUID = 123456-789
         sut.downloadStreams(with: sessionUUID, completion: { _ in })
-        guard let url = URLComponents(string: clientSpy.request.description)?.url else { XCTFail("Problem when trying to get url."); return }
+        
+        XCTAssert(clientSpy.callHistory.count == 1, "Number of requests should be equal to one, as only once call was requested.")
+        let url = try XCTUnwrap(URLComponents(string: clientSpy.callHistory.first!.description)?.url, "Problem when trying to get url.")
         XCTAssertEqual(url.path, "/api/fixed/sessions/\(sessionUUID)/streams.json")
     }
     
-    class APIClientSpy: APIClient {
+    func test_requestingDataFromServer_usesCorrectNameForLimit() throws {
+        sut.downloadStreams(with: Int.default, completion: { _ in })
         
-        struct URLSessionCancellable: Cancellable {
-            weak var dataTask: URLSessionDataTask?
-            
-            func cancel() {
-                dataTask?.cancel()
-            }
-        }
+        XCTAssert(clientSpy.callHistory.count == 1, "Number of requests should be equal to one, as only once call was requested.")
+        let queryItem = try XCTUnwrap(URLComponents(string: clientSpy.callHistory.first!.description)?.queryItems, "Problem when trying to get url.")
+        XCTAssert(queryItem.contains(where: { $0.name == "measurements_limit" }))
+    }
+    
+    func test_requestingDataFromServer_givesCorrectMeasurementNumber() throws {
+        sut.downloadStreams(with: Int.default, completion: { _ in })
         
-        var request: URLRequest!
-        
-        func requestTask(for request: URLRequest, completion: @escaping (Result<(data: Data, response: HTTPURLResponse), Error>, URLRequest) -> Void) -> Cancellable {
-            self.request = request
-            return URLSessionCancellable()
-        }
+        XCTAssert(clientSpy.callHistory.count == 1, "Number of requests should be equal to one, as only once call was requested.")
+        let queryItem = try XCTUnwrap(URLComponents(string: clientSpy.callHistory.first!.description)?.queryItems, "Problem when trying to get url.")
+        let measurementsValue = try XCTUnwrap(queryItem.first(where: { $0.name == "measurements_limit" }))
+        XCTAssertEqual(measurementsValue.value, "1440")
     }
 }
--- a/AirCastingTests/SearchAndFollowTests/MapDownloaderMeasurementTypeTests.swift
+++ b/AirCastingTests/SearchAndFollowTests/MapDownloaderMeasurementTypeTests.swift
@@ -5,7 +5,7 @@ import XCTest
 @testable import AirCasting
 
 class MapDownloaderMeasurementTypeTests: XCTestCase {
-    func testGettingApiName_whenAsked_shouldReturnBelowString() {
+    func test_gettingApiName_returnsCorrectString() {
         let particulateMatter: String = MapDownloaderMeasurementType.particulateMatter.apiName
         XCTAssertEqual(particulateMatter, "Particulate Matter")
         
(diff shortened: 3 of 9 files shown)


AirCastingTests/SearchAndFollowTests/AirBeamMeasurementDownloaderTests.swift
```diff
@@ -0,0 +1,49 @@
...
+        
+        struct URLSessionCancellable: Cancellable {
+            weak var dataTask: URLSessionDataTask?
+            
+            func cancel() {
+                dataTask?.cancel()
+            }
+        }
+        
+        var request: URLRequest!
```
pawelgil, line 42: 1. This should be an optional because what happens if our `sut` doesn't call `requestTask` at all? (tests would crash). 
2. I also like more verbose naming in tests. Maybe `capturedRequest`? It will make test-site much more readable
3. What if the `sut` calls `requestTask` multiple times by mistake? We won't get a failing test. This is why a good practice in `Spy` objects is to make capturing variables an arrays. Like that: `var capturedRequests: [URLRequest]`. If you do so you won't need to make it an optional too.

AirCastingTests/SearchAndFollowTests/AirBeamMeasurementDownloaderTests.swift
```diff
@@ -0,0 +1,49 @@
...
+    func test_whenRequestingDataFromServer_shouldUseAssertedPath() {
+        Resolver.test.register { self.clientSpy as APIClient }
+        sut.downloadStreams(with: Int.default, completion: { _ in })
+        guard let url = URLComponents(string: clientSpy.request.description)?.url else { XCTFail("Problem when trying to get url."); return }
+        XCTAssertEqual(url.path, "/api/fixed/sessions/\(Int.default)/streams.json")
+    }
+    
+    func test_whenRequestingDataFromServer_shouldUseGetMethod() {
+        Resolver.test.register { self.clientSpy as APIClient }
+        sut.downloadStreams(with: Int.default, completion: { _ in })
```
pawelgil, line 20: See - here using `default` is perfect because streamID is not participating in what we're testing, so we're telling the reader not to worry about the value.

AirCastingTests/SearchAndFollowTests/MapDownloaderMeasurementTypeTests.swift
```diff
@@ -0,0 +1,31 @@
+// Created by Lunar on 03/08/2022.
+//
+
+import XCTest
+@testable import AirCasting
+
+class MapDownloaderMeasurementTypeTests: XCTestCase {
+    func testGettingApiName_whenAsked_shouldReturnBelowString() {
```
pawelgil, line 8: Below -> Correct

AirCastingTests/SearchAndFollowTests/MapDownloaderMeasurementTypeTests.swift
```diff
@@ -0,0 +1,31 @@
...
+        let particulateMatter: String = MapDownloaderMeasurementType.particulateMatter.apiName
+        XCTAssertEqual(particulateMatter, "Particulate Matter")
+        
+        let ozone: String = MapDownloaderMeasurementType.ozone.apiName
+        XCTAssertEqual(ozone, "Ozone")
+    }
+    
+    func testGettingSensorSuffixName_whenAsked_shouldReturnBelowString() {
```
pawelgil, line 16: Below -> Correct

AirCastingTests/SearchAndFollowTests/MapDownloaderMeasurementTypeTests.swift
```diff
@@ -0,0 +1,31 @@
...
+        let particulateMatter: String = MapDownloaderMeasurementType.particulateMatter.sensorNameSuffix
+        XCTAssertEqual(particulateMatter, "-pm2.5")
+        
+        let ozone: String = MapDownloaderMeasurementType.ozone.sensorNameSuffix
+        XCTAssertEqual(ozone, "-o3")
+    }
+    
+    func testGettingCapitalizedName_whenAsked_shouldReturnBelowString() {
```
pawelgil, line 24: Below -> Correct

AirCastingTests/SearchAndFollowTests/MapDownloaderSensorTypeTests.swift
```diff
@@ -0,0 +1,29 @@
+// Created by Lunar on 03/08/2022.
+//
+
+import XCTest
+@testable import AirCasting
+
+class MapDownloaderSensorTypeTests: XCTestCase {
+    func testGettingSensorNamePrefix_whenAsked_shouldReturnBelowString() {
```
pawelgil, line 8: Below -> Correct

AirCastingTests/SearchAndFollowTests/MapDownloaderSensorTypeTests.swift
```diff
@@ -0,0 +1,29 @@
...
+        XCTAssertEqual(airbeam, "airbeam")
+        
+        let openAQ: String = MapDownloaderSensorType.OpenAQ.sensorNamePrefix
+        XCTAssertEqual(openAQ, "openaq")
+        
+        let purpleAir: String = MapDownloaderSensorType.PurpleAir.sensorNamePrefix
+        XCTAssertEqual(purpleAir, "purpleair")
+    }
+    
+    func testGettingCapitalizedName_whenAsked_shouldReturnBelowString() {
```
pawelgil, line 19: Below -> Correct

AirCastingTests/SearchAndFollowTests/MapDownloaderUnitSymbolTests.swift
```diff
@@ -0,0 +1,15 @@
+// Created by Lunar on 03/08/2022.
+//
+
+import XCTest
+@testable import AirCasting
+
+class MapDownloaderUnitSymbolTests: XCTestCase {
+    func testGettingName_whenAsked_shouldReturnBelowString() {
```
pawelgil, line 8: Below -> Correct

AirCastingTests/SearchAndFollowTests/SearchViewModelTests.swift
```diff
@@ -0,0 +1,56 @@
...
+import XCTest
+@testable import AirCasting
+
+class SearchViewModelTests: XCTestCase {
+    var sut: SearchViewModel!
+    
+    override func setUp() {
+        super.setUp()
+        sut = .init()
+    }
```
pawelgil, line 12: Not needed, you can just do `let sut = SearchViewModel()` in this case I think

AirCastingTests/SearchAndFollowTests/SearchViewModelTests.swift
```diff
@@ -0,0 +1,56 @@
...
+        
+        sut.onSensorTap(with: tappingSensorName)
+        
+        let selected = sut.sensorTypes.filter({ $0.isSelected })
+        
+        XCTAssertEqual(selected.count, 1)
+        XCTAssertEqual(selected.first!.name, tappingSensorName)
+    }
+    
+    func test_whenAskedToDeselectAll_shouldIsSelectedBeFalse() {
```
pawelgil, line 35: `shouldIsSelectedBeFalse` doesn't make much sense to me 😅

AirCastingTests/SearchAndFollowTests/SearchViewModelTests.swift
```diff
@@ -0,0 +1,56 @@
...
+        sut.sensorTypes = [.init(isSelected: true, name: "1"),
+                           .init(isSelected: false, name: "2"),
+                           .init(isSelected: true, name: "3")]
+        
+        sut.deselectAll(sut.sensorTypes)
+        let unSelected = sut.sensorTypes.filter({ !$0.isSelected })
+        XCTAssertEqual(unSelected.count, sut.sensorTypes.count)
```
pawelgil, line 42: why not just like before - get all selected and make sure the count is 0? It feels a bit backwards here

AirCastingTests/SearchAndFollowTests/SearchViewModelTests.swift
```diff
@@ -0,0 +1,56 @@
...
+    }
+    
+    func test_particulateMatterTap_correspondingSensorsShouldAppear() {
+        let particulateMatter = "Particulate Matter"
+        sut.onParameterTap(with: particulateMatter)
+        XCTAssertEqual(sut.sensorTypes.count, 3)
```
pawelgil, line 48: instead of counting sensor types can we do some more meaningful assertion here? Maybe do `sut.sensorTypes.map(\.name)`?

AirCastingTests/SearchAndFollowTests/SearchViewModelTests.swift
```diff
@@ -0,0 +1,56 @@
...
+    }
+    
+    func test_ozoneTap_correspondingSensorsShouldAppear() {
+        let particulateMatter = "Ozone"
+        sut.onParameterTap(with: particulateMatter)
+        XCTAssertEqual(sut.sensorTypes.count, 1)
```
pawelgil, line 54: here too - maybe compare name instead of count?

AirCastingTests/SearchAndFollowTests/SessionForLocationDownloaderTests.swift
```diff
@@ -0,0 +1,87 @@
...
+    }
+    
+    struct MockedSession {
+        let geoSquare: GeoSquare
+        let timeFrom: Double
+        let timeTo: Double
+        let measurementType: MapDownloaderMeasurementType
+        let sensor: MapDownloaderSensorType
+        
+        static func getAnySession(using sut: SessionsForLocationDownloader) {
```
pawelgil, line 69: passing `sut` here is not needed, just do `var any: Self { ... }`

AirCastingTests/SearchAndFollowTests/SearchViewModelTests.swift
```diff
@@ -0,0 +1,56 @@
...
+        sut.sensorTypes = [.init(isSelected: true, name: "1"),
+                           .init(isSelected: false, name: "2"),
+                           .init(isSelected: true, name: "3")]
+        
+        sut.deselectAll(sut.sensorTypes)
+        let unSelected = sut.sensorTypes.filter({ !$0.isSelected })
+        XCTAssertEqual(unSelected.count, sut.sensorTypes.count)
+    }
+    
+    func test_particulateMatterTap_correspondingSensorsShouldAppear() {
```
pawelgil, line 45: a small tip on naming tests: using `should` in the **then** part of the name tends to repeat in each case. It's nothing bad, but I found that switching from conditional clause to just simple present tense make test names better IMO: `test_particulareMatterTapped_correspondingSensorAppears`. It switches test names to be **describing** the system instead of asking

AirCastingTests/SearchAndFollowTests/SessionForLocationDownloaderTests.swift
```diff
@@ -0,0 +1,87 @@
...
+    lazy var sut = SessionsForLocationDownloaderDefault()
+    let clientSpy = APIClientSpy()
+    
+    func test_whenRequestingDataFromServer_shouldUseGetMethod() {
+        Resolver.test.register { self.clientSpy as APIClient }
+        MockedSession.getAnySession(using: sut)
+        XCTAssertEqual(clientSpy.request.httpMethod, "GET")
+    }
+    
+    func test_whenCreatingQueryItems_shouldUseSmallQAsKey() {
```
pawelgil, line 17: Key -> ParameterName

AirCastingTests/SearchAndFollowTests/SessionForLocationDownloaderTests.swift
```diff
@@ -0,0 +1,87 @@
...
+        Resolver.test.register { self.clientSpy as APIClient }
+        MockedSession.getAnySession(using: sut)
+        guard let queryItems = URLComponents(string: clientSpy.request.description)?.queryItems else { XCTFail("Problem when trying to get query items from url."); return }
+        XCTAssertEqual(queryItems[0].name, "q")
```
pawelgil, line 21: no guarantee that there will be a `[0]` element and when not, it will crash. Use an optional `first` with combination with `XCTUnwrap`

AirCastingTests/SearchAndFollowTests/SessionForLocationDownloaderTests.swift
```diff
@@ -0,0 +1,87 @@
...
+    }
+    
+    func test_whenCreatingUrl_shouldUseAssertedPath() {
```
pawelgil, line 24: Asserted -> Correct

AirCastingTests/SearchAndFollowTests/SessionForLocationDownloaderTests.swift
```diff
@@ -0,0 +1,87 @@
...
+        Resolver.test.register { self.clientSpy as APIClient }
+        MockedSession.getAnySession(using: sut)
+        guard let url = URLComponents(string: clientSpy.request.description)?.url else { XCTFail("Problem when trying to get url request."); return }
+        XCTAssertEqual(url.path, "/api/fixed/active/sessions.json")
+    }
+    
+    // This test is considered as valid one, having in mind that [MapDownloaderUnitSymbol] is tested alongside.
+    func test_whenGettingProperUnitSymbolForParticulateMatter_shouldReturn_uqm3() {
```
pawelgil, line 32: you're getting it 💯

AirCastingTests/SearchAndFollowTests/SessionForLocationDownloaderTests.swift
```diff
@@ -0,0 +1,87 @@
...
+    }
+    
+    func test_whenCreatingUrl_shouldUseAssertedPath() {
+        Resolver.test.register { self.clientSpy as APIClient }
+        MockedSession.getAnySession(using: sut)
+        guard let url = URLComponents(string: clientSpy.request.description)?.url else { XCTFail("Problem when trying to get url request."); return }
+        XCTAssertEqual(url.path, "/api/fixed/active/sessions.json")
+    }
+    
+    // This test is considered as valid one, having in mind that [MapDownloaderUnitSymbol] is tested alongside.
```
pawelgil, line 31: alongside -> "in a separate unit test"

AirCastingTests/SearchAndFollowTests/SessionForLocationDownloaderTests.swift
```diff
@@ -0,0 +1,87 @@
...
+    func test_whenGettingProperUnitSymbolForParticulateMatter_shouldReturn_uqm3() {
+        Resolver.test.register { self.clientSpy as APIClient }
+        let symbol = sut.getProperUnitSymbol(using: .particulateMatter)
+        XCTAssertEqual(symbol, .uqm3)
+    }
+    
+    // This test is considered as valid one, having in mind that [MapDownloaderUnitSymbol] is tested alongside.
```
pawelgil, line 38: alongside -> "in a separate unit test"

AirCastingTests/SearchAndFollowTests/SessionForLocationDownloaderTests.swift
```diff
@@ -0,0 +1,87 @@
...
+    func test_whenGettingProperUnitSymbolForOzone_shouldReturn_ppb() {
+        Resolver.test.register { self.clientSpy as APIClient }
+        let symbol = sut.getProperUnitSymbol(using: .ozone)
+        XCTAssertEqual(symbol, .ppb)
+    }
+    
+    class APIClientSpy: APIClient {
```
pawelgil, line 45: same as with the other one

pawelgil (changes requested):
Wow, those tests are GOOD!

AirCastingTests/SearchAndFollowTests/AirBeamMeasurementDownloaderTests.swift
jankrzempek: Thanks so much for this step-by-step explanation - really helps a lot 💙

AirCasting/SearchAndFollow/SearchLocation/SearchViewModel.swift
```diff
@@ -56,11 +56,15 @@ class SearchViewModel: ObservableObject {
     }
     
     func onSensorTap(with name: String) {
-        self.sensorTypes.forEach({ $0.isSelected = false })
+        deselectAll(sensorTypes)
         self.sensorTypes.first(where: { $0.name == name })?.isSelected = true
         self.objectWillChange.send()
     }
     
+    func deselectAll(_ all: [SearchParameter]) {
```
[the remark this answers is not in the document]
jankrzempek: Zrzut ekranu 2022-08-8 o 10 47 38
I think I cannot, is there a way to use/test private functions?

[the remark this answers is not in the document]
jankrzempek: I can see that there is a way, but it could be a bit overkill in this case I suppose. Should I consider not testing this part of the code and moving it into private? or testing and applying something like this solution from the link? https://www.avanderlee.com/swift/testing-private-methods-variables/

AirCastingTests/SearchAndFollowTests/AirBeamMeasurementDownloaderTests.swift
```diff
@@ -0,0 +1,49 @@
...
+import XCTest
+import Resolver
+@testable import AirCasting
+
+class AirBeamMeasurementDownloaderTests: ACTestCase {
+    lazy var sut = AirBeamMeasurementsDownloaderDefault()
+    let clientSpy = APIClientSpy()
+    
+    func test_whenRequestingDataFromServer_shouldUseAssertedPath() {
+        Resolver.test.register { self.clientSpy as APIClient }
```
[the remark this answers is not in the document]
jankrzempek: 🫡

AirCastingTests/SearchAndFollowTests/AirBeamMeasurementDownloaderTests.swift
```diff
@@ -0,0 +1,49 @@
...
+//
+import XCTest
+import Resolver
+@testable import AirCasting
+
+class AirBeamMeasurementDownloaderTests: ACTestCase {
+    lazy var sut = AirBeamMeasurementsDownloaderDefault()
+    let clientSpy = APIClientSpy()
+    
+    func test_whenRequestingDataFromServer_shouldUseAssertedPath() {
```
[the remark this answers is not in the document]
jankrzempek: 🙌🏻

AirCasting/SearchAndFollow/SearchLocation/SearchViewModel.swift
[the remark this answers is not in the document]
anna1901: I don't have a full context, but can't you just check if this deselection happens when you call `onParameterTap` and `onSensorTap` since those are public methods that are supposed to do the deselection and keep this deselect func private?

AirCastingTests/SearchAndFollowTests/AirBeamMeasurementDownloaderTests.swift
```diff
@@ -0,0 +1,49 @@
+// Created by Lunar on 03/08/2022.
+//
+import XCTest
+import Resolver
+@testable import AirCasting
+
+class AirBeamMeasurementDownloaderTests: ACTestCase {
```
anna1901, line 7: I think it would be great to add some more tests here:
1. checking if completion is called correctly
2. checking for query items (1440 measurements limit)
3. perhaps we can mock a response from server and check decoding (if snake case is decoded correctly etc.)

AirCastingTests/SearchAndFollowTests/MapDownloaderMeasurementTypeTests.swift
anna1901: what about a name `test_whenGettingApiName_returnsCorrectString`?

AirCastingTests/SearchAndFollowTests/SearchViewModelTests.swift
```diff
@@ -0,0 +1,56 @@
...
+                           .init(isSelected: false, name: "3")]
+        
+        let tappingSensorName = "1"
+        
+        sut.onSensorTap(with: tappingSensorName)
+        
+        let selected = sut.sensorTypes.filter({ $0.isSelected })
+        
+        XCTAssertEqual(selected.count, 1)
+        XCTAssertEqual(selected.first!.name, tappingSensorName)
```
anna1901, line 32: I'm not sure if tests go forward after one assertion fails, but if yes, then the test will crash if selected.count is 0

AirCastingTests/SearchAndFollowTests/SearchViewModelTests.swift
anna1901, line 35: Okay, I have more context now. I think you are already checking the deselecting behavior in the test above (when one sensor get's selected, all of the others sensors get deselected). I think it's not necessary to test this method separately, as what matters is the final behavior.

AirCastingTests/SearchAndFollowTests/SessionForLocationDownloaderTests.swift
```diff
@@ -0,0 +1,87 @@
...
+import Resolver
+@testable import AirCasting
+
+class SessionForLocationDownloaderTests: ACTestCase {
+    lazy var sut = SessionsForLocationDownloaderDefault()
+    let clientSpy = APIClientSpy()
+    
+    func test_whenRequestingDataFromServer_shouldUseGetMethod() {
+        Resolver.test.register { self.clientSpy as APIClient }
+        MockedSession.getAnySession(using: sut)
```
anna1901, line 13: Hm, I'm not sure about making it a func of MockedSession. Maybe it can be changed to a function `requestSessions()` that would be a method of SessionForLocationDownloaderTests and it would call `sut.requestSessions(....)` with some mocked data? I'm not sure about this MockSessions struct, as it is not needed for anything else that calling this function and it's storing geoSquare as well, which is no a part of a sessions.

AirCastingTests/SearchAndFollowTests/SessionForLocationDownloaderTests.swift
```diff
@@ -0,0 +1,87 @@
+// Created by Lunar on 02/08/2022.
+//
+import XCTest
+import Resolver
+@testable import AirCasting
+
+class SessionForLocationDownloaderTests: ACTestCase {
```
anna1901, line 7: I think it would be good to test the query string as well

anna1901 (approved):
Awesome! 🙌🏻 I added some comments for you to consider

AirCastingTests/SearchAndFollowTests/MapDownloaderUnitSymbolTests.swift
jankrzempek: 😅

AirCastingTests/SearchAndFollowTests/SearchViewModelTests.swift
jankrzempek: I just did every test using new ideas every time 😅

AirCastingTests/SearchAndFollowTests/SearchViewModelTests.swift
jankrzempek: I tested and as I can see using breakpoints, it goes forward but the fail message which I got was connected to the line that ensures there is one element inside. I am going to investigate a bit and make myself sure about its correctness

AirCastingTests/SearchAndFollowTests/AirBeamMeasurementDownloaderTests.swift
jankrzempek: I am going to add one, but as this PR got enough big and I am near holiday time - I going to push it more onto merge state then new thing state. With all the pleasure, I can add more tests in the future 🫡

AirCastingTests/SearchAndFollowTests/SearchViewModelTests.swift
jankrzempek: I see your point and don't have hard opinion here, so I will delete

AirCastingTests/SearchAndFollowTests/SessionForLocationDownloaderTests.swift
jankrzempek: Agreed, but same as above. I will not focus on extending features, but rather on improvements.

AirCastingTests/SearchAndFollowTests/SearchViewModelTests.swift
pawelgil: Ania is right, this code will crash, failures don't stop execution. You can use `XCTUnwrap` or just do optional unwrap in possible:
```Swift
XCTAssertEqual(selected.filrst?.name, tappingSensorName)
```

AirCastingTests/SearchAndFollowTests/AirBeamMeasurementDownloaderTests.swift
```diff
@@ -0,0 +1,57 @@
...
+        super.setUp()
+        Resolver.test.register { self.clientSpy as APIClient }
+        clientSpy.returning((data: Data.default, response: .success()))
+    }
+    
+    func test_requestingDataFromServer_usesCorrectPath() throws {
+        sut.downloadStreams(with: Int.default, completion: { _ in })
+        
+        XCTAssert(clientSpy.callHistory.count == 1, "Number of requests should be equal to one, as only once call was requested.")
+        let url = try XCTUnwrap(URLComponents(string: clientSpy.callHistory.first!.description)?.url, "Problem when trying to get url.")
```
pawelgil, line 21: this will crash if `callHistory` is empty, use `XCTUnwrap`

AirCastingTests/SearchAndFollowTests/AirBeamMeasurementDownloaderTests.swift
```diff
@@ -0,0 +1,57 @@
...
+        XCTAssertEqual(url.path, "/api/fixed/sessions/\(Int.default)/streams.json")
+    }
+    
+    func test_requestingDataFromServer_usesGetMethod() {
+        sut.downloadStreams(with: Int.default, completion: { _ in })
+        
+        XCTAssert(clientSpy.callHistory.count == 1, "Number of requests should be equal to one, as only once call was requested.")
+        XCTAssertEqual(clientSpy.callHistory.first!.httpMethod, "GET")
```
pawelgil, line 29: this will crash too, `XCTUnwrap`

AirCastingTests/SearchAndFollowTests/AirBeamMeasurementDownloaderTests.swift
```diff
@@ -0,0 +1,57 @@
...
+    }
+    
+    func test_requestingDataFromServer_usesGivenSesionUUID() throws {
+        let sessionUUID = 123456-789
+        sut.downloadStreams(with: sessionUUID, completion: { _ in })
+        
+        XCTAssert(clientSpy.callHistory.count == 1, "Number of requests should be equal to one, as only once call was requested.")
+        let url = try XCTUnwrap(URLComponents(string: clientSpy.callHistory.first!.description)?.url, "Problem when trying to get url.")
```
pawelgil, line 37: this will crash too, `XCTUnwrap`

AirCastingTests/SearchAndFollowTests/AirBeamMeasurementDownloaderTests.swift
```diff
@@ -0,0 +1,57 @@
...
+        XCTAssertEqual(url.path, "/api/fixed/sessions/\(sessionUUID)/streams.json")
+    }
+    
+    func test_requestingDataFromServer_usesCorrectNameForLimit() throws {
+        sut.downloadStreams(with: Int.default, completion: { _ in })
+        
+        XCTAssert(clientSpy.callHistory.count == 1, "Number of requests should be equal to one, as only once call was requested.")
+        let queryItem = try XCTUnwrap(URLComponents(string: clientSpy.callHistory.first!.description)?.queryItems, "Problem when trying to get url.")
```
pawelgil, line 45: this will crash too, `XCTUnwrap`

AirCastingTests/SearchAndFollowTests/AirBeamMeasurementDownloaderTests.swift
```diff
@@ -0,0 +1,57 @@
...
+        XCTAssert(queryItem.contains(where: { $0.name == "measurements_limit" }))
+    }
+    
+    func test_requestingDataFromServer_givesCorrectMeasurementNumber() throws {
+        sut.downloadStreams(with: Int.default, completion: { _ in })
+        
+        XCTAssert(clientSpy.callHistory.count == 1, "Number of requests should be equal to one, as only once call was requested.")
+        let queryItem = try XCTUnwrap(URLComponents(string: clientSpy.callHistory.first!.description)?.queryItems, "Problem when trying to get url.")
```
pawelgil, line 53: this will crash too, `XCTUnwrap`

AirCastingTests/SearchAndFollowTests/SessionForLocationDownloaderTests.swift
```diff
@@ -0,0 +1,79 @@
...
+    override func setUp() {
+        super.setUp()
+        Resolver.test.register { self.clientSpy as APIClient }
+    }
+    
+    func test_requestingDataFromServer_usesGetMethod() {
+        clientSpy.returning((data: Data.default, response: .success()))
+        requestMockSession()
+        XCTAssert(clientSpy.callHistory.count == 1, "Number of requests should be equal to one, as only once call was requested.")
+        XCTAssertEqual(clientSpy.callHistory.first!.httpMethod, "GET")
```
pawelgil, line 20: this will crash too, `XCTUnwrap`

AirCastingTests/SearchAndFollowTests/SessionForLocationDownloaderTests.swift
```diff
@@ -0,0 +1,79 @@
...
+    }
+    
+    func test_creatingQueryItems_usesSmallQAsParameterName() throws {
+        clientSpy.returning((data: Data.default, response: .success()))
+        requestMockSession()
+        XCTAssert(clientSpy.callHistory.count == 1, "Number of requests should be equal to one, as only once call was requested.")
+        let firstQueryItems = try XCTUnwrap(URLComponents(string: clientSpy.callHistory.first!.description)?.queryItems?.first, "Problem when trying to get query items from url.")
```
pawelgil, line 27: this will crash too, `XCTUnwrap`

AirCastingTests/SearchAndFollowTests/SessionForLocationDownloaderTests.swift
```diff
@@ -0,0 +1,79 @@
...
+        XCTAssertEqual(firstQueryItems.name, "q")
+    }
+    
+    func test_creatingUrl_usesCorrectPath() throws {
+        clientSpy.returning((data: Data.default, response: .success()))
+        requestMockSession()
+        XCTAssert(clientSpy.callHistory.count == 1, "Number of requests should be equal to one, as only once call was requested.")
+        let url = try XCTUnwrap(URLComponents(string: clientSpy.callHistory.first!.description)?.url, "Problem when trying to get url request.")
```
pawelgil, line 35: this will crash too, `XCTUnwrap`

pawelgil (changes requested):
some `XCTUnwrap` action needed

PR #837, opened 2022-08-08
Build incrementation fix

jankrzempek:
For now, the base version if somewhere = 1 and somewhere = 2 which result in push Build #2 to Firebase 3rd time.

Maybe it can be resolved better, but I know this is working. Did my best for now

To test:
run `fastlane run increment_build_number` and see the results before and after use of this PR


anna1901 (changes requested):
I do not understand the description and changes. What is the problem? And why is it solved by changing current_project_version?

anna1901 (approved):
I do not understand the description, and why should it be tested. But I see that project version was changed by mistake in one of the previous PRs, so it's good that you're changing it back. 👍

PR #838, opened 2022-08-10
Feature flags default value changes to always true.

jankrzempek:
https://trello.com/c/kuNnq5DC/774-cleanup-feature-flags


anna1901 (changes requested):
Let's just change values. I can imagine a situation when we will want to turn off some flags by default and leave some on. Also when adding a new flag, if we will have this default value it will be easy to forget to put it here and set it to false when not i DEBUG

anna1901 (approved):

pawelgil (approved):

PR #840, opened 2022-08-11
Threshold alerts

anna1901:
https://trello.com/c/qtsaM8EY


anna1901:
[fixed & following: threshold alert](https://trello.com/c/qtsaM8EY/744-fixed-following-threshold-alert)

AirCasting/SessionViews/ThresholdAlerts/ThresholdAlertSheet.swift
```diff
@@ -0,0 +1,93 @@
+// Created by Lunar on 01/08/2022.
+//
+
+import SwiftUI
+import AirCastingStyling
+
+struct ThresholdAlertSheet<VM: ThresholdAlertSheetViewModel>: View {
```
pawelgil, line 7: we resigned from injecting VMs into Views a while ago, now they just create them themselves which is a bit less Generic-awkwardy. We can re-discuss this tho if you want

AirCasting/SessionViews/ThresholdAlerts/ThresholdAlertsController.swift
```diff
@@ -0,0 +1,167 @@
...
+                case .success(_):
+                    Log.info("Created alert for: \(alert.sensorName)")
+                case .failure(let error):
+                    Log.error("Failed to create an alert on backend: \(error)")
+                    failed = true
+                }
+                group.leave()
+            }
+        }
+        group.wait()
```
pawelgil, line 130: same - maybe notify will be a better option here

AirCasting/SessionViews/ThresholdAlerts/ThresholdAlertsController.swift
```diff
@@ -0,0 +1,167 @@
...
+                        group.leave()
+                    }
+                case .failure(let error):
+                    Log.error("Failed to delete an alert on backend: \(error)")
+                    failed = true
+                    group.leave()
+                }
+            }
+        }
+        group.wait()
```
pawelgil, line 160: same - maybe notify will be a better option here

AppDelegate+Injection.swift
```diff
@@ -255,18 +256,18 @@ extension Resolver: ResolverRegistering {
...
             self.children = children
         }
-        
+
         func isFeatureOn(_ feature: FeatureFlag) -> Bool? {
             children.compactMap { $0.isFeatureOn(feature) }.first
         }
     }
-    
+
```
pawelgil, line 270: would it be possible to revert unneeded changes here? I think this file is untouched except whitespaces and this is a pretty conflict-prone one

pawelgil (changes requested):

AirCasting/SessionViews/ThresholdAlerts/ThresholdAlertsController.swift
```diff
@@ -0,0 +1,167 @@
...
+                let alertsOfTheSession = existingAlerts.filter({ $0.sessionUuid == sessionUUID.rawValue })
+                completion(.success(alertsOfTheSession))
+            case .failure(let error):
+                completion(.failure(error))
+            }
+        }
+    }
+    
+    func processAlerts(delete toDelete: [DeleteAlertData], create toCreate: [NewThresholdAlertData], update toUpdate: [UpdateAlertData], completion: @escaping (Result<Void, ThresholdAlertsError>) -> Void) {
+        var failed = false
```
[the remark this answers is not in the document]
anna1901: We used this approach in the app a couple of times already, but maybe it's time to switch to some better alternative. Could you tell me how this could be properly done?

AirCasting/SessionViews/ThresholdAlerts/ThresholdAlertSheetViewModel.swift
```diff
@@ -0,0 +1,188 @@
...
+                }
+                let sessionStreams = self.session.sortedStreams.filter( {!$0.gotDeleted} )
+                var newAlertOptions: [AlertOption] = []
+                var i = 0
+                sessionStreams.forEach { stream in
+                    guard let sensorName = stream.sensorName else { return }
+                    let streamAlert = alerts.first(where: { $0.sensorName == self.shorten(sensorName) })
+                    newAlertOptions.append(AlertOption(id: i, sensorName: sensorName, shortSensorName: self.shorten(sensorName), isOn: streamAlert != nil, thresholdValue: streamAlert?.thresholdValue ?? "", frequency: streamAlert?.frequency ?? .oneHour))
+                    i+=1
+                }
```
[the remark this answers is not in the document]
anna1901: Yes, it can. Thanks :)

AirCasting/SessionViews/ThresholdAlerts/ThresholdAlertsController.swift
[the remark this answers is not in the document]
pawelgil: > 

@anna1901 It's a common approach, I'm also guilty of doing that. The reason is that it's almost impossible to hit a race condition with network callbacks. The intervals between server responses are **years** in the CPU frame of reference. But we always need to asses if that's the case - if those were CoreData calls for example I wouldn't be so sure about it.

To make it right here we'd probably add some lock (mutex) around this bool value

AirCasting/SessionViews/ThresholdAlerts/ThresholdAlertSheetViewModel.swift
```diff
@@ -0,0 +1,164 @@
...
+                
+                DispatchQueue.main.async {
+                    self.streamOptions = newAlertOptions
+                    self.activeAlerts = .ready(alerts)
+                }
+            case .failure(let error):
+                Log.error(error.localizedDescription)
+                DispatchQueue.main.async {
+                    self.alert = InAppAlerts.failedThresholdAlertsFetchingAlert {
+                        DispatchQueue.main.async {
```
pawelgil, line 152: is this needed?

pawelgil (approved):

PR #844, opened 2022-08-12
Allow for editing not synced sessions

anna1901:
Before, user couldn't edit a session if it wasn't previously sent to backend. It was happening because when entering session editing view we were attempting to get its most recent data from backend and it was failing because session wasn't there. Now, we will allow users to edit sessions that weren't sent to backend (because eg. user has "sync only through wifi" turned on).


anna1901:
[[Wi-Fi Sync] Edit session not allowed when Wi-Fi Sync option is on](https://trello.com/c/otUXNaeK/773-wi-fi-sync-edit-session-not-allowed-when-wi-fi-sync-option-is-on)

pawelgil (approved):

PR #855, opened 2022-08-23
Expand card when doing a mobile session.

jankrzempek:
**What was done:**
The initial state of the card should be set to _expanded_ when doing a mobile session (BT or MIC)

https://trello.com/c/4zpZcHT1/784-mobile-active-make-expanded-card-the-default-state


anna1901 (approved):
Looks good 🥳

PR #860, opened 2022-08-24
Create alert with full sensor name

anna1901:
Turns out we need to sent a full sensor name for the alerts to work


jankrzempek (approved):
Awesome! 🤩

PR #861, opened 2022-08-24
Assign the same timestamp to given measurements chunk.

jankrzempek:
We were having different timestamps assigned to one chunk (5 of them: F, PM1..) of measurements. The reason is that: we were saving each one of them separately which sometimes resulted in a slight delay.

It should be tested for +2 h and +9 hours sessions, but I think Michael and Grzester can do it later, as for now, those early tests done by me were successful.

https://trello.com/c/Y4OVpvQH/347-improve-2hr-9hr-averaging-function-so-all-the-averages-are-precisely-5-sec-1-min

### DIFF
--- a/AirCasting/ABConnector/BluetoothManager.swift
+++ b/AirCasting/ABConnector/BluetoothManager.swift
@@ -34,18 +34,6 @@ extension BluetoothCommunicator {
 
 class BluetoothManager: NSObject, BluetoothCommunicator, ObservableObject {
     
-    class PeripheralMeasurementTimeManager {
-        private(set) var collectedValuesCount: Int = 5
-        private(set) var currentTime: Date = DateBuilder.getRawDate().currentUTCTimeZoneDate
-        
-        func startNewValuesRound() {
-            currentTime = DateBuilder.getRawDate().currentUTCTimeZoneDate
-            collectedValuesCount = 0
-        }
-        
-        func incrementCounter() { collectedValuesCount += 1 }
-    }
-
     lazy var centralManager: CBCentralManager = {
         let centralManager = CBCentralManager()
         centralManager.delegate = self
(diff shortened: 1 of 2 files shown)


AirCasting/ABConnector/BluetoothManager.swift
```diff
@@ -33,6 +33,18 @@ extension BluetoothCommunicator {
 }
 
 class BluetoothManager: NSObject, BluetoothCommunicator, ObservableObject {
+    
+    class PeripheralMeasurementTimeManager {
```
anna1901, line 37: Could we move this logic to mobilePeripheralSessionManager, so that we won't add more session-related code to BluetoothManager?

AirCasting/ABConnector/BluetoothManager.swift
```diff
@@ -33,6 +33,18 @@ extension BluetoothCommunicator {
...
+        private(set) var collectedValuesCount: Int = 5
+        private(set) var currentTime: Date = DateBuilder.getRawDate().currentUTCTimeZoneDate
+        
+        func startNewValuesRound() {
+            currentTime = DateBuilder.getRawDate().currentUTCTimeZoneDate
```
anna1901, line 42: We have a DateBuilder method for this: `getFakeUTCDate`

pawelgil (approved):

pawelgil:
Adding more domain specific stuff to the `BluetoothManager` is not great, but I know we don't have a choice here without a major refactor

AirCasting/Assets.xcassets/Colors/AircastingSecondaryBackground.colorset/Contents.json
```diff
@@ -5,9 +5,9 @@
         "color-space" : "srgb",
         "components" : {
           "alpha" : "0.250",
-          "blue" : "1.000",
-          "green" : "0.941",
-          "red" : "0.878"
+          "blue" : "255",
```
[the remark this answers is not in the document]
jankrzempek: Thx! I was checking something for Mohsen

AirCasting/ABConnector/BluetoothManager.swift
jankrzempek: oo, I did not come up with that kind of idea - will be changed!

PR #863, opened 2022-08-25
Prevent a user from setting the same number for different thresholds.

jankrzempek:
https://trello.com/c/rQkwZeTW/790-sensorthreshold-make-sure-that-values-for-thresholds-are-ascending


AirCasting/SessionViews/ThresholdSettings/ThresholdsSettingsViewModel.swift
```diff
@@ -37,10 +37,13 @@ class ThresholdSettingsViewModel: ObservableObject {
             return
         }
         
-        let newValues: [Int32] = [thresholdVeryHigh, thresholdHigh, thresholdMedium, thresholdLow, thresholdVeryLow]
+        var newValues: [Int32] = [thresholdVeryHigh, thresholdHigh, thresholdMedium, thresholdLow, thresholdVeryLow]
             .map { formatter.value(from: $0) ?? 0 }
             .sorted { $0 < $1 }
         
+        // Prevents us from the situation when the user could change thresholds to have the same values
+        for value in (1...newValues.count - 1).reversed() { if newValues[value - 1] == newValues[value] { newValues[value - 1] -= 1 }}
```
pawelgil, line 45: I don't like how unreadable this is. I wrote a simple extension for `Sequence` that is basically a `map` with an ability to have a peek what the next `Element` is:
```Swift
extension Sequence {
    func mapWithNext(_ map: ((Element, Element) -> Element)) -> [Element] {
        var iter = self.makeIterator()
        var prev = iter.next()
        guard prev != nil else { return [] }
        
        var pairs: [(Element, Element)] = []
        while let elem = iter.next() {
            pairs.append((prev!, elem))
            prev = elem
        }
        var transformed = pairs.map { map($0.0, $0.1) }
        transformed.append(pairs.last!.1)
        return transformed
    }
}
```

We could use it here like that:
```Swift
newValues.reversed().mapWithNext { lowerThreshold, higherThreshold in
    guard lowerThreshold != higherThreshold else { return lowerThreshold - 1.0 }
    return lowerThreshold
} 
```

AirCasting/Extensions/Sequence.swift
```diff
@@ -0,0 +1,21 @@
+// Created by Lunar on 30/08/2022.
```
pawelgil, line 1: just one more thing - please rename this file to be of `Sequence+MapWithNext.swift`. This is a standard convention for files containing extensions. Alternatively we can just make it `Sequence+Helpers.swift` so in the future we can dump all convenience extensions in a single file

pawelgil (approved):

AirCasting/SessionViews/ThresholdSettings/ThresholdsSettingsViewModel.swift
```diff
@@ -37,10 +37,16 @@ class ThresholdSettingsViewModel: ObservableObject {
             return
         }
         
-        let newValues: [Int32] = [thresholdVeryHigh, thresholdHigh, thresholdMedium, thresholdLow, thresholdVeryLow]
+        var newValues: [Int32] = [thresholdVeryHigh, thresholdHigh, thresholdMedium, thresholdLow, thresholdVeryLow]
             .map { formatter.value(from: $0) ?? 0 }
             .sorted { $0 < $1 }
         
+        // Prevents us from the situation when the user could change thresholds to have the same values
+        newValues = newValues.reversed().mapWithNext { lowerThreshold, higherThreshold in
```
anna1901, line 45: You are getting those elements here in reversed order actually, so it should be `higherThreshold, lowerThreshold`. It still works, but only because afterwards the values are sorted again. But this logic is a bit misleading and works kind of accidentally. :D @jankrzempek @pawelgil

PR #866, opened 2022-08-29
Selects PM2.5 as default stream for every AB mobile session.

jankrzempek:
https://trello.com/c/M7CnOYCg/793-11102-ab3-mobile-active-session-default-parameter-should-be-pm25


anna1901 (commented):
When I did it without breakpoints it did work, but this means this solution is not "bulletproof"

AirCasting/SessionViews/SessionCardView.swift
```diff
@@ -117,11 +117,15 @@ struct SessionCardView: View {
     }
 
     private func selectDefaultStreamIfNeeded(streams: [MeasurementStreamEntity]) {
-        if selectedStream == nil {
-            if let newStream = session.streamWith(sensorName: session.userInterface?.sensorName ?? "") {
-                return selectedStream = newStream
+        // AirBeam is giving us two streams at first (F, RH) when doing mobile active, which prevents us from switching to PM2.5 as selected stream.
+        // Solution below is not ideal, as in the future (this was written in 29/08/2022) the number of streams can change. Yet, it is working well - selecting PM2.5 as deafult stream for every session.
+        if streams.count == 5 || streams.count == 1 {
```
pawelgil, line 122: 1. Use `guard` whenever the `if` statement is the last statement in a scope. This way you'll avoid indentation. Here especially since you right away have a next `if` inside (which should be changed to `guard` too)
2. Why don't you do a lookup in this `streams` array instead? Pseudocode:
```Swift
if streams.contains(session.userInterface?.sensorName) { 
     selectedStream = session.userInterface?.sensorName 
} else if streams.contains(session.defaultStreamSelection()) {
    selectedStream = session.defaultStreamSensorName
} else {
    selectedStream = streams.first
}
```

AirCasting/SessionViews/SessionCardView.swift
```diff
@@ -117,11 +117,15 @@ struct SessionCardView: View {
...
+            if selectedStream == nil {
+                if let newStream = session.streamWith(sensorName: session.userInterface?.sensorName ?? "") {
+                    return selectedStream = newStream
```
pawelgil, line 125: this is fancy, but let's not do it 😅 and change the order to a standard one: `selectedStream = newStream; return`

pawelgil (changes requested):

AirCasting/SessionViews/SessionCardView.swift
jankrzempek: 1. I will do that, thx!
2. Having this kind of solution, I think we will execute something every time this is called. The case here is to wait until we have all 5 streams so that we can make the right selection. If I am wrong please correct me

AirCasting/SessionViews/SessionCardView.swift
jankrzempek: > When I did it without breakpoints it did work, but this means this solution is not "bulletproof"

Where did you put those breakpoints at first? I was trying to reproduce, but even with using those breakpoints, I can get it right (PM2.5 is selected). If something like this happens (without breakpoints it works only), can we be sure it can be made "bulletproof" somehow?- I mean, we are using BT and receiving data constantly so maybe this is something we are not able to handle. I am asking because I don't have another idea for this right now.

AirCasting/SessionViews/SessionCardView.swift
pawelgil: @jankrzempek let's discuss after you're back from your PTO

anna1901 (approved):

pawelgil (approved):

PR #867, opened 2022-08-29
The same location for every 5 measurements.

jankrzempek:
Ensures that every measurement bunch gets the same location. This was a missing part after applying the same time for every bunch.

https://trello.com/c/Y4OVpvQH/347-improve-2hr-9hr-averaging-function-so-all-the-averages-are-precisely-5-sec-1-min


anna1901 (approved):
Looks good :)

PR #868, opened 2022-08-29
Change share sheet viewModel to StateObject

anna1901:
This PR fixes a bug that was causing the share sheet to refresh every time a new measurement was downloaded. Because of that, the input of the user was cleared and that could have been an annoying experience.


jankrzempek (approved):
I cannot see the difference honestly, but as it is still working well I am going to approve 🥳

PR #869, opened 2022-08-29
Clearing sessions during reorder

anna1901:
# What was done?

The user has the possibility to unfollow a fixed session by swiping it to the left during reordering

https://trello.com/c/cOvvI3rG

### DIFF
--- a/AirCasting/Dashboard/Reordering/CustomSwipeAction.swift
+++ b/AirCasting/Dashboard/Reordering/CustomSwipeAction.swift
@@ -1,0 +1,239 @@
+// Created by Lunar on 29/08/2022.
+//
+
+import Foundation
+import SwiftUI
+
+struct SwipeActionButton: View, Identifiable {
+  static let width: CGFloat = 70
+
+  let id = UUID()
+  let text: Text?
+  let icon: Image?
+  let action: () -> Void
+  let tint: Color?
+
+  init(text: Text? = nil,
+       icon: Image? = nil,
+       action: @escaping () -> Void,
+       tint: Color? = nil) {
+    self.text = text
+    self.icon = icon
+    self.action = action
+    self.tint = tint ?? .gray
+  }
+
+  var body: some View {
+    ZStack {
+      tint
+      VStack {
+        icon?
+          .foregroundColor(.white)
+        if icon == nil {
+          text?
+            .foregroundColor(.white)
+        }
+      }
+      .frame(width: SwipeActionButton.width)
+    }
+  }
+}
+
+// Adds custom swipe actions to a given view
+struct SwipeActionView: ViewModifier {
+  // How much does the user have to swipe at least to reveal buttons on either side
+  private static let minSwipeableWidth = SwipeActionButton.width * 0.8
+
+  // Buttons at the leading (left-hand) side
+  let leading: [SwipeActionButton]
+  // Can you full swipe the leading side
+  let allowsFullSwipeLeading: Bool
+  // Buttons at the trailing (right-hand) side
+  let trailing: [SwipeActionButton]
+  // Can you full swipe the trailing side
+  let allowsFullSwipeTrailing: Bool
+
+  private let totalLeadingWidth: CGFloat!
+  private let totalTrailingWidth: CGFloat!
+
+  @State private var offset: CGFloat = 0
+  @State private var prevOffset: CGFloat = 0
+
+  init(leading: [SwipeActionButton] = [],
+       allowsFullSwipeLeading: Bool = false,
+       trailing: [SwipeActionButton] = [],
+       allowsFullSwipeTrailing: Bool = false) {
+    self.leading = leading
+    self.allowsFullSwipeLeading = allowsFullSwipeLeading && !leading.isEmpty
+    self.trailing = trailing
+    self.allowsFullSwipeTrailing = allowsFullSwipeTrailing && !trailing.isEmpty
+    totalLeadingWidth = SwipeActionButton.width * CGFloat(leading.count)
+    totalTrailingWidth = SwipeActionButton.width * CGFloat(trailing.count)
+  }
+
+  func body(content: Content) -> some View {
+    // Use a GeometryReader to get the size of the view on which we're adding
+    // the custom swipe actions.
+    GeometryReader { geo in
+      // Place leading buttons, the wrapped content and trailing buttons
+      // in an HStack with no spacing.
+      HStack(spacing: 0) {
+        // If any swiping on the left-hand side has occurred, reveal
+        // leading buttons. This also resolves button flickering.
+        if offset > 0 {
+          // If the user has swiped enough for it to qualify as a full swipe,
+          // render just the first button across the entire swipe length.
+          if fullSwipeEnabled(edge: .leading, width: geo.size.width) {
+            button(for: leading.first)
+              .frame(width: offset, height: geo.size.height)
+          } else {
+            // If we aren't in a full swipe, render all buttons with widths
+            // proportional to the swipe length.
+            ForEach(leading) { actionView in
+              button(for: actionView)
+              .frame(width: individualButtonWidth(edge: .leading),
+                     height: geo.size.height)
+            }
+          }
+        }
+
+        // This is the list row itself
+        content
+          // Add horizontal padding as we removed it to allow the
+          // swipe buttons to occupy full row height.
+          .padding(.horizontal, 16)
+          .frame(width: geo.size.width, height: geo.size.height, alignment: .leading)
+          .offset(x: (offset > 0) ? 0 : offset)
+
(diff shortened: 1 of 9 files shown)


AirCasting/Dashboard/Reordering/CustomSwipeAction.swift
```diff
@@ -0,0 +1,79 @@
...
+                                 coordinateSpace: .local)
+                .onChanged { gesture in
+                    offset = gesture.translation.width
+                }
+                .onEnded { _ in
+                    if !checkAndHandleFullSwipe(for: trailingAction, edge: .trailing, width: -UIScreen.main.bounds.size.width) {
+                        offset = 0
+                    }
+                    prevOffset = offset
+                })
```
jankrzempek, line 38: Indentation seems weird here, I know XCode likes it but just pointing it out

jankrzempek: 1/2 I think you can use a shorter form here, please try to use this code here and on 2/2 part of the one that I will provide, together. This is just a suggestion but I think it can be cleaner:
```
 .onEnded { _ in
                    checkAndHandleFullSwipe(for: trailingAction, edge: .trailing, width: -UIScreen.main.bounds.size.width)
                    prevOffset = offset
                })
```

AirCasting/Dashboard/Reordering/CustomSwipeAction.swift
```diff
@@ -0,0 +1,79 @@
...
+            action()
+            DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
+                offset = 0
+                prevOffset = 0
+            }
+            return true
+        } else {
+            return false
+        }
+    }
```
jankrzempek, line 68: 2/2 Herę
```
    // Checks if the view is in full swipe. If so, trigger the action
    private func checkAndHandleFullSwipe(for action: () -> Void,
                                         edge: Edge,
                                         width: CGFloat) {
        if fullSwipeEnabled(edge: edge, width: width) {
            offset = width * 1.2
            action()
        } else {
            offset = 0
        }
    }
```

AirCasting/MainTabBarView.swift
```diff
@@ -104,7 +104,7 @@ private extension MainTabBarView {
                         if !searchAndFollow.isHidden && featureFlagsViewModel.enabledFeatures.contains(.searchAndFollow) && selectedSection.section == .following {
                             searchAndFollowButton
                         }
-                        if !reorderButton.isHidden && sessions.count > 1 && selectedSection.section == .following {
+                        if !reorderButton.isHidden && (reorderButton.reorderIsOn || sessions.count > 1) && selectedSection.section == .following {
```
jankrzempek, line 107: Please tell me, why do we need to have it here. I can't get it 😅

jankrzempek (approved):
Working awesome, swiping so smooth! 🥳
Please consider those two suggestion comments and answer for me this mysterious `reorderIsOn`

AirCasting/Dashboard/Reordering/CustomSwipeAction.swift
```diff
@@ -0,0 +1,79 @@
+// Created by Lunar on 29/08/2022.
+//
+
+import Foundation
+import SwiftUI
+
+// Adds custom swipe action to a given view
```
pawelgil, line 7: you can use a triple `/` to make a documentation out of this comment (the benefit of a documentation comment is that it will pop up whenever the user alt-clicks to reveal quick help. For example:
image

Which I think is quite helpful. We can also build the whole projects documentation and have it as a reference in XCode just like the standard SDKs have (or in the future export it to a webpage/wiki/etc).

I think that's a great value for just one more slash :)

AirCasting/Dashboard/Reordering/CustomSwipeAction.swift
```diff
@@ -0,0 +1,79 @@
...
+                                         width: CGFloat) -> Bool {
+        if fullSwipeEnabled(edge: edge, width: width) {
+            offset = width * 1.2
+            action()
+            DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
+                offset = 0
+                prevOffset = 0
+            }
+            return true
+        } else {
```
pawelgil, line 65: else is not needed here since you `return` from the `if` statement

AirCasting/Dashboard/Reordering/ReorderingDashboard.swift
```diff
@@ -20,13 +20,19 @@ struct ReorderingDashboard: View {
...
+                        .overlay(viewModel.currentlyDraggedSession?.uuid == session.uuid && changedView ? Color.aircastingWhite.opacity(0.8) : Color.clear)
                         .onDrag({
                             viewModel.currentlyDraggedSession = session
                             changedView = false
                             return NSItemProvider(object: String(describing: session.uuid) as NSString)
                         })
                         .onDrop(of: [.text], delegate: DropViewDelegate(sessionAtDropDestination: session, currentlyDraggedSession: $viewModel.currentlyDraggedSession, sessions: $viewModel.sessions, changedView: $changedView))
+                        .trailingSwipeAction {
+                            withAnimation(.easeOut(duration: 0.5)) {
+                                viewModel.sessions.removeAll(where: { $0.uuid == session.uuid })
```
pawelgil, line 32: let's not operate on a viewmodels properties directly like that. The correct way will be to create a special purpose VM method for that

AirCasting/Dashboard/Reordering/ReorderingDashboardViewModel.swift
```diff
@@ -10,16 +10,37 @@ class ReorderingDashboardViewModel: ObservableObject {
...
+    @Injected private var externalSessionsStore: ExternalSessionsStore
+    @Injected private var measurementStreamStorage: MeasurementStreamStorage
     
     init(sessions: [Sessionable], thresholds: [SensorThreshold]) {
         self.sessions = sessions
         self.thresholds = thresholds
     }
     
+    func clear(session: Sessionable) {
+        if session.isExternal {
```
pawelgil, line 22: plz move the body of `if` and `else` to separate private methods

AirCasting/Dashboard/Reordering/ReorderingDashboardViewModel.swift
```diff
@@ -10,16 +10,37 @@ class ReorderingDashboardViewModel: ObservableObject {
...
+            externalSessionsStore.deleteSession(uuid: session.uuid) { result in
+                switch result {
+                case .success:
+                    Log.info("Deleted external session")
+                case .failure(let error):
+                    Log.error("Failed to delete External Session: \(error)")
```
pawelgil, line 28: If `clear` fails we'll have an invalid UI state since the removal didn't succeed. It's quite unlikely, but maybe we can add a reload here? So that whenever this delete fails the session will pop back on the UI

AirCasting/MainTabBarView.swift
pawelgil: I agree, maybe we should pull this part `(reorderButton.reorderIsOn || sessions.count > 1)` to a helper variable and give it a descriptive name? Or comment this line (but always prefer helper variables over comments since those are being checked by the compiler and will produce a warning if unused)

pawelgil (changes requested):

AirCasting/Dashboard/Reordering/CustomSwipeAction.swift
anna1901: So you think this part:
```
DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
                offset = 0
                prevOffset = 0
            }
```
is unnecessary?

AirCasting/MainTabBarView.swift
anna1901: This is added because otherwise, the "Finish" button disappeared if I cleared all of the sessions (or all except one). Now, the button stays even if there are less than 2 sessions as long as reordering is on. I don't know how could I call such variable

anna1901: But I can change the logic here and hopefully it will make it more clear :)

AirCasting/Dashboard/Reordering/CustomSwipeAction.swift
anna1901: Thank you 🙂 It does look better

AirCasting/Dashboard/Reordering/ReorderingDashboardViewModel.swift
anna1901: Hm, but then the user might try to swipe again and again and most likely it still wouldn't work. I'm not sure what is the best way to handle this failure, but I would just keep it as is and keep an eye out for this error popping up in firebase 👀

pawelgil (approved):

PR #872, opened 2022-08-31
Icon for rearrange dissapears.

jankrzempek:
https://trello.com/c/Z3CVr3Pg/794-11102-following-rearrange-and-search-icons-are-missing

### DIFF
--- a/AirCasting/Dashboard/DashboardView.swift
+++ b/AirCasting/Dashboard/DashboardView.swift
@@ -80,6 +80,9 @@ struct DashboardView: View {
         .onAppear() {
             try! coreDataHook.setup(selectedSection: self.selectedSection.section)
         }
+        .onChange(of: selectedSection.section) { newValue in
+            try! coreDataHook.setup(selectedSection: newValue)
+        }
     }
     
     private var customNavigationBar: some View {


pawelgil (approved):

anna1901 (commented):
Doesn't this change make us call Core Data for sessions twice from different places every time selectedSection changes? Is that okay? Maybe in SessionsListView we could set some variable on ReorderButton and switch the `coreDataHook.sessions > 1` condition in MainTabBar view to checking this variable? What do you think about such solution?

jankrzempek:
> Doesn't this change make us call Core Data for sessions twice from different places every time selectedSection changes? Is that okay? Maybe in SessionsListView we could set some variable on ReorderButton and switch the `coreDataHook.sessions > 1` condition in MainTabBar view to checking this variable? What do you think about such solution?

To be honest I don't know if that's okey, but I hope so 😅. @pawelgil could you please help in this question?
In addition to the change, I am not standing strong with the current implementation so that we couldn't change it. The one thing which I have in my mind is: whenever we will change something in ListView now, it won't affect the idea that is going to be applied here. How do you feel strong about this change, should I proceed?

pawelgil:
> > Doesn't this change make us call Core Data for sessions twice from different places every time selectedSection changes? Is that okay? Maybe in SessionsListView we could set some variable on ReorderButton and switch the `coreDataHook.sessions > 1` condition in MainTabBar view to checking this variable? What do you think about such solution?
> 
> To be honest I don't know if that's okey, but I hope so 😅. @pawelgil could you please help in this question? In addition to the change, I am not standing strong with the current implementation so that we couldn't change it. The one thing which I have in my mind is: whenever we will change something in ListView now, it won't affect the idea that is going to be applied here. How do you feel strong about this change, should I proceed?

TBH I'm not sure I understand the alternative. @anna1901 this code fixes the bug where the Dashboards `coreDataHook` was not updated when sections were changing (`onAppear` happens basically only on dashboard startup and modals dismissal).

jankrzempek:
> > > Doesn't this change make us call Core Data for sessions twice from different places every time selectedSection changes? Is that okay? Maybe in SessionsListView we could set some variable on ReorderButton and switch the `coreDataHook.sessions > 1` condition in MainTabBar view to checking this variable? What do you think about such solution?
> > 
> > 
> > To be honest I don't know if that's okey, but I hope so 😅. @pawelgil could you please help in this question? In addition to the change, I am not standing strong with the current implementation so that we couldn't change it. The one thing which I have in my mind is: whenever we will change something in ListView now, it won't affect the idea that is going to be applied here. How do you feel strong about this change, should I proceed?
> 
> TBH I'm not sure I understand the alternative. @anna1901 this code fixes the bug where the Dashboards `coreDataHook` was not updated when sections were changing (`onAppear` happens basically only on dashboard startup and modals dismissal).

So as I understand, the alternative would be to have some kind of observable which could be updated as the SessionListView updated - then we could retrieve value of it in MainTabBarView

anna1901 (changes requested):
The thing is that this way we have two CoreDataHooks objects which are holding all fetched sessions in memory, and for this one which is used here in dashboard we only need it to know the number of sessions to decide if we want to show the button or not. It seems like it's not worth to fetch all sessions only for that. So I was thinking, that maybe we can fetch sessions only in SessionsListView and there we can change some variable on Reordering Button (eg. `ReorderingButton.sessionsCountThresholdExceeded`) to manage its visibility.

anna1901 (approved):

PR #874, opened 2022-09-01
SD Sync: Add timer which checks if no more new data is being sent

anna1901:
# What was done?

This PR adds a monitoring mechanism to SDSyncAirmbeamServices. When we receive metadata payload that sd sync has finished, we start a timer that checks if any new measurements are still appearing. If no new measurement appears within 1 second, or if we already have all of the measurements that we are expecting, then we are calling a completion which informs the caller that AirBeam finished syncing.

# Why was it done?
This is a possible solution for problems with SD sync when AB is plugged in. However, it solves only the problems for AB with new Firmware.

Here is a detailed description of the problem described by Raymond:
"I did a preliminary test with the AB3 plugged in and the SD Sync failed about half way.

I really want the iOS AC app to work with the AB3 plugged in as the AB3 was intended to work that way.

It might possibly work throughout all AB3s FW versions with the changes mentioned below.

The SD Sync was working well until the corruption message popped up. I believe it is not corrupted but ended prematurely before all the data was synced because of lag.

From looking at the app code in the past I also believe if 80 percent of the data is synced or processed it is able to pass the data to the server.

The main issue is when the AB3 is plugged in the data sent is quicker than the iOS device can receive it (not certain if it is constraints of the iOS device or the iOS AC app).

I believe the message "SD_SYNC_FINISH" sent to the iOS AC app is making the SD Sync end prematurely as it is under a different UUID.

If the developers at Lunar Logic can check in the app if there is still an increment in SD Sync index count or if it is still receiving data from the AB3 the SD Sync should still be actively running even if the "SD_SYNC_FINISH" has been reported to the iOS AC app.

Also because the UUID for the SD Sync and the UUID for the "SD_SYNC_FINISH" message are both using a different channels and because the AB3 is sending faster than the iOS AC app can receive, it prematurely cuts the SD Sync off before sending all the data.

This is because the iOS AC app is lagging behind the AB3 and there is a "slack buffer" between what is sent and received and different UUID channels (one UUID is for SD Sync which has lag and one UUID is for the "SD_SYNC_FINISH" message which has no lag)."

### DIFF
--- a/AirCasting/ABConnector/SDCardAirBeamServices.swift
+++ b/AirCasting/ABConnector/SDCardAirBeamServices.swift
@@ -44,9 +44,12 @@ class BluetoothSDCardAirBeamServices: SDCardAirBeamServices {
     private var metadataCharacteristicObserver: AnyHashable?
     private var clearCardCharacteristicObserver: AnyHashable?
     
+    private var expectedMeasurementsCount: [SDCardSessionType: Int] = [:]
+    private var receivedMeasurementsCount: [SDCardSessionType: Int] = [:]
+    private var timerToken: AnyObject?
+    private let timer: TimerScheduler = FoundationTimerScheduler()
+    
     func downloadData(from peripheral: CBPeripheral, progress: @escaping (SDCardDataChunk) -> Void, completion: @escaping (Result<SDCardDownloadSummary, Error>) -> Void) {
-        var expectedMeasurementsCount: [SDCardSessionType: Int] = [:]
-        var receivedMeasurementsCount: [SDCardSessionType: Int] = [:]
         var currentSessionType: SDCardSessionType?
         
         Log.info("[SD Sync] Downloading data")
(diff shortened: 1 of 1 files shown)


jankrzempek (approved):
🦾

AirCasting/ABConnector/SDCardAirBeamServices.swift
```diff
@@ -153,8 +160,33 @@ class BluetoothSDCardAirBeamServices: SDCardAirBeamServices {
...
+        receivedMeasurementsCount = [:]
+        if let timerToken = timerToken {
+            timer.stop(token: timerToken)
+        }
         completion()
     }
+    
+    private func startMonitoringForEnd(completion: @escaping () -> Void) -> Void {
+        var checkedMeasurementsCount: [SDCardSessionType: Int] = [:]
+        timerToken = timer.schedule(every: 1) { [weak self] in
```
pawelgil, line 173: we're using a lot of time-based stuff in our core logic components, we shouldn't be using regular main-runloop bound timers. It can cause hard to track bugs. Good thing we're using abstraction (here, but not in other places)

Something to put on a todo list.

AirCasting/ABConnector/SDCardAirBeamServices.swift
```diff
@@ -153,8 +160,33 @@ class BluetoothSDCardAirBeamServices: SDCardAirBeamServices {
...
+            guard let self = self else { return }
+            Log.debug("Checking with:\n expected \(self.expectedMeasurementsCount)\n received \(self.receivedMeasurementsCount)\n checked \(checkedMeasurementsCount)")
+            guard checkedMeasurementsCount != self.receivedMeasurementsCount else {
+                Log.debug("NO NEW MEASUREMENT IN 1 SEC")
+                completion()
+                return
+            }
+            checkedMeasurementsCount = self.receivedMeasurementsCount
+            for temp in self.receivedMeasurementsCount {
+                if self.expectedMeasurementsCount[temp.key] ?? 0 < temp.value {
```
pawelgil, line 183: As far as I cal tell this is checked on a different thread than being written too (checking here on the main thread, writing on a arbitrary BT manager queue). Potential race condition. Please use locks or queues to synchronize access

pawelgil (changes requested):

pawelgil (approved):

PR #876, opened 2022-09-05
Alternative reordering button fix - coreDataHook refactor

anna1901:
# What was done?

* coreDataHook was removed from MainTabBarView and DashboardView. We are only using it in SessionsListView.
* fetch request was added to MainTabBarView to determine if home button should take user to following or mobile active tab
* fetch request was added to ReorderingDashboard view so display followed sessions
* monitoring for database changes was changed in CoreDataHook. Before we were refreshing the sessions list event if the sessions list didn't change but eg. measurements were downloaded. This resulted in a lot of views being refreshed. Now it is limited.

# Why was it done?

This was done to fix the problem with the disappearing reordering button, but most of the changes were done to improve performance and memory usage.

### DIFF
--- a/AirCasting/Dashboard/Reordering/ReorderingDashboardViewModel.swift
+++ b/AirCasting/Dashboard/Reordering/ReorderingDashboardViewModel.swift
@@ -10,15 +10,14 @@ class ReorderingDashboardViewModel: ObservableObject {
     @Published var isLoading = true
     @Published var currentlyDraggedSession: Sessionable?
     var thresholds: [SensorThreshold]
-    private var context: NSManagedObjectContext
 
     @Injected private var uiStorage: UIStorage
     @Injected private var externalSessionsStore: ExternalSessionsStore
     @Injected private var measurementStreamStorage: MeasurementStreamStorage
+    @Injected private var persistanceController: PersistenceController
 
-    init(thresholds: [SensorThreshold], context: NSManagedObjectContext) {
+    init(thresholds: [SensorThreshold]) {
         self.thresholds = thresholds
-        self.context = context
         fetchSessions()
     }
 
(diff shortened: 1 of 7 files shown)


pawelgil:
While I like most changes here, I'm not convinced of moving responsibility of managing the reordering button from Dashboard to SessionsList. We have just decoupled dashboard from list views and it kind of undos that. I don't think `CoreDataHook` doesn't belong to the `DashboardView`(sure, it could use some treatment by making it a protocol, adding injection etc.).

AirCasting/Dashboard/Reordering/ReorderingDashboardViewModel.swift
```diff
@@ -54,4 +57,32 @@ class ReorderingDashboardViewModel: ObservableObject {
...
+                let allSessions: [Sessionable] = (sessionEntities + externalSessionEntities).sorted {
+                    ($0.userInterface?.rowOrder ?? 0) > ($1.userInterface?.rowOrder ?? 0)
+                }
+                completion(allSessions)
+            } catch {
+                Log.error("Failed to fetched sessions for reordering")
+                completion([])
+            }
+        }
+    }
```
jankrzempek, line 87: Shouldn't this be more of a `CoreDataHook` part or some other `Hook` thing?

AirCasting/MainTabBarView.swift
```diff
@@ -189,9 +179,34 @@ private extension MainTabBarView {
...
+                let request = NSFetchRequest<SessionEntity>(entityName: "SessionEntity")
+                request.sortDescriptors = [NSSortDescriptor(key: "startTime", ascending: false)]
+                request.predicate = NSPredicate(format: "type == %@ AND (status == %li || status == %li || status == %li)", SessionType.mobile.rawValue,
+                                                SessionStatus.RECORDING.rawValue,
+                                                SessionStatus.DISCONNECTED.rawValue,
+                                                SessionStatus.NEW.rawValue)
+                request.fetchLimit = 1
+                let sessionEntities = try context.fetch(request)
+                completion(sessionEntities.count == 1 ? DashboardSection.mobileActive : DashboardSection.following)
+            } catch {
```
jankrzempek, line 199: Shouldn't this be more of a CoreDataHook part or some other Hook thing?

jankrzempek (approved):
In terms of "working solution" and code, everything seems accurate. I am not going to argue here with Paweł or you as I don't feel so confident in this topic yet. So please have in mind that approval is somehow limited.

PR #877, opened 2022-09-07
Upload sessions before sharing.

jankrzempek:
**WHY:**
There were sometimes measurements missing when someone wanted to share a session that was only partially uploaded.


pawelgil (approved):

anna1901 (commented):
@jankrzempek was in considered what happens if sync fails? Cause we get an error log on firebase that there was no URL for session maybe that's because the sync here was unsuccessful

PR #880, opened 2022-09-08
Changed ui in doing sync before session share.

jankrzempek:
https://trello.com/c/cfqfsCQf/814-upload-sessions-before-sharing

**The problem:**
In case the session was not synced and the person wanted to share a file, via the .CSV he could send it with missing data inside. (Logic for sync before share was implemented before)

**The solution:**
Make sync (upload only) sessions onto the backend. Use the same UI which Android does.

**Aim of this PR:** 
Improve UI so that it matches the design used in Android.


anna1901 (commented):
Could you elaborate on why this was added? Looking at this PR description or at the ticket in trello I have no idea what was the issue and how it was solved. Please, try to add more context to the PR descriptions and trello ticket so that they can serve as a standalone "documentiation" without a need to ask for context.

AirCasting/SessionViews/ShareSession/ShareSessionView.swift
```diff
@@ -107,6 +103,22 @@ struct ShareSessionView<VM: ShareSessionViewModel>: View {
         .buttonStyle(BlueButtonStyle())
     }
     
+    private var syncButton: some View {
+        Button {
+            // no action here - that's ok.
```
pawelgil, line 108: that's ok

pawelgil (approved):

anna1901 (commented):
How about this comment from Michael: "now i have to wait to share even old sessions that were already synced ages ago. in android i believe i only have to wait for recently recorded sessions to sync before i can share."? Was it already solved in the PR with logic?

PR #890, opened 2022-09-19
Hide threshold alerts feature behind a feature flag

anna1901:


jankrzempek (approved):
I am not up to date, so just to ensure - The feature flag on firebase is set to `false`, Is this right?

pawelgil (approved):

PR #891, opened 2022-09-26
Sensor threshold - values should be ascending 

jankrzempek:
Ensures that inside returned values there are no duplicates.

**The problem:**
When the user inputs more than two, the same values - returned threshold values can be wrong.

**The solution:**
Make sure to iterate over the given data as long as we will get a set of data without duplicates.

**The ticket:**
https://trello.com/c/rQkwZeTW/790-sensorthreshold-make-sure-that-values-for-thresholds-are-ascending


AirCasting/SessionViews/ThresholdSettings/ThresholdsSettingsViewModel.swift
```diff
@@ -42,10 +42,12 @@ class ThresholdSettingsViewModel: ObservableObject {
             .sorted { $0 < $1 }
         
         // Prevents us from the situation when the user could change thresholds to have the same values
-        newValues = newValues.reversed().mapWithNext { lowerThreshold, higherThreshold in
-            guard lowerThreshold != higherThreshold else { return lowerThreshold - 1 }
-            return lowerThreshold
-        }.sorted { $0 < $1 }
+        while Set(newValues.map { $0 }).count != newValues.count {
```
pawelgil, line 45: that's nice, how about some more algorithmic and universal approach to this? Take a look at this playground:
```Swift
import UIKit

let hello = [1,1,2,3,4,4,5]

extension Array where Element: Equatable {
    /// Iterates all elements and gets rid of the duplicates using provided strategy
    /// - Parameter strategy: a block invoked when conflict appears. It an `Element` that causes collision as a parameter.
    /// If this block returns `nil`, the new element won't be added to the resulting array.
    /// If it returns a value, this value will be added to the resulting array.
    /// - Returns: A new `Array` with conflicts resolved.
    func byRemovingDuplicates(strategy: (_ colliding: Element) -> Element?) -> Self {
        var buffer: Self = []
        for element in self {
            print("Element: \(element)")
            appendBuffer(&buffer, with: element, strategy: strategy)
        }
        return buffer
    }
    
    private func appendBuffer(_ buffer: inout Self, with value: Element, strategy: (_ colliding: Element) -> Element?) {
        guard buffer.contains(value) else {
            print("Appending buffer with \(value)")
            buffer.append(value)
            return
        }
        
        if let modified = strategy(value) {
            print("\(value) already present, resolving to \(modified)...")
            appendBuffer(&buffer, with: modified, strategy: strategy)
        }
    }
}

extension Array where Element: Equatable & Numeric {
    var incrementRemovalStrategy: (Element) -> Element? {
        { $0.increment() }
    }
}

extension Numeric {
    func increment() -> Self {
        self + 1
    }
}

print(hello.byRemovingDuplicates(strategy: hello.incrementRemovalStrategy))
```

What do you think?

jankrzempek: I changed small things there, please have a look what do you think?

AirCasting/Extensions/Array+byRemovingDuplicates.swift
```diff
@@ -0,0 +1,50 @@
+// Created by Lunar on 30/08/2022.
+//
+import Foundation
+
+extension Numeric {
+    func decrement() -> Self {
+        self - 1
```
anna1901, line 7: In my opinion incrementing makes a bit more sense logically. If we have a sorted array with duplicated values like: `[1, 2, 2, 3]`, then for the second `2` the appendBuffer function would change it to `0`, and the caller of this function would have to sort it to have the order right. With incrementing, the order of elements would stay the same. But it is more of a neat pick. If you prefer to decrement, I think it works okay as well.

AirCasting/Extensions/Array+byRemovingDuplicates.swift
```diff
@@ -0,0 +1,50 @@
...
+    /// If this func returns `nil`, the new element won't be added to the resulting array.
+    /// If it returns a value, this value will be added to the resulting array.
+    /// - Returns: A new `Array` with conflicts resolved.
+    
+    enum StrategyType {
+        case decrementation
+        case none
+    }
+    
+    func byRemovingDuplicates(strategy: Self.StrategyType) -> Self {
```
anna1901, line 23: I don't like the name "byRemovingDuplicates". What does it mean?  Maybe it could be more `eliminateDuplicates(by strategy: Self.StrategyType)`? And the call would be: `eliminateDuplicates(by: .incrementation)`. What do you think?

anna1901 (approved):
Works well, smooth solution. 😎 I added two small suggestions.

AirCasting/Extensions/Array+byRemovingDuplicates.swift
```diff
@@ -0,0 +1,50 @@
+// Created by Lunar on 30/08/2022.
+//
+import Foundation
+
+extension Numeric {
```
pawelgil, line 5: should be in a separate Numeric+Increment (or Decrement) file

AirCasting.xcodeproj/project.pbxproj
```diff
@@ -3471,7 +3471,7 @@
 				ACCEEAB1286EE152005A32B5 /* NotesTextView.swift in Sources */,
 				DDEE3A48269747250098CCA9 /* EditView.swift in Sources */,
 				AC8CFF3625E502DD00FB2918 /* CreateAccountView.swift in Sources */,
-				FAC2672C28BE3F3500D5C13A /* Sequence+MapWithNext.swift in Sources */,
+				FAC2672C28BE3F3500D5C13A /* Array+byRemovingDuplicates.swift in Sources */,
```
pawelgil, line 3474: extensions start with an uppercase letter by convention (Array+ByRemovingDuplicates)

AirCasting/Extensions/Array+byRemovingDuplicates.swift
pawelgil: I agree, the name is not great. But wouldn't use `by` as a label for parameter. Maybe `using`?

AirCasting/Extensions/Array+byRemovingDuplicates.swift
```diff
@@ -0,0 +1,50 @@
...
+}
+
+extension Array where Element: Equatable & Numeric {
+    /// Iterates all elements and gets rid of the duplicates using provided strategy
+    /// - Parameter strategy: a type of strategy which will be used inside `implement` function.
+    /// If this func returns `nil`, the new element won't be added to the resulting array.
+    /// If it returns a value, this value will be added to the resulting array.
+    /// - Returns: A new `Array` with conflicts resolved.
+    
+    enum StrategyType {
```
pawelgil, line 18: I don't think changing the block-based strategy to `enum` is good idea.
If we ever want to change how duplicates are being dealt with in the future (which I can definitely see happening) we're gonna need to go to this file and implement new strategy here in this file.

For example if we ever want to get rid of duplicated streams in a stream array using sensor names we're gonna need to go here, add a new case to this `enum` and append `switch` statement below with the specific streams implementation. It will get messy and coupled.
With block-based strategy we provide a general solution that can be specialized onsite wherever we need it, however we need it, without any modification to this base implementation.
Consider this extension and overall this kind of extensions as a low-level building blocks that provide basic, universal behavior. It can be reused in many places or even across projects without ever changing the implementation. It's like `map` or `filter`.

(also this current approach narrows down the extension to *only* `Numerics` which is a bummer)

I also don't think this makes the callsite particularly easier to read, comparison:
`newValues.byRemovingDuplicates(strategy: .decrementation)`
`newValues.byRemovingDuplicates(strategy: { $0.decrement() })`

pawelgil (changes requested):

AirCasting/Extensions/Array+byRemovingDuplicates.swift
jankrzempek: Got it, thanks! I will change then

AirCasting/Extensions/Array+ByRemovingDuplicates.swift
```diff
@@ -0,0 +1,29 @@
+// Created by Lunar on 30/08/2022.
+//
+import Foundation
+
+extension Array where Element: Equatable & Numeric {
```
pawelgil, line 5: Doesn't need to require `Numeric` anymore

AirCasting/Extensions/Array+IncrementRemovalStrategy.swift
```diff
@@ -0,0 +1,10 @@
+// Created by Lunar on 27/09/2022.
+//
+
+import Foundation
+
+extension Array where Element: Equatable & Numeric {
+    var incrementRemovalStrategy: (Element) -> Element? {
```
pawelgil, line 7: it's not incrementing ;)

pawelgil: btw. it's not used at all anywhere, I'd remove this extension

pawelgil (changes requested):

pawelgil (approved):

PR #895, opened 2022-10-03
Prevents a UI from showing fetching and sync at the same time.

jankrzempek:
**The problem:**

When the user logs in and "Fetching..." is happening, he pulls down to start sync - both spinners show up.

**The solution:**

Do not allow a user to pull down for sync when the first-time "Fetching..." is happening.

### DIFF
--- a/AirCasting/Dashboard/EmptyFixedDashboardView.swift
+++ b/AirCasting/Dashboard/EmptyFixedDashboardView.swift
@@ -12,7 +12,7 @@ struct EmptyFixedDashboardView: View {
     @InjectedObject private var featureFlagsViewModel: FeatureFlagsViewModel
     
     var shouldSessionFetch: Bool {
-        (selectedSection.section == .mobileDormant || selectedSection.section == .fixed) && defaultSessionSynchronizerViewModel.syncInProgress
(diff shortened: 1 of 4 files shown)


pawelgil:
We are currently using 2 separate implementations of visually identical and logically almost identical progress controls in the app. 
image

Not good.

In my opinion we should get to the bottom of this and implement a proper, clean solution instead of layering current problems with even more conditions and imports.
I created a sample PR (#897) with a cleaner solution. It uses a single refresh control instead of two, and cleans up our refreshing logic a bit:
* When cold-launching the app, the `Dashboard` will adjust its `RefreshControl`'s state to reflect the global synchronization state (so currently it will show the refresh control)
* Removed refresh capability from dashboard cards where it doesn't really make sense to have it (Mobile Active and Following since both of those cannot really be refreshed - they depend on local data).
* The only thing that's changed from the users perspective is the text (`Syncing...` vs `Fetching...`). It can be changed, but by adding complexity, it would be best to keep only one label.

Note how much simpler the code got in that PR 

https://user-images.githubusercontent.com/10663361/193658574-f819b574-b91e-4f4d-9515-a3ffe60b693a.mov

PR #905, opened 2022-10-14
Move to mobile dormant after finishing active session

anna1901:
When a mobile active session is finished we should move to mobile dormant tab. However, if there is any other mobile active session, then we should stay in mobile active tab.


AirCasting/Dashboard/DashboardView.swift
```diff
@@ -85,6 +85,13 @@ struct DashboardView: View {
         .onChange(of: selectedSection.section) { newValue in
             try! coreDataHook.setup(selectedSection: newValue)
         }
+        .onChange(of: selectedSection.mobileActiveSessionWasFinished) { newValue in
+            if newValue &&
+                sessions.count == 1 {
```
jankrzempek, line 90: Can it be a one-liner? I think it is easier to read when its size allows us to do that

jankrzempek (approved):
What about the case, when we are recording AB BT session in standalone mode and then `finish and don't sync`? Should we proceed to` mobile dormant`?

jankrzempek (approved):
Always on mobile dormant when nedeed! Nice!

PR #906, opened 2022-10-14
Disabling sync and standalone features.

jankrzempek:
As a result of situations where users were not able to use those features properly - disabling them from now for iPhone SE 1st GEN is the solution.

**What is the scope of this PR:**

- Disable timeout for  BT reconnection feature when using iPhone SE. Instead, show the new card view and wait until AB is reconnected.
- Disable the ability to `finish and sync` when using iPhone SE. Instead, show the card for infinite reconnection


anna1901 (changes requested):
I checked on the iPhone SE 1st gen simulator and SD sync button is present. Does Device.current work also with simulators?

anna1901 (changes requested):
When the AB disconnects from the phone we are automatically putting session in standalone mode. I think this situation should also be handled somehow and we should show some other card than for standalone sessions (because there is this finish and sync button on those cards)

anna1901:
simulator_screenshot_CC9ED068-A4B8-442F-BA89-067AF4456C7E
I this the look we are going for? 😄

anna1901 (changes requested):
I added my device to excluded devices list and I can still see sync button and can conduct sync

AirCasting/SessionViews/ReconnectSessionCardView.swift
```diff
@@ -0,0 +1,101 @@
...
+                .multilineTextAlignment(.center)
+            reconnectionLabel
+            finishAndDontSyncButton
+                .padding()
+        }
+        .alert(item: $alert, content: { $0.makeAlert() })
+        .padding()
+    }
+    
+    var reconnectionLabel: some View {
```
anna1901, line 57: In my opinion this design should be rethinked because blue button suggest that it is tappable and some action will follow while it is not

anna1901 (changes requested):

AirCasting/SessionViews/ReconnectSessionCardView.swift
jankrzempek: This is a design provided to me by Michael (it comes from Android) - and I don't think it is worth the effort to now change this design in both systems. What do you think?

Screenshot_20221017-161911

AirCasting/ABConnector/BluetoothManager.swift
```diff
@@ -211,9 +211,13 @@ extension BluetoothManager: CBCentralManagerDelegate {
         guard mobilePeripheralSessionManager.activeSessionInProgressWith(peripheral) else { return }
         mobilePeripheralSessionManager.markActiveSessionAsDisconnected(peripheral: peripheral)
         connect(to: peripheral)
```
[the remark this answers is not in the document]
jankrzempek: Yes, we are then trying to connect unless we will. Infinite reconnection should be applied to iPhone SE (info in the PR description).

jankrzempek:
> I added my device to excluded devices list and I can still see sync button and can conduct sync

Thanks so much, I forgot that I removed it when doing testing 🫡

AirCasting/SessionViews/ReconnectSessionCardView.swift
jankrzempek: I will ask Michael, never thought about this

AirCasting.xcodeproj/project.pbxproj
```diff
@@ -2724,6 +2736,16 @@
 			path = OldMeasurementRemovers;
 			sourceTree = "<group>";
 		};
+		FA2438F228F996600028B126 /* DeviceKitExtenions */ = {
+			isa = PBXGroup;
+			children = (
+				FA2438F328F996820028B126 /* Features.swift */,
+				FA2438F028F990260028B126 /* DeviceKit+checkDeviceSupportFor.swift */,
+				FA2438F528F996A00028B126 /* Device+excluded.swift */,
```
pawelgil, line 2744: extension names are capitalized by convention:

`Device+excluded.swift -> Device+Excluded.swift`
`DeviceKit+check... -> DeviceKit+Check...`

AirCasting/SessionViews/ReconnectSessionCardView.swift
```diff
@@ -0,0 +1,95 @@
+// Created by Lunar on 21/10/2022.
+//
+
+import SwiftUI
+import AirCastingStyling
+import Resolver
+
+struct ReconnectSessionCardView: View {
+    let session: SessionEntity
```
pawelgil, line 9: do we really need this here? Maybe we can use `Sessionable` instead? We're trying not to use CoreData entities in views anymore

AirCasting/SessionViews/ReconnectSessionCardView.swift
```diff
@@ -0,0 +1,95 @@
...
+        Button(Strings.ReconnectSessionCardView.finishSessionLabel) {
+            alert = InAppAlerts.finishSessionAlert(sessionName: session.name) {
+                self.finishSessionAlertAction()
+            }
+        }
+        .foregroundColor(.accentColor)
+        .font(Fonts.moderateRegularHeading2)
+    }
+    
+    func finishSessionAlertAction() {
```
pawelgil, line 86: can we create a proper ViewModel for this view and move business logic there?

AirCasting/SessionViews/SessionHeaderView.swift
```diff
@@ -289,3 +289,14 @@ private extension SessionHeaderView {
         }
     }
 }
+
+private extension SessionHeaderView {
+    func checkStandaloneActionAvaibility() -> Bool {
+        if session.deviceType == .AIRBEAM3 &&
+            session.isActive && featureFlagsViewModel.enabledFeatures.contains(.standaloneMode) &&
+            checkDeviceSupportFor(feature: .standalone) {
+            return true
```
pawelgil, line 298: we can shorten it from
```
if condition {
  return true
} else {
  return false
}
```
to just:
```
return condition
```

pawelgil (changes requested):
I'd go a different route here.

### Why
This `protocol FeatureProtocol` and `enum Features`  can and will be easily confused with our feature flags. It's adding a lot of complexity.

### Observation
I think this doesn't really differ from our "standard" features controlled by feature flags.

### Idea
Instead of introducing a whole separate mechanism for those new features and spilling the new `DeviceKit` dependency all over the project files maybe it would be cleaner to fit this into our existing feature flagging code? We already have all the machinery at its core including compositional helpers (composite class). I think this design would be good:
current feature flags setup (for `RELEASE` builds):
flags_before

to this we just add a new provider (note how nicely the open/closed principle works here):
flags_new

------
### Implementation
* This new provider should be the most important one - if the feature is not available (`on`) on this device type we don't care if it's enabled on firebase
* There already is a feature flag for standalone mode and sd sync - will it be ok to use those or do we need new ones? (I think we need new ones)
* Make sure we don't leak `DeviceKit` outside of the provider impl
* Wherever needed just `@Inject private var featureFlags: FeatureFlagProvider` and then `featureFlags.isFeatureOn(...)` instead of the current `Device` based code.

EDIT: I know I'm late to the party and dropping this comment, sorry for that @jankrzempek

AirCasting/ABConnector/BluetoothManager.swift
```diff
@@ -49,9 +49,12 @@ class BluetoothManager: NSObject, BluetoothCommunicator, ObservableObject {
     @Published var centralManagerState: CBManagerState = .unknown
     @Published var mobileSessionReconnected = false
     var observed: NSKeyValueObservation?
+    @Published var isReconnectionOn = false
+    @Published var didAlreayCancellConnection = false
```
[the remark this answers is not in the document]
anna1901: I'm not sure about the logic with this variable, because it is used in different functions with the assumption of a specific flow, but okay. I don't like the name of this variable. What do you think about `connectingTimedOut`?

anna1901 (changes requested):
I had a session in standalone mode. Then I excluded standalone feature for iphoneX instead of iphoneSE. I tapped Reconnect. I get that it shouldn't work, but I also didn't get any alert either. The spinner was spinning and then it wasn't.

anna1901 (changes requested):
Additionally, again, I can see the sd sync button even if I exchange iPhone SE to my device in DeviceFeatureFlagsProvider

AirCasting/ABConnector/BluetoothManager.swift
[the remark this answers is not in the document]
jankrzempek: Typo is fixed, thanks. In addition to the name - I think it appropriately says what it does and what is its purpose. If you will force the change of name I will do it, but another way I would like it to stay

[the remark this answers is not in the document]
anna1901: Okay, I will not force anything, but I have to say I don't understand the name. Who already cancelled connection? User? Device?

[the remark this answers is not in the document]
jankrzempek: That's bad, I will come up with something new and place it here so, please let me know then what you think

jankrzempek:
> Additionally, again, I can see the sd sync button even if I exchange iPhone SE to my device in DeviceFeatureFlagsProvider

Checked. After setting `Device.current` to my iPhone, the SD Sync button disappeared nicely. Could tell me more about your case so I can try to reproduce?

AirCasting/ABConnector/BluetoothManager.swift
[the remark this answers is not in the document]
jankrzempek: What would you say for `peripheralConnectionIsCancelled`?

pawelgil (changes requested):

pawelgil (approved):

anna1901 (changes requested):
This doesn't work well. I cannot reconnect with a device. I get an alert that app no longer can connect with a device. I finish session. Then I try to record another session with this AB, I get a connection error, but the previous session that was already finished is revived. I think it is best if you remove the whole reconnecting functionality from this PR and leave just the view (perhaps with the reconnect button commented out) and we will implement the reconnection logic once the bt refactor is done. This implementation is just too buggy for my taste and there is no point merging it to develop branch. But all of the other changes are good, so I think it's best to merge them without the reconnection logic. :)

AirCasting/ABConnector/BluetoothManager.swift
```diff
@@ -210,6 +210,10 @@ extension BluetoothManager: CBCentralManagerDelegate {
 
         guard mobilePeripheralSessionManager.activeSessionInProgressWith(peripheral) else { return }
         mobilePeripheralSessionManager.markActiveSessionAsDisconnected(peripheral: peripheral)
+        connectWithTimeout(using: peripheral)
```
anna1901, line 213: Please remove those changes so that there are no conflicts with BT manager (this class is getting deleted anyway)

AirCasting/SessionViews/ReconnectSessionCardView.swift
```diff
@@ -0,0 +1,88 @@
...
+                .padding()
+        }
+        .alert(item: $viewModel.alert, content: { $0.makeAlert() })
+        .padding()
+    }
+    
+    var reconnectionLabel: some View {
+        Button {
+            if !viewModel.isSpinnerOn {
+                // Custom method of button disable: this can keep the UI style always the same
```
anna1901, line 56: What do you mean by this comment?

AirCasting/SessionViews/ReconnectSessionCardViewModel.swift
```diff
@@ -0,0 +1,43 @@
...
+        }
+    }
+    
+    private func finishSessionAlertAction(completion: () -> Void) {
+        let sessionStopper = Resolver.resolve(SessionStoppable.self, args: session)
+        do {
+            try sessionStopper.stopSession()
+            completion()
+        } catch {
+            Log.info("error when stpoing session - \(error)")
```
anna1901, line 36: Alert here?

AppDelegate+Injection.swift
```diff
@@ -111,23 +111,27 @@ extension Resolver: ResolverRegistering {
...
         main.register { DefaultFeatureFlagProvider() }.scope(.cached)
+        main.register { DeviceFeatureFlagProvider() }.scope(.cached)
 #if !DEBUG
         main.register { FirebaseFeatureFlagProvider() }.scope(.cached)
 #endif
         main.register {
 #if DEBUG
             CompositeFeatureFlagProvider(children: [
                 Resolver.resolve(OverridingFeatureFlagProvider.self),
+//                Resolver.resolve(DeviceFeatureFlagProvider.self),
```
anna1901, line 122: Why is it commented out?

anna1901 (changes requested):
Looks quite good, I just added a few comments. I'm gonna test later, so this CR is based only on looking at the code changes

AirCasting/SessionViews/ReconnectSessionCardView.swift
jankrzempek: That we are not using the `.disable` property instead we have let's say custom disable method, which can keep the blue color of the button all the time.

AirCasting/SessionViews/ReconnectSessionCardViewModel.swift
```diff
@@ -0,0 +1,43 @@
...
+
+import Foundation
+import Resolver
+import Combine
+
+class ReconnectSessionCardViewModel: ObservableObject {
+    let session: SessionEntity
+    @Published var alert: AlertInfo?
+    @Published var isSpinnerOn = false
+    @Injected private var bm: BluetoothManager
```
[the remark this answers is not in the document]
jankrzempek: The idea for this was to stay, so it will be easier to inspire and implement when I will be doing this soon.

AirCasting/SessionViews/ReconnectSessionCardViewModel.swift
jankrzempek: could be, why not 😅

AppDelegate+Injection.swift
jankrzempek: To be sure, it won't work unless the whole feature will be implemented.

anna1901 (approved):
No you are not showing that Reconnection card at all, so I'm not sure what I can test here. Maybe implement also the logic for deciding what session card should be shown and then we can merge it? I'm giving the approve, cause I'm leaving this decision to you

PR #907, opened 2022-10-14
Change standalone icon color in the darkmode

anna1901:
https://trello.com/c/7rCmeiGi

Adding .foregroundColor modifier to Label didn't work. There needs to be this change of setting Image renderingMode to template.


jankrzempek (approved):
Looks good 🥳

PR #910, opened 2022-10-17
Fix colorForCelsius crash

anna1901:
This is a potential fix for this bug:
https://trello.com/c/kk8iJ4Nd


jankrzempek (commented):
Cannot reproduce this bug anymore

PR #911, opened 2022-10-17
New lane for easier build incrementation after release.

jankrzempek:
The idea here was to improve the process of changing the version after release.
For some time we are dealing with a situation when changing the version in one place does not change it in all others. I was trying not to use external tools when doing this lane, but after hours of trying, I think it's reasonable to give it a chance as it is doing its job well.


fastlane/Fastfile
```diff
@@ -151,6 +151,25 @@ platform :ios do
...
+         version_number: new_version
+      )
+      increment_build_number(
+         build_number: "0"
+      )
+      puts "Setting a version number to #{new_version} and reseting build to #{get_build_number}"
+      sh("git checkout -b bumped_version_to_#{new_version}")
+      sh("git add -A")
+      sh("git commit -m \"Change version number to #{new_version} and build reset\"")
+      sh("git push -u origin bumped_version_to_#{new_version}")
```
pawelgil, line 170: I'm on the fence when it comes to pushing from fastlane. Not sure that's a good idea

pawelgil (approved):
great stuff!! 👏👏👏

jankrzempek:
> Looks cool! 👏🏻 Did you test it after the last release?

Not exactly after the release, but I tested it few times, and seems to be working fine

fastlane/Fastfile
jankrzempek: I am going to delete the push and extend the repo description then. Could you tell me why could be not a good idea?

PR #914, opened 2022-10-18
Add timezone offset to request when creating threshold alert

anna1901:
https://trello.com/c/WByJ1gCG

Backend need timezone_offset to calculate whether the mail should be sent or not.


anna1901:
[Add timezone_offset to the request body when creating threshold alert](https://trello.com/c/WByJ1gCG/836-add-timezoneoffset-to-the-request-body-when-creating-threshold-alert)

jankrzempek (approved):

PR #921, opened 2022-10-26
Overfiring timer for statistics container fixed

pawelgil:
### What was done?

Added a mechanism to suspend timers scheduled in `MeasurementsStatisticsController` so that they are not firing when the user is not in map or graph view.


anna1901 (approved):
I added two commits. Works great 👌🏻

AirCasting/SessionViews/SingleMeasurementView.swift
```diff
@@ -52,7 +52,7 @@ struct SingleMeasurementView: View {
                         let formatter = Resolver.resolve(ThresholdFormatter.self, args: threshold)
                         HStack(spacing: 3) {
                             if value != nil {
```
jankrzempek, line 54: maybe we could use `if let value here` so that force unwrap could be deleted?

jankrzempek (approved):

PR #923, opened 2022-10-27
Remove unnecessary error log and sharing session cleanup

anna1901:
### What was done?

Some error logs were removed and some were changed to info logs to void cluttering firebase crashlitics issues


pawelgil:
But are those really non-error situations? Maybe we should address the core of those problems?

pawelgil (approved):

jankrzempek (approved):

PR #929, opened 2022-11-23
Show measurements downloading progress in following tab

anna1901:
https://trello.com/c/JJclGtOq

We want to show progress view when we are downloading measurements for followed session when user enters the app, because sometimes it can take a significant time.

### Additional change
I noticed that in reordering dashboard we are showing a card for deleted session so that got fixed as well

### DIFF
--- a/AirCasting/Dashboard/DashboardView.swift
+++ b/AirCasting/Dashboard/DashboardView.swift
@@ -18,6 +18,7 @@ struct DashboardView: View {
     @EnvironmentObject var reorderButton: ReorderButton
     @EnvironmentObject var searchAndFollowButton: SearchAndFollowButton
     @State var isRefreshing: Bool
+    @Binding var measurementsDownloadingInProgress: Bool
     @State private var alert: AlertInfo?
     @InjectedObject private var userSettings: UserSettings
     @Injected private var networkChecker: NetworkChecker
@@ -28,12 +29,13 @@ struct DashboardView: View {
         coreDataHook.sessions
     }
 
-    init(coreDataHook: CoreDataHook) {
+    init(coreDataHook: CoreDataHook, measurementsDownloadingInProgress: Binding<Bool>) {
         let navBarAppearance = UINavigationBar.appearance()
         navBarAppearance.largeTitleTextAttributes = [.foregroundColor: UIColor(Color.darkBlue)]
         _coreDataHook = StateObject(wrappedValue: coreDataHook)
         self.sessionSynchronizer = Resolver.resolve(SessionSynchronizer.self)
         _isRefreshing = .init(wrappedValue: sessionSynchronizer.syncInProgress.value)
+        _measurementsDownloadingInProgress = .init(projectedValue: measurementsDownloadingInProgress)
     }
 
     var body: some View {
@@ -48,7 +50,7 @@ struct DashboardView: View {
                 sessionTypePicker
                 TabView(selection: $selectedSection.section) {
                     ForEach(DashboardSection.allCases, id: \.self) {
-                        SessionsListView(selectedSection: $0, isRefreshing: $isRefreshing, context: persistenceController.viewContext)
+                        SessionsListView(selectedSection: $0, isRefreshing: $isRefreshing, isDownloading: $measurementsDownloadingInProgress, context: persistenceController.viewContext)
                     }
                 }
                 .tabViewStyle(.page(indexDisplayMode: .never))
--- a/AirCasting/Dashboard/DashboardView.swift
+++ b/AirCasting/Dashboard/DashboardView.swift
@@ -36,6 +36,8 @@ struct DashboardView: View {
         self.sessionSynchronizer = Resolver.resolve(SessionSynchronizer.self)
         _isRefreshing = .init(wrappedValue: sessionSynchronizer.syncInProgress.value)
         _measurementsDownloadingInProgress = .init(projectedValue: measurementsDownloadingInProgress)
+        var log = isRefreshing
+        Log.info("##### IS REFRESHING: \(log)")
     }
 
     var body: some View {
@@ -50,7 +52,7 @@ struct DashboardView: View {
                 sessionTypePicker
                 TabView(selection: $selectedSection.section) {
                     ForEach(DashboardSection.allCases, id: \.self) {
-                        SessionsListView(selectedSection: $0, isRefreshing: $isRefreshing, isDownloading: $measurementsDownloadingInProgress, context: persistenceController.viewContext)
+                        SessionsListView(selectedSection: $0, isRefreshing: $isRefreshing, measurementsDownloadingInProgress: $measurementsDownloadingInProgress, context: persistenceController.viewContext)
                     }
                 }
                 .tabViewStyle(.page(indexDisplayMode: .never))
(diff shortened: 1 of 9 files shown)


AirCasting/Dashboard/Reordering/ReorderingDashboard.swift
```diff
@@ -18,7 +18,7 @@ struct ReorderingDashboard: View {
     var body: some View {
         ScrollView {
             LazyVGrid(columns: columns) {
-                ForEach(viewModel.sessions, id: \.uuid) { session in
+                ForEach(viewModel.sessions.filter { $0.uuid != "" && !$0.gotDeleted }, id: \.uuid) { session in
```
pawelgil, line 21: this should be performed on a viewmodel level

AirCasting/Dashboard/SessionListView.swift
```diff
@@ -14,14 +14,16 @@ struct SessionsListView: View {
...
     @StateObject private var coreDataHook: CoreDataHook
     @Binding private var isRefreshing: Bool
+    @Binding private var isDownloading: Bool
     @InjectedObject private var featureFlagsViewModel: FeatureFlagsViewModel
     
     private let listCoordinateSpaceName = "listCoordinateSpace"
     private let selectedSection: DashboardSection
 
-    init(selectedSection: DashboardSection, isRefreshing: Binding<Bool>, context: NSManagedObjectContext) {
+    init(selectedSection: DashboardSection, isRefreshing: Binding<Bool>, isDownloading: Binding<Bool>, context: NSManagedObjectContext) {
```
pawelgil, line 23: this can be a bit confusing. Without a good project understanding it will be hard to tell what's different between `isRefreshing` and `isDownloading`. Maybe `isDownloadingMeasurements`? (same for the property name)

AirCasting/Dashboard/SessionListView.swift
```diff
@@ -44,6 +46,10 @@ struct SessionsListView: View {
                 if selectedSection.allowsRefreshing {
                     RefreshControl(coordinateSpace: .named(listCoordinateSpaceName), isRefreshing: $isRefreshing)
                 }
+                if selectedSection == .following && isDownloading {
```
pawelgil, line 49: maybe we can do the same as with refresh control - move the `selectedSection == .following` to a computed var like `selectedSection.shouldShowMeasurementDownloadProgress` or smth like that? This way we'll have all of those conditions in a single place

AirCasting/MainTabBarView.swift
```diff
@@ -54,14 +55,17 @@ struct MainTabBarView: View {
             
         }
         .onAppCameToForeground {
-            measurementUpdatingService.updateAllSessionsMeasurements()
+            measurementsDownloadingInProgress = true
+            measurementUpdatingService.updateAllSessionsMeasurements() { measurementsDownloadingInProgress = false }
```
pawelgil, line 59: doesn't Xcode warn about this state being set on a non-main thread? Because the completion here is called from a background queue

AirCasting/MainTabBarView.swift
```diff
@@ -54,14 +55,17 @@ struct MainTabBarView: View {
...
         }
         .onAppear {
             UITabBar.appearance().backgroundColor = .aircastingBackground
             let appearance = UITabBarAppearance()
             appearance.backgroundImage = UIImage()
             appearance.shadowImage = UIImage.mainTabBarShadow
             UITabBar.appearance().standardAppearance = appearance
+            measurementsDownloadingInProgress = true
+            measurementUpdatingService.updateAllSessionsMeasurements() { measurementsDownloadingInProgress = false }
```
pawelgil, line 68: doesn't Xcode warn about this state being set on a non-main thread? Because the completion here is called from a background queue

AirCasting/Services/DownloadMeasurmentsService.swift
```diff
@@ -118,6 +127,7 @@ final class DownloadMeasurementsService: MeasurementUpdatingService {
                     try UpdateSessionParamsService().updateSessionsParams(session: session, output: output)
```
pawelgil, line 127: In this case I think we can place a `defer { completion() }` right after `do {` and remove all other completion calls

pawelgil (changes requested):

AirCasting/Services/DownloadMeasurmentsService.swift
anna1901: How about those in `catch` blocks?

anna1901: Okay, I know already ;)

AirCasting/MainTabBarView.swift
anna1901: Interestingly, it doesn't 🤔

anna1901: And what's more, when I add this dispatch, cards are not updated after measurements get downloaded. I have to toggle card expanding to see new time on the card.

anna1901: Or maybe not. Very misterious. :D

AirCasting/Dashboard/DashboardView.swift
```diff
@@ -28,12 +29,15 @@ struct DashboardView: View {
...
-    init(coreDataHook: CoreDataHook) {
+    init(coreDataHook: CoreDataHook, measurementsDownloadingInProgress: Binding<Bool>) {
         let navBarAppearance = UINavigationBar.appearance()
         navBarAppearance.largeTitleTextAttributes = [.foregroundColor: UIColor(Color.darkBlue)]
         _coreDataHook = StateObject(wrappedValue: coreDataHook)
         self.sessionSynchronizer = Resolver.resolve(SessionSynchronizer.self)
         _isRefreshing = .init(wrappedValue: sessionSynchronizer.syncInProgress.value)
+        _measurementsDownloadingInProgress = .init(projectedValue: measurementsDownloadingInProgress)
+        var log = isRefreshing
+        Log.info("##### IS REFRESHING: \(log)")
```
jankrzempek, line 40: Should we keep that log? If so can we remove those # characters?

AirCasting/MainTabBarView.swift
```diff
@@ -54,14 +55,25 @@ struct MainTabBarView: View {
             
         }
         .onAppCameToForeground {
-            measurementUpdatingService.updateAllSessionsMeasurements()
+            measurementsDownloadingInProgress = true
+            measurementUpdatingService.updateAllSessionsMeasurements() {
+                DispatchQueue.main.async {
+                    measurementsDownloadingInProgress = false
+                }
```
jankrzempek, line 62: I think we don't need  `DispatchQueue.main.async` as long as this code is View related

AirCasting/MainTabBarView.swift
```diff
@@ -54,14 +55,25 @@ struct MainTabBarView: View {
...
             UITabBar.appearance().backgroundColor = .aircastingBackground
             let appearance = UITabBarAppearance()
             appearance.backgroundImage = UIImage()
             appearance.shadowImage = UIImage.mainTabBarShadow
             UITabBar.appearance().standardAppearance = appearance
+            measurementsDownloadingInProgress = true
+            measurementUpdatingService.updateAllSessionsMeasurements() {
+                DispatchQueue.main.async {
+                    measurementsDownloadingInProgress = false
+                }
```
jankrzempek, line 75: I think we don't need  `DispatchQueue.main.async` as long as this code is View related

jankrzempek (commented):
Working well, but from the user perspective, I think it seems a bit odd. When I have 2/3 sessions on my account and enter the app, the spinner is so fast that I am not able to indicate what it was and why is this happening. Maybe we can guarantee that this spinner will always last at least 1 sec?

AirCasting/MainTabBarView.swift
anna1901: Could you elaborate? Cause I thought we need it specifically because it is view related 🤔

AirCasting/Dashboard/DashboardView.swift
anna1901: Sorry, this was left by mistake

PR #930, opened 2022-11-24
Reconnect disconnected session

anna1901:
## What was done?
When standalone mode feature is disabled (eg. for iphone SE) we should allow user to reconnect with AirBeam

## Additional changes
This PR fixes a bug introduced in bt manager refactor which caused measurements for new active sessions be saved for session that was previously finished additionally to being saved for active session.

https://trello.com/c/3ymr4roL


AirCasting/ABConnector/AirBeam3Configurator.swift
```diff
@@ -11,7 +11,7 @@ import Resolver
 
 protocol AirBeamConfigurator {
     func configureMobileSession(location: CLLocationCoordinate2D, completion: @escaping (Result<Void, Error>) -> Void)
-    func configureFixed(uuid: SessionUUID, completion: @escaping (Result<Void, Error>) -> Void)
+    func configureSession(uuid: SessionUUID, completion: @escaping (Result<Void, Error>) -> Void)
     func configureFixedCellularSession(uuid: SessionUUID,
```
pawelgil, line 15: side note: this protocol is also a good candidate for a small refactor so it uses proper polymorphism - we'd have a single method here like `configureSession(completion:)` and then different implementations for fixed/mobile/cellular/whatever. This way we'd make one step towards unification of the session creation process

AirCasting/BluetoothSession/BluetoothSessionRecordingController.swift
```diff
@@ -56,33 +57,39 @@ class MobileAirBeamSessionRecordingController: BluetoothSessionRecordingControll
         Resolver.resolve(AirBeamConfigurator.self, args: device)
             .configureMobileSession(location: locationTracker.location.value?.coordinate ?? .undefined,
                                     completion: completion)
+        
         // Info: we're not changing the sessions `status` property here to `.RECORDING` because it is currently
         // being done by the MeasurementStreamStorage class automagically.
         // This is something we might want to change at some point.
```
pawelgil, line 63: is this actionable?

AirCasting/BluetoothSession/BluetoothSessionRecordingController.swift
```diff
@@ -56,33 +57,39 @@ class MobileAirBeamSessionRecordingController: BluetoothSessionRecordingControll
...
+        
+        guard !isRecording else { return } // To make sure we are not recording more than one session at once
```
pawelgil, line 65: if this scenario (a call to `startRecording`) is unexpected and shouldn't happen maybe we should put an assertion here so that we get a feedback while coding in the future?

AirCasting/BluetoothSession/BluetoothSessionRecordingController.swift
```diff
@@ -56,33 +57,39 @@ class MobileAirBeamSessionRecordingController: BluetoothSessionRecordingControll
...
     }
     
-    func stopRecordingSession(with uuid: SessionUUID) {
-        // Database change should be performed for active and disconnected sessions
-        performDatabaseChange(for: uuid)
+    func stopRecordingSession(with uuid: SessionUUID, databaseChange: (MobileSessionStorage) -> Void) {
+        // Database change should be performed for both active and disconnected sessions
+        databaseChange(storage)
         
         guard let activeSession = activeSessionProvider.activeSession, activeSession.session.uuid == uuid else { return }
```
pawelgil, line 77: `isRecording` won't set to `false` if this `guard` fails. I think it might be a bug. Also - should we put an assert here to make sure noone calls stop without a start first?

AirCasting/BluetoothSession/ReconnectionController.swift
```diff
@@ -14,7 +14,7 @@ protocol ReconnectionController {
     var delegate: ReconnectionControllerDelegate? { get set }
 }
 
-class DefaultReconnectionController: ReconnectionController, BluetoothConnectionObserver {
+class DefaultReconnectionObserver: ReconnectionController, BluetoothConnectionObserver {
```
pawelgil, line 17: I'll leave it to you, but I'm not sure about this new name. Usually the likes of `Observer` or `Delegate` protocols are secondary characteristics of a class, not their main responsibility. In other words - this classes main purpose is to be a `ReconnectionController` and the `BluetoothConnectionObserver` conformance is accidental - it needs to be one to do the job, but it's a detail.

AirCasting/BluetoothSession/UserTriggeredReconnectionController.swift
```diff
@@ -0,0 +1,67 @@
+// Created by Lunar on 24/11/2022.
+//
+
+import Foundation
+import Resolver
+
+enum UserTroggeredReconnectionError: Error {
```
pawelgil, line 7: typo (Troggered -> Triggered)

AirCasting/BluetoothSession/UserTriggeredReconnectionController.swift
```diff
@@ -0,0 +1,67 @@
...
+}
+
+class DefaultUserTriggeredReconnectionController: UserTriggeredReconnectionController {
+    @Injected private var btScanner: BluetoothScanner
+    @Injected private var bluetootConnector: BluetoothConnectionHandler
+    @Injected private var activeSessionProvider: ActiveMobileSessionProvidingService
+    @Injected private var sessionRecorder: BluetoothSessionRecordingController
+    
+    func reconnectWithPeripheral(deviceUUID: String, session: Session, completion: @escaping (Result<Void, UserTroggeredReconnectionError>) -> Void) {
+        guard activeSessionProvider.activeSession == nil else {
```
pawelgil, line 26: big function, could use some breaking up into smaller private ones

AirCasting/BluetoothSession/UserTriggeredReconnectionController.swift
```diff
@@ -0,0 +1,67 @@
...
+            Log.info("Tried to resume session when there was another active session in progress")
+            completion(.failure(UserTroggeredReconnectionError.anotherActiveSessionInProgress))
+            return
+        }
+        
+        var discoveredDevices: [NewBluetoothManager.BluetoothDevice] = []
+        btScanner.startScanning(scanningWindow: 10, onDeviceDiscovered: { discoveredDevices.append($0) }, onScanningFinished: {
```
pawelgil, line 33: I feel like this can be significantly improved: why do we wait until the scanning is finished here? We can use `onDeviceDiscovered` to check if that's "our" device and do stuff there (stop scanning and reconnect). We will probably go down to 1-2 sec instead od 10. Or am I missing something here?

AirCasting/BluetoothSession/UserTriggeredReconnectionController.swift
```diff
@@ -0,0 +1,67 @@
...
+
+import Foundation
+import Resolver
+
+enum UserTroggeredReconnectionError: Error {
+    case anotherActiveSessionInProgress
+    case failedToDiscoverCharacteristics
+    case failedToConnect
+    case deviceNotDiscovered
+    case airbeamConfigurationFailure
```
pawelgil, line 12: I think we don't really need that detailed cases here. In my opinion we can just use 3: `activeSessionInProgress` + `failedToConnect` + `deviceUnavailable`. Generally each layer usually boils down the errors to match its domain. Here we're pretty high up, so we don't really care if reconnection failed because BT failed to discover characteristics or because the configuration failed. Those are both connection errors and can be merged at this point (at the `Controller` level the errors should be thought of as user-facing ones. We might not display them ofc, it's just my own rule of thumb)

AirCasting/SessionViews/ReconnectSessionCardViewModel.swift
```diff
@@ -26,6 +44,11 @@ class ReconnectSessionCardViewModel: ObservableObject {
         }
     }
     
+    // This is just in case something goes wrong and the viewModel stays in memory
```
pawelgil, line 47: I still don't get why this would happen 🤷‍♂️

pawelgil (changes requested):

AirCasting/SessionViews/ReconnectSessionCardViewModel.swift
jankrzempek: this method will be triggered every time the view refreshes or the tab is changed to another and back again. Do we need it at all?

jankrzempek (commented):
Working very well! 👏

AirCasting/BluetoothSession/BluetoothSessionRecordingController.swift
anna1901: I think this can possibly be done during MeasurementStramStorage refator

AirCasting/BluetoothSession/BluetoothSessionRecordingController.swift
anna1901: It's not a bug. `isRecording` should already be set to false cause it means that `stopRecordingSession` was called for disconnected session. (So maybe it is a bit misleading, but I would keep it like that for now cause I don't have a better idea how to do it.)

AirCasting/BluetoothSession/UserTriggeredReconnectionController.swift
anna1901: Yeah, good catch, thanks

AirCasting/SessionViews/ReconnectSessionCardViewModel.swift
anna1901: True, okay, I'll remove it

AirCasting/Utils/Strings.swift
```diff
@@ -826,8 +826,20 @@ struct Strings {
                                                    comment: "")
         static let reconnectLabel = NSLocalizedString("Reconnect",
                                                                 comment: "")
+        static let connectingLabel = NSLocalizedString("Connecting...",
+                                                                comment: "")
+        static let connectedLabel = NSLocalizedString("Connected...",
```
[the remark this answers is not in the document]
anna1901: What do you think about "Finalizing..."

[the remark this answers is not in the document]
jankrzempek: Awesome

AirCasting/BluetoothSession/UserTriggeredReconnectionController.swift
anna1901: I'm not able to implement it so I would leave it as is since this won't be used by many user and I don't want to spend too much time on it

AirCasting/BluetoothSession/UserTriggeredReconnectionController.swift
```diff
@@ -0,0 +1,81 @@
...
+                case .success():
+                    self?.resume(session, device: device, completion: completion)
+                case .failure(let error):
+                    completion(.failure(error))
+                }
+            }
+        })
+    }
+    
+    func connect(to device: NewBluetoothManager.BluetoothDevice, session: Session, completion: @escaping (Result<Void, UserTriggeredReconnectionError>) -> Void) {
```
pawelgil, line 48: private

AirCasting/BluetoothSession/UserTriggeredReconnectionController.swift
```diff
@@ -0,0 +1,81 @@
...
+                        completion(.failure(.failedToConnect))
+                    }
+                }
+            case .failure(_):
+                completion(.failure(.failedToConnect))
+            }
+        }
+    }
+    
+    func resume(_ session: Session, device: NewBluetoothManager.BluetoothDevice, completion: @escaping (Result<Void, UserTriggeredReconnectionError>) -> Void) {
```
pawelgil, line 69: private

pawelgil (approved):

PR #934, opened 2022-11-29
Show refresh control when sync was trigerred from code.

jankrzempek:
**Goal:**
Now, when user logs in to the app, there is no indication of sync from the UI perspective.
Goal was to add this spinner, and cover possible future cases when sync could be triggered from code.

### DIFF
--- a/AirCasting/Dashboard/DashboardView.swift
+++ b/AirCasting/Dashboard/DashboardView.swift
@@ -28,6 +28,10 @@ struct DashboardView: View {
     private var sessions: [Sessionable] {
         coreDataHook.sessions
     }
+    
+    private var noDormantNorFixedSessions: Bool {
+        !sessions.contains(where: { $0.isFixed || !$0.isActive })
+    }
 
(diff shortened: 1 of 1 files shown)


anna1901 (commented):
Hmmm... Is it good to show the spinner every time the sync is triggered from the code? I think not necessarily, cause we are triggering sync quite often. Eg. after session is stopped we are using sync to send it to backend and I don't think it is needed to show the spinner. Can we do it eg. only when there are no sessions or somehow just show it after user loggs in?

jankrzempek:
> Hmmm... Is it good to show the spinner every time the sync is triggered from the code? I think not necessarily, cause we are triggering sync quite often. Eg. after session is stopped we are using sync to send it to the backend and I don't think it is needed to show the spinner. Can we do it eg. only when there are no sessions or somehow just show it after user loggs in?

At first, I thought it was a good idea, but maybe you are right.
I am more into checking if there are any sessions (dormant or fixed) or not. As long as a user hasn't collected those, I will keep the refresh control active.

PR #938, opened 2022-12-02
Fix end time updating for active session

anna1901:
# What was the problem?
Turn out active session end time was updating only because the whole session card was updating when statistics were calculated, which was happening even if user was not in map/graph view. When we fixed excessive statistics updating, session end time stopped updating.

# What was done?
I added new DynamicSessionEndTime view for active sessions which observes stream and shows end time based on the last measurement time. It has to be done like that because before the session is finished it's end time is nil.

https://trello.com/c/t3GQHAVF

### DIFF
--- a/AirCasting/SessionViews/SessionTime.swift
+++ b/AirCasting/SessionViews/SessionTime.swift
@@ -1,0 +1,48 @@
+// Created by Lunar on 02/12/2022.
+//
+
+import Foundation
+import SwiftUI
+
+struct SessionTime: View {
+    @ObservedObject var session: SessionEntity
+    
+    private var sessionStreams: [MeasurementStreamEntity] {
+        return session.sortedStreams
+    }
+    
+    var body: some View {
+        if session.isActive, let stream = sessionStreams.first {
+            DynamicSessionTime(session: session, stream: stream)
+        } else {
+            staticTimeAndDate()
+        }
+    }
+    
+    func staticTimeAndDate() -> Text {
+        let formatter: DateIntervalFormatter = DateFormatters.SessionCardView.shared.utcDateIntervalFormatter
+        guard let start = session.startTime else { return Text("") }
+        let end = session.endTime ?? DateBuilder.getFakeUTCDate()
+        
+        let string = formatter.string(from: start, to: end)
+        return Text(string)
+    }
+}
+
+struct DynamicSessionTime: View {
+    @ObservedObject var session: SessionEntity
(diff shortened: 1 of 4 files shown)


AirCasting/SessionViews/SessionTime.swift
```diff
@@ -0,0 +1,48 @@
...
+        let formatter: DateIntervalFormatter = DateFormatters.SessionCardView.shared.utcDateIntervalFormatter
+        guard let start = session.startTime else { return Text("") }
+        let end = session.endTime ?? DateBuilder.getFakeUTCDate()
+        
+        let string = formatter.string(from: start, to: end)
+        return Text(string)
+    }
+}
+
+struct DynamicSessionTime: View {
```
jankrzempek, line 32: This struct can be mark as private and moved into `SessionTime`

AirCasting/SessionViews/SessionTime.swift
```diff
@@ -0,0 +1,48 @@
...
+        Text(formatedTimeAndDate())
+    }
+    
+    func formatedTimeAndDate() -> String {
+        let formatter: DateIntervalFormatter = DateFormatters.SessionCardView.shared.utcDateIntervalFormatter
+        guard let start = session.startTime else { return "" }
+        let end = session.endTime ?? stream.lastMeasurementTime ?? DateBuilder.getFakeUTCDate()
+        
+        let string = formatter.string(from: start, to: end)
+        return string
```
jankrzempek, line 46: Maybe here it could return Text as `staticTimeAndDate` func does? Or the other way? just to make those to consistent?

jankrzempek (approved):
👏

AirCasting/SessionViews/SessionTime.swift
```diff
@@ -0,0 +1,48 @@
+// Created by Lunar on 02/12/2022.
+//
+
+import Foundation
+import SwiftUI
+
+struct SessionTime: View {
```
pawelgil, line 7: I'd put a `View` suffix here per convention to make it a `SessionTimeView`

AirCasting/SessionViews/SessionTime.swift
```diff
@@ -0,0 +1,48 @@
...
+    @ObservedObject var session: SessionEntity
```
pawelgil, line 8: a simple ViewModel would be nice to handle this text formatting (but not really required since this file is just 50LOC)

pawelgil (approved):

PR #940, opened 2022-12-05
Logger deadlock fixed

jankrzempek:
Taken from GoogleMaps refractor.
Thanks Paweł!

### DIFF
--- a/AppDelegate+Injection.swift
+++ b/AppDelegate+Injection.swift
@@ -8,68 +8,11 @@ import DeviceKit
 import Firebase
 
 extension Resolver: ResolverRegistering {
-    public static let fileLoggerQueue = DispatchQueue(label: "com.habitatmap.filelogger", qos: .utility, attributes: [], autoreleaseFrequency: .workItem, target: nil)
     public static func registerAllServices() {
         // We do Firebase config here as this is actually the first place that gets called in the app.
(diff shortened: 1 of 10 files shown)


AppDelegate+Injection.swift
```diff
@@ -8,68 +8,11 @@ import DeviceKit
...
-        main.register { (_, _) -> GarbageCollector in
-            let collector = GarbageCollector()
-            let logsHolder = FileLoggerDisposer(disposeQueue: fileLoggerQueue)
-            collector.addHolder(logsHolder)
-            return collector
-        }.scope(.application)
-
-        // MARK: Persistence
+        
+        // TODO: Reintroduce garbage collection
```
anna1901, line 15: Why is garbage collection removed and when should it be reintroduced?

anna1901 (commented):
Please think about someone who will look at this code without knowing the context. I think the TODO comments are not really self explanatory. Could you add more information to them?

AppDelegate+Injection.swift
pawelgil: it shouldn't. We can remove the TODO and forget about this. It was an overkill since the logs are not a big file so the system won't really care for this ~20KB unfreed space

pawelgil (approved):

PR #946, opened 2022-12-15
Fix heatmap - wrong colors

jankrzempek:
https://trello.com/c/7FeVqntE/861-mobile-active-heatmap-colors-inaccurate
When seeing a heatMap, it tends to change immediately out of nowhere. Please use `Path` session which can be found on mobile dormant section in Michael's account to test.

To test: Use the `RH` param and zoom in to see a path, and it changes a lot. Now switch to Pm.2.5 and RH again. You should see that Pm2.5 overlaps RH, even when RH is chosen.


anna1901 (changes requested):
IMG_0289
Unfortunately heatmap is not updating when I change thresholds values with the slider

anna1901 (approved):
Works well

PR #965, opened 2024-02-22
Change context

smialko:
https://trello.com/c/ALD7OISt/891-mobile-standalone-sessions-wont-sync-to-backend

**Bug**: the session wasn't uploaded to the backend and didn't show up on habitatmap map
**Problem**: Server didn't accept the synced session because it didn't have the end_time. That happened because we've been saving synced data on the `edit context` and fetching it from `create context` which wasn't updated with the endtime :)
**Solution**:  changed the context


jankrzempek (approved):
I tested what I was able to, please go with it!

PR #977, opened 2024-03-07
Fix error given by new XCode 15.13

jankrzempek:


smialko (approved):
Great job but I can't test it until I update code so I believe that is works fine and give you an accept :D

PR #979, opened 2024-03-14
Condition for graph update and averaging threshold changed

jankrzempek:
https://trello.com/c/5XiL8wQO/873-following-graph-update-graph-in-real-time

Currently, after 24 hours of collecting measurements, we always have the same number of total measurements. The reason for this is that when we collect new measurements, we always delete the last ones. This makes the total count always the same. To continue preventing all the time graph updates, we are introducing date checking of the lastest measurement.

The next thing was about time correctness on the graph and the session itself. There was some discrepancy, becouse of our averaging service and the fact that the time is taken from x point of the latest point (which was previously averaged). To prevent such a scenario we are increasing the threshold to 1500, as the max value for measurements should not exceed 1440 for the fixed session.

### DIFF
--- a/AirCasting/Graph/Graph.swift
+++ b/AirCasting/Graph/Graph.swift
@@ -39,7 +39,7 @@ struct Graph: UIViewRepresentable {
     private var notesHandler: NotesHandler?
     
     var isAutozoomEnabled: Bool
-    let simplifiedGraphEntryThreshold = 1000
+    let simplifiedGraphEntryThreshold = 1500
     private let formatter: ThresholdFormatter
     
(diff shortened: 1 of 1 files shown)


smialko (approved):
Code lookin' good, although I don't have sessions long enough to test it well enough.  
PS WHAT'S WITH EMOJIS HERE?! WHY THEY'RE NOT APPEARING

jankrzempek:
> Code lookin' good, although I don't have sessions long enough to test it well enough. PS WHAT'S WITH EMOJIS HERE?! WHY THEY'RE NOT APPEARING

😎 For example this one? For me they are working well 😅

PR #985, opened 2024-04-02
Measurements from AB mini should be recorded every second

smialko:
Long story short: Before, we used AB3 to record 5 streams together, waiting for all 5 measurements before timestamping. But with ABMini's 2 streams, timing got a bit off. Don't worry, it's fixed now! ;D

### DIFF
--- a/AirCasting/BluetoothSession/MeasurementsSavingService.swift
+++ b/AirCasting/BluetoothSession/MeasurementsSavingService.swift
@@ -6,7 +6,7 @@ import Resolver
 import CoreLocation
 
 protocol MeasurementsSavingService {
-    func handlePeripheralMeasurement(_ measurement: ABMeasurementStream, sessionUUID: SessionUUID, locationless: Bool, device: DeviceType?)
+    func handlePeripheralMeasurement(_ measurement: ABMeasurementStream, sessionUUID: SessionUUID, locationless: Bool)
     func createSession(session: Session, device: any BluetoothDevice, completion: @escaping (Result<Void, Error>) -> Void)
     func changeStatusToRecording(for sessionUUID: SessionUUID)
 }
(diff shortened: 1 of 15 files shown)


AirCasting/BluetoothSession/MeasurementsSavingService.swift
```diff
@@ -55,9 +55,21 @@ class DefaultMeasurementsSaver: MeasurementsSavingService {
...
+        let valuesCount: Int
+        switch device {
+        case .AIRBEAM3:
+            valuesCount = 5
+        case .AIRBEAMMINI:
+            valuesCount = 2
+        default:
+            valuesCount = 1
+            Log.warning("There's an unknown device recording session")
+        }
```
jankrzempek, line 68: Handling this here, means we need to execute this whenever this function is called - which is often. Please execute this once and save for future. What about:

 ```
private var totalDeviceValues = 1
 func createSession(session: Session, device: any BluetoothDevice, completion: @escaping (Result<Void, Error>) -> Void) {
        persistence.accessStorage { [weak self] storage in
            do {
                guard let self else { return }
                let sessionReturned = try storage.createSession(session)
                let entity = BluetoothConnectionEntity(context: sessionReturned.managedObjectContext!)
                entity.peripheralUUID = device.uuid
                entity.session = sessionReturned
                self.uiStorage.accessStorage { storage in
                    do {
                        try storage.switchCardExpanded(to: true, sessionUUID: session.uuid)
                    } catch {
                        Log.error("\(error)")
                    }
                }
                checkForDeviceValues(basedOn: entity.session?.deviceType)
                completion(.success(()))
            } catch {
                Log.info("\(error)")
                completion(.failure(error))
            }
        }
    }
    
    func checkForDeviceValues(basedOn type: DeviceType?) {
        /* Explanation: We anticipate receiving 5 measurements from AirBeam 3 (AB3) and 2 from AirBeam Mini (AB2). It's crucial that these batches arrive with timestamps accurate to the second. Therefore, we wait for the expected number of measurements before updating streams with the current time. */
        Log.debug("Total values: \(self.totalDeviceValues)")
        switch type {
        case .AIRBEAM3:
            totalDeviceValues = 5
        case .AIRBEAMMINI:
            totalDeviceValues = 2
        default:
            totalDeviceValues = 1
            Log.warning("There's an unknown device recording session")
        }
    }

    func handlePeripheralMeasurement(_ measurement: ABMeasurementStream, sessionUUID: SessionUUID, locationless: Bool) {
        Log.debug("Total values: \(self.totalDeviceValues)")
        if peripheralMeasurementManager.collectedValuesCount == totalDeviceValues {
            peripheralMeasurementManager.startNewValuesRound(locationless: locationless)
        }
        updateStreams(stream: measurement, sessionUUID: sessionUUID, location: peripheralMeasurementManager.currentLocation, time: peripheralMeasurementManager.currentTime)
        peripheralMeasurementManager.incrementCounter()
    }

```

jankrzempek: few things only changed in here, posting part of the code to show the whole concept

jankrzempek (changes requested):
Seems to be working for AB3 well! Please see one thing in there we might improve

AirCasting/BluetoothSession/MeasurementsSavingService.swift
smialko: gooooooood :D

jankrzempek (approved):
😎💥

PR #992, opened 2024-04-09
Fix recording mobile session with ABMini

smialko:


AirCasting/BluetoothSession/MeasurementsSavingService.swift
```diff
@@ -14,15 +14,21 @@ protocol MeasurementsSavingService {
 class DefaultMeasurementsSaver: MeasurementsSavingService {
     @Injected private var persistence: MobileSessionRecordingStorage
     @Injected private var uiStorage: UIStorage
-    private var peripheralMeasurementManager = PeripheralMeasurementTimeLocationManager()
+    private var peripheralMeasurementManager: PeripheralMeasurementTimeLocationManager?
```
mpanuszewska, line 17: Does this saver make sense without this manager? If not having this manger is not a valid state, it's a good practice to avoid optionals.

AirCasting/BluetoothSession/MeasurementsSavingService.swift
```diff
@@ -14,15 +14,21 @@ protocol MeasurementsSavingService {
...
 
     class PeripheralMeasurementTimeLocationManager {
         @Injected private var locationTracker: LocationTracker
 
-        private(set) var collectedMeasurementsCount: Int = 0
-        private(set) var currentTime: Date = DateBuilder.getFakeUTCDate()
-        private(set) var currentLocation: CLLocationCoordinate2D? = .undefined
+        private(set) var collectedMeasurementsCount: Int
+        private(set) var currentTime: Date
+        private(set) var currentLocation: CLLocationCoordinate2D?
```
mpanuszewska, line 25: Same here, if we have .undefined as a possible state then I'm assuming currentLocation should always be known.

AirCasting/BluetoothSession/MeasurementsSavingService.swift
mpanuszewska: It seems to me looking at the rest of the code that at least some functionality doesn't make sense without it. 
If it's only that functionality, maybe we can ensure that this manager is created before running "handlePeripheralMeasurement"? 
If it's the whole class, can we ensure that it is created in the init?

AirCasting/CoreData/Session.swift
```diff
@@ -135,16 +135,17 @@ enum StreamingMethod: Int {
...
 
     public var description: String {
         switch self {
         case .MIC: return "Device's Microphone"
-        case .AIRBEAM3: return "AirBeam 3"
-        case .AIRBEAMMINI: return "AirBeamMini"
+        case .AIRBEAM: return "AirBeam"
         }
     }
+    func isAirBeam() -> Bool {
```
mpanuszewska: This seems a little redundant, I don't think I'm getting what is the upside of calling 
`deviceType.isAirBeam() 
`
over doing 
`deviceType == .AIRBEAM 
`

AirCasting/CreateSessionViews/SelectPeripheralView/SelectPeripheralView.swift
```diff
@@ -69,6 +69,8 @@ struct SelectPeripheralView: View {
             Button(action: {
                 selection = availableDevice
                 sessionContext.device = availableDevice
+                // TODO: AB shouldn't be hardcoded
```
mpanuszewska, line 72: TODO? ^_^

mpanuszewska (commented):
I didn't really check any logic, I have to dig into the code a little more before I know what's going on. 

I left a couple of notes strictly about the code though. Nothing too concerning. 

The only thing that worries is is that there are a lot of TODOs in the code, do we have any tickets to cover them later on?

AirCasting/CoreData/Session.swift
smialko: True :D it was created before for different thing and we should have remove it :D

AirCasting/CreateSessionViews/SelectPeripheralView/SelectPeripheralView.swift
smialko: ;D yes, I know. 
But, we've talked about this with Janek, and the question is - can we record sessions with devices other than AB? Should we care?

PR #993, opened 2024-04-12
Geolocalization

smialko:
Task: https://trello.com/c/sYeQkzDd/887-add-geolocation-from-last-measurement-of-the-session-from-local-db

Note: we spliced this ticket into two, here you'll find just adding geolocalization when we already have measurement with localization in local database.

### DIFF
--- a/AirCasting/CoreData/Session.swift
+++ b/AirCasting/CoreData/Session.swift
@@ -143,9 +143,6 @@ public enum DeviceType: Int, CustomStringConvertible {
         case .AIRBEAM: return "AirBeam"
         }
     }
-    func isAirBeam() -> Bool {
-        self == .AIRBEAM
-    }
 }
 
 extension Session {
--- a/AirCasting/CoreData/Session.swift
+++ b/AirCasting/CoreData/Session.swift
@@ -143,6 +143,9 @@ public enum DeviceType: Int, CustomStringConvertible {
         case .AIRBEAM: return "AirBeam"
         }
     }
+    func isAirBeam() -> Bool {
+        self == .AIRBEAM
+    }
 }
 
 extension Session {
--- a/AirCasting/SDSync/MobileSessions/SDCardMobileSessionsSavingService.swift
+++ b/AirCasting/SDSync/MobileSessions/SDCardMobileSessionsSavingService.swift
@@ -83,7 +83,6 @@ class SDCardMobileSessionsSavingService: SDCardMobileSessionssSaver {
             completion(.failure(SDMobileSavingErrors.noUUID))
             return
         }
-        
         sessionUUID = SessionUUID(stringLiteral: String(sessionUUIDString))
         
         guard let sessionUUID else { return }
@@ -95,11 +94,10 @@ class SDCardMobileSessionsSavingService: SDCardMobileSessionssSaver {
             completion(.success(()))
             return
         }
-        
-        processFile(fileURL: fileURL, session: processedSession, deviceID: deviceID, completion: completion)
+        processFile(sessionUUID: sessionUUID, fileURL: fileURL, session: processedSession, deviceID: deviceID, completion: completion)
     }
     
-    private func processFile(fileURL: URL, session: SDSessionData, deviceID: String, completion: @escaping (Result<Void, Error>) -> Void) {        
+    private func processFile(sessionUUID: SessionUUID, fileURL: URL, session: SDSessionData, deviceID: String, completion: @escaping (Result<Void, Error>) -> Void) {
         var streamsWithMeasurements: [SDStream: [SDSyncMeasurement]] = [:]
         let bufferThreshold = 5000
         var savingFailed = false
@@ -136,7 +134,7 @@ class SDCardMobileSessionsSavingService: SDCardMobileSessionssSaver {
                         return
                     }
                     
-                    self.enqueueForSaving(measurements: measurements, buffer: &streamsWithMeasurements, deviceID: deviceID)
+                    self.enqueueForSaving(sessionUUID: sessionUUID, measurements: measurements, buffer: &streamsWithMeasurements, deviceID: deviceID)
                     
                     savedLines += 1
                     guard savedLines == bufferThreshold else { return }
@@ -192,9 +190,9 @@ class SDCardMobileSessionsSavingService: SDCardMobileSessionssSaver {
         return session
     }
     
-    private func enqueueForSaving(measurements: SDCardMeasurementsRow, buffer streamsWithMeasurements: inout [SDStream: [SDSyncMeasurement]], deviceID: String) {
+    private func enqueueForSaving(sessionUUID: SessionUUID, measurements: SDCardMeasurementsRow, buffer streamsWithMeasurements: inout [SDStream: [SDSyncMeasurement]], deviceID: String) {
        
-        location = location ?? setProperLocation(deviceID: deviceID, measurements: measurements)
+        location = location ?? setProperLocation(sessionUUID: sessionUUID, deviceID: deviceID, measurements: measurements)
         
         if let f = measurements.f {
             streamsWithMeasurements[SDStream(sessionUUID: measurements.sessionUUID, deviceID: deviceID, name: .f, header: .f), default: []].append(SDSyncMeasurement(measuredAt: measurements.date, value: f, location: location))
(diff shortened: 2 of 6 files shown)


AirCasting/CoreData/Session.swift
```diff
@@ -143,6 +143,9 @@ public enum DeviceType: Int, CustomStringConvertible {
         case .AIRBEAM: return "AirBeam"
         }
     }
+    func isAirBeam() -> Bool {
```
mpanuszewska: https://www.youtube.com/watch?v=5mysyCT3pYw&ab_channel=MagnitoMotion

AirCasting/SDSync/MobileSessions/SDCardMobileSessionsSavingService.swift
```diff
@@ -189,9 +193,8 @@ class SDCardMobileSessionsSavingService: SDCardMobileSessionssSaver {
     }
     
     private func enqueueForSaving(measurements: SDCardMeasurementsRow, buffer streamsWithMeasurements: inout [SDStream: [SDSyncMeasurement]], deviceID: String) {
-        var location: CLLocationCoordinate2D?
-        if let lat = measurements.lat, let long = measurements.long {
-            location = CLLocationCoordinate2D(latitude: lat, longitude: long)
+        if location == nil {
```
mpanuszewska: maybe guard let location else { ... } instead if this check?

AirCasting/SDSync/MobileSessions/SDCardMobileSessionsSavingService.swift
```diff
@@ -207,6 +210,47 @@ class SDCardMobileSessionsSavingService: SDCardMobileSessionssSaver {
         }
     }
     
+    private func addMeasurementLocation(deviceID: String, measurements: SDCardMeasurementsRow) -> CLLocationCoordinate2D? {
+
+        let isMini = deviceID.lowercased().contains(AirBeamDeviceType.airBeamMini.rawName)
```
mpanuszewska: this seems like it could be moved to be a device property, so that we can do soemthing like 
deviceID.isMini

AirCasting/SDSync/MobileSessions/SDCardMobileSessionsSavingService.swift
```diff
@@ -207,6 +210,47 @@ class SDCardMobileSessionsSavingService: SDCardMobileSessionssSaver {
...
+        
+        if isMini {
+            guard let sessionUUID else {
```
mpanuszewska: Does this object "SessionSavingService" even make sense without sessionUUID in place? If not, can we easily get rid of this optional?

AirCasting/SDSync/MobileSessions/SDCardMobileSessionsSavingService.swift
```diff
@@ -207,6 +210,47 @@ class SDCardMobileSessionsSavingService: SDCardMobileSessionssSaver {
...
+                Log.error("[SD Sync] Session doesn't have sessionUUID")
+                fatalError()
+            }
+            do {
+                let session = try context.existingSession(uuid: sessionUUID)
+                let filteredStreams = session.allStreams.filter { MeasurementStreamEntity in
+                    !MeasurementStreamEntity.allMeasurements!.isEmpty
```
mpanuszewska: I'm always suspicious with "!" in code. Can we remove it somehow?

AirCasting/SDSync/MobileSessions/SDCardMobileSessionsSavingService.swift
```diff
@@ -207,6 +210,47 @@ class SDCardMobileSessionsSavingService: SDCardMobileSessionssSaver {
...
+                }
+                guard let latestMeasurement = sortedMeasurements.first else {
+                    /* We need to change it to current phone's location */
+                    return session.location!
+                }
+                guard let location = latestMeasurement?.location else {
+                    // ^
+                    return session.location!
+                }
+                return location
```
mpanuszewska: This seems like pipe processing data. It could probably be written functionally, with .map{} etc. But if we don't want to do that, I'd probably at least extract this to a function and name it accordingly. Like... getLastSessionLocation maybe?

mpanuszewska (commented):
Only checked the code, not the logic so just leaving a comment, not an approval. 

But looks good to me.

AirCasting/CoreData/Session.swift
smialko: That's funny bc I searched for this in my code just  before creating a PR because I thought I spotted it. And it wasn't there, and now I know why :extra-grimace: 
Screenshot 2024-04-15 at 13 25 37

AirCasting/SDSync/MobileSessions/SDCardMobileSessionsSavingService.swift
jankrzempek: I think we can make this line a bit easier ex.
` let filteredStreams = session.allStreams.filter { $0.allMeasurements?.isEmpty == false }`

AirCasting/SDSync/MobileSessions/SDCardMobileSessionsSavingService.swift
```diff
@@ -207,6 +210,47 @@ class SDCardMobileSessionsSavingService: SDCardMobileSessionssSaver {
...
+                }
+                let latestMeasurements = filteredStreams.map({ $0.allMeasurements?.last })
+                let sortedMeasurements = latestMeasurements.sorted { first, second in
+                    guard let first, let second else { return false }
+                    return first.time > second.time
+                }
+                guard let latestMeasurement = sortedMeasurements.first else {
+                    /* We need to change it to current phone's location */
+                    return session.location!
+                }
```
jankrzempek: What about changing it a bit, use `compactMap` ex.

```
 let allLastMeasurements = filteredStreams.compactMap { $0.allMeasurements?.last }
                
                if let newestMeasurement = allLastMeasurements.max(by: { $0.time < $1.time }),
                   let location = newestMeasurement.location {
                    return location
                } else {
                .....
```

jankrzempek (approved):
Few suggestions in the comments. Sync works well using AB3. Great work 🥳

AirCasting/SDSync/MobileSessions/SDCardMobileSessionsSavingService.swift
smialko: Hmm, I think no. We do not return anything inside this and guard requires return/throw. 
Or maybe there's something I don't see here?

AirCasting/SDSync/MobileSessions/SDCardMobileSessionsSavingService.swift
smialko: It might not be that easy, we do get the sessionUUID from the file when we process it. 
We could pass it as an argument instead of using the guard statement. Is that what you have in mind?

AirCasting/SDSync/MobileSessions/SDCardMobileSessionsSavingService.swift
smialko: done

smialko: done

AirCasting/SDSync/MobileSessions/SDCardMobileSessionsSavingService.swift
mpanuszewska: True dat, my bad. 

I'm used to syntax like that: 
` location == location ?? setProperLocation(deviceID: deviceID, measurements: measurements)`

but your call, I do't mind either way.

smialko: Changed 🙌🏻

AirCasting/SDSync/MobileSessions/SDCardMobileSessionsSavingService.swift
smialko: For future: we agreed it requires not-so-small refactor so it's added as a separate task in the trello

AirCasting/SDSync/Models/SDStream.swift
```diff
@@ -10,3 +10,8 @@ struct SDStream: Hashable {
     let header: SDCardCSVFileFactory.Header
 }
 
+extension String {
```
mpanuszewska, line 13: hm, I'd probably go for a class SessionUUID that is an extension of string and has this property, since it doesn't make sense for ALL the strings.

mpanuszewska (approved):
One last comment, feel free to merge 👏

AirCasting/SDSync/Models/SDStream.swift
smialko: I'm not sure I understand. 
SessionUUID is a different thing than deviceID. I could change it to extension of `AirBeamDeviceType` but it requires some changes that would require another CR probably. So I'm skipping this for now and it's to be discussed, ok? :)

PR #1002, opened 2024-05-21
fix: Fix for issues with syncing the data from ABMini standalone sessions

mpanuszewska:
This fixes the errors with ABmini standalone session sync. 

There were 3 main issues: 
- asking for wrong characteristics when setting up the sessions 
- not saving any data that didn't come in the first data chunk due to checking for UUID being present in each data chunk (for ABMini it is not) 
- race condition while accessing data from a peripheral that led to chunks being changed during function execution, and therefore to loosing/duplicationg some data. 

Also includes some cosmetic changes.

### DIFF
--- a/AirCasting/ABConnector/SDCardAirBeamServices.swift
+++ b/AirCasting/ABConnector/SDCardAirBeamServices.swift
@@ -112,17 +112,12 @@ class BluetoothSDCardAirBeamServices: SDCardAirBeamServices, BluetoothConnection
             metadataCharacteristicObserver = try bluetoothManager.subscribeToCharacteristic(for: device, characteristic: DOWNLOAD_META_DATA_FROM_SD_CARD_CHARACTERISTIC_UUID) { result in
                 switch result {
                 case .success(let data):
-                    Log.warning("MARTA: Metadata success outside queue")
                     self.queue.async { [weak self] in
-                        Log.warning("MARTA: Metadata success inside queue")
                         currentSessionType = currentSessionType.next
                         self?.handleMetadata(data, device: device, currentSessionType: currentSessionType!, completion: completion)
                     }
                 case .failure(let error):
-                    Log.warning("MARTA: Metadata failure outside queue")
                     self.queue.async { [weak self] in
-                        Log.warning("MARTA: Metadata failure inside queue")
-                        Log.warning("[SD SYNC]  Error while receiving metadata from SD card: \(error.localizedDescription)")
                         self?.finishSync(device: device) { completion(.failure(error)) }
                     }
                 }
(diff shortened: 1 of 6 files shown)


AirCasting/SDSync/Parsing/MiniSDCardMeasurementsParser.swift
```diff
@@ -68,14 +69,13 @@ class MiniSDCardMeasurementsParser: SDMeasurementsParser {
...
     
-    func enumerateSessionLines(lines: [String], action: (String, String) -> Void) {
+    func enumerateSessionLines(lines: [String], action: (String?, String) -> Void) {
         var lastKnownUUID: String?
         lines.forEach { lineString in
             let measurementInfo = lineString.split(separator: ",")
             if measurementInfo.count == firstLineColumns {
                lastKnownUUID = getUUID(lineString: lineString)
             }
-            guard let lastKnownUUID else { return }
```
mpanuszewska, line 78: This had to be removed because ABMini doesn't send UUID with each data pack so the data without it wasn't saved.

AirCasting/ABConnector/SDCardAirBeamServices.swift
```diff
@@ -210,13 +209,17 @@ class BluetoothSDCardAirBeamServices: SDCardAirBeamServices, BluetoothConnection
             self.finishSync(device: device) { completion(.failure(SDCardSyncError.wrongOrderOfReceivedPayload)) }
             return
         }
-        self.receivedMeasurementsCount[sessionType, default: 0] += Constants.SDCardSync.numberOfMeasurementsInDataChunk
+        
+        
+        let numberOfMeasurementsInChunk =  payload.components(separatedBy: "\r\n").filter { !$0.trimmingCharacters(in: ["\n"]).isEmpty }.count
```
mpanuszewska, line 214: Measurements aren't ALWAYS coming in fours, so this is taking into account the fact that there might not be full 4 measurements in the data pack.

AirCasting/Utils/Bluetooth/BluetoothManager.swift
```diff
@@ -376,18 +376,19 @@ final class BluetoothManager: NSObject, BluetoothCommunicator, CBCentralManagerD
...
     func peripheral(_ peripheral: CBPeripheral, didUpdateValueFor characteristic: CBCharacteristic, error: Error?) {
+        let value = characteristic.value
         characteristicsMappingLock.lock()
         defer { characteristicsMappingLock.unlock()}
         guard let containgObserver = charactieristicsMapping.first(where: { CBUUID(string: $0.key.value) == characteristic.uuid }) else { return }
         containgObserver.value.forEach { observer in
             observer.triggerCounter += 1
             guard error == nil else { observer.action(.failure(error!)); return }
-            callbackQueue.async { observer.action(.success(characteristic.value)) }
+            callbackQueue.async { observer.action(.success(value)) }
```
mpanuszewska, line 391: Due to a race condition, sometimes characteristic.value was changing during this function's execution and we were loosing the data.

AirCasting/ABConnector/SDCardAirBeamServices.swift
```diff
@@ -264,7 +269,7 @@ class BluetoothSDCardAirBeamServices: SDCardAirBeamServices, BluetoothConnection
     }
     
     private func allMeasurementsDownloaded() -> Bool {
-        receivedMeasurementsCount.allSatisfy( { $1 >= expectedMeasurementsCount[$0] ?? 0 })
+        return receivedMeasurementsCount.allSatisfy( { $1 >= expectedMeasurementsCount[$0] ?? 0 })
```
smialko, line 272: why adding the return? :)

AirCasting/SDSync/Parsing/MiniSDCardMeasurementsParser.swift
smialko: Yeah, but that was the idea to keep the "lastKnownUUID" for those packs to be able to assign them to a session. What do you do with packs without uuids?

AirCasting/SDSync/SDSyncFileWritingService.swift
```diff
@@ -36,8 +36,15 @@ final class SDSyncFileWritingService: SDSyncFileWriter {
...
         parser.enumerateSessionLines(lines: lines) { uuid, lineString in
-            let url = fileURL(for: sessionType, with: uuid)
-            currentURL = url
+            var url = currentURL
+            
+            // UUID can be nil when data comes from ABMini, in that case just keep using the current URL
+            if let uuid = uuid {
+                url = fileURL(for: sessionType, with: uuid)
+                currentURL = url
+            }
```
smialko, line 45: oh, that's what you're doing 🤔

smialko (commented):
Better late than sorry 🫠 Nothing major in comments, nice fix!

PR #1048, opened 2025-01-21
fix: Remove mechanism for closing sheet - it was causing the update loop

mpanuszewska:
Using exit route on CompleteScreen led to a loop of updateUIView function in SearchAndFollowMap being run again and again on iOS18. 

Refactored to an easier solution of sheet dismissing itself.

### DIFF
--- a/AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift
+++ b/AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift
@@ -50,7 +50,6 @@ class CompleteScreenViewModel: ObservableObject {
     let sessionStartTime: Date
     let sessionEndTime: Date
     let sensorType: String
-    let exitRoute: () -> Void
     
     private var isOwnSession: Bool { userAuthenticationSession.user?.username == session.provider }
     private var session: PartialExternalSession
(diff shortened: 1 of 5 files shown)


AirCasting/SearchAndFollow/SearchAndFollowMap/SearchAndFollowMap.swift
```diff
@@ -47,7 +47,7 @@ struct SearchAndFollowMap: UIViewRepresentable {
     }
     
     func makeUIView(context: Context) -> GMSMapView {
-        GMSServices.provideAPIKey(GOOGLE_MAP_KEY)
```
mpanuszewska, line 50: duplicate, it's instantiated in the app setup.

AirCasting/SearchAndFollow/CompleteScreen/Views/CompleteScreenViewModel.swift
```diff
@@ -253,7 +243,7 @@ class CompleteScreenViewModel: ObservableObject {
     
     private func showAlert() {
         DispatchQueue.main.async {
-            self.alert = InAppAlerts.failedSessionDownloadAlert(dismiss: self.dismissView)
+            self.alert = InAppAlerts.failedSessionDownloadAlert(dismiss: {} )
```
mpanuszewska, line 246: Dismissing the view here isn't really a thing that is userfriendly anyway and it muddled the logic, therefore - it's out.

PR #1056, opened 2025-01-29
Search and Follow session bottomsheet ui loop fix

xameYeoh:
The loop was tied to continuous setting of exploreSessionsButtonTapped property to false by SearchView. The fix breaks the loop by only setting a new value to the property if it's different from an old value

The button in question:


mpanuszewska:
👏 👏 👏

PR #1061, opened 2025-06-10
fix/unnecessary-create-mobile-requests-sd-sync-unfinished-session

xameYeoh:
Contains fixes for:
* [Unnecessary Create session API calls](https://trello.com/c/CYQH40Qo/947-create-session-endpoint-being-called-lots-of-times-unnecessarily)
* [ABMini mobile session not getting finished during SD sync if it's under one minute and ABMini standalone mode has been under one minute too](https://trello.com/c/yhm8IIcv/948-ab-mini-mobile-session-cant-sync-finish-when-quickly-entering-standalone-and-attempting-sd-sync)

### DIFF
--- a/AirCasting.xcodeproj/project.pbxproj
+++ b/AirCasting.xcodeproj/project.pbxproj
@@ -95,6 +95,7 @@
 		339AC2902934DEBE0083D8FF /* SessionNotesStorage.swift in Sources */ = {isa = PBXBuildFile; fileRef = 339AC28F2934DEBE0083D8FF /* SessionNotesStorage.swift */; };
 		339BF3E127D231F000BBCB23 /* StaticSessionHeaderViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 339BF3E027D231F000BBCB23 /* StaticSessionHeaderViewModel.swift */; };
 		339E1A24292F9A7D003910A5 /* UserTriggeredReconnectionController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 339E1A23292F9A7D003910A5 /* UserTriggeredReconnectionController.swift */; };
+		33A16E7B2DF6F76400D9142A /* Concurrency.swift in Sources */ = {isa = PBXBuildFile; fileRef = 33A16E7A2DF6F76400D9142A /* Concurrency.swift */; };
 		33A7FF1A27C5115700A138D6 /* CompleteScreen.swift in Sources */ = {isa = PBXBuildFile; fileRef = 33A7FF1927C5115700A138D6 /* CompleteScreen.swift */; };
 		33A7FF1D27C511A200A138D6 /* CompleteScreenViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 33A7FF1C27C511A200A138D6 /* CompleteScreenViewModel.swift */; };
 		33A7FF2127C5160A00A138D6 /* StaticSessionHeader.swift in Sources */ = {isa = PBXBuildFile; fileRef = 33A7FF2027C5160A00A138D6 /* StaticSessionHeader.swift */; };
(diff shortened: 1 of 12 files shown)


AirCasting.xcodeproj/project.pbxproj
```diff
@@ -10,6 +10,7 @@
 		0A4CB0BCB65AE32375CEBE95 /* Pods_AirCastingTests.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 4AFCF31E099030C84C1A23F9 /* Pods_AirCastingTests.framework */; };
 		33050871276B568D00D18C69 /* ShareSessionViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 33050870276B568D00D18C69 /* ShareSessionViewModel.swift */; };
 		33050876276B7A4400D18C69 /* ShareSessionStreamOptionViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 33050875276B7A4400D18C69 /* ShareSessionStreamOptionViewModel.swift */; };
+		33056A812DF0773F0087FA75 /* SDCardMobileSessionFinisher.swift in Sources */ = {isa = PBXBuildFile; fileRef = 33056A802DF0773F0087FA75 /* SDCardMobileSessionFinisher.swift */; };
```
mpanuszewska, line 13: Gotta love the name 😂

AirCasting/SDSync/SDSyncController.swift
```diff
@@ -184,8 +184,10 @@ class SDSyncController {
     
     private func process(mobileSessionFilesDirectory: URL, deviceID: String, completion: @escaping (Bool) -> Void) {
         Log.info("[SD Sync] Processing mobile file")
-        Task {
-            let location = try await locationTracker.oneTimeLocationUpdate()
+        
+        do {
+            let location = try waitFor(locationTracker.oneTimeLocationUpdate)
+
```
xameYeoh, line 191: I wanted to make the whole process synchronous and wait for it, but since it's using async for getting location, this would mean changing all the methods to async too, which seems like quite a refactor to me. So I reserved to merging the async code with non-async one via a blocking call and waiting for the location here

AirCasting/SDSync/MobileSessions/SDCardMobileSessionFinisher.swift
```diff
@@ -0,0 +1,27 @@
...
+
+protocol SessionFinisher {
+    func callAsFunction(uuid: SessionUUID) throws
+}
+
+class SDCardMobileSessionFinisher : SessionFinisher {
+    @Injected private var persistenceController: PersistenceController
+    private lazy var context: NSManagedObjectContext = persistenceController.editContext
+    
+    func callAsFunction(uuid: SessionUUID) throws {
```
mpanuszewska, line 17: I'm honestly not a fan of this option in language. I know it's popular, so it's not even a change suggestion, just an academic discussion :D 
but having usage of 
`SessionFinisher.finishSession() 
instead of calling a function named (potentialy) differently in each file that it is being used seems more clear and with less potential to break during refactor or sth. 

But I know it's popular now, I just don't see the upside of that.

AirCasting/Utils/Concurrency.swift
```diff
@@ -0,0 +1,30 @@
+// Created by Lunar on 9.06.25.
+//
+
+import Foundation
+
+func waitFor<T>(_ operation: @escaping () async throws -> T) throws -> T {
```
mpanuszewska, line 6: There has to be a library for that... right? RIGHT?! 😂

mpanuszewska (approved):
Looks good to me 🫡

AirCasting/SDSync/MobileSessions/SDCardMobileSessionFinisher.swift
xameYeoh: TLDR: I am a fan of this :D 

In my view this is logical when you are making what in kotlin world is known as a Functional interface (so essentially your class implementing the interface will only have one public method - which will do exactly what the class and interface name suggests - in this case Finish a session). Which promotes the whole idea of One class One purpose. And in these cases, having a client code call like sessionFinisher.finishSession(uuid) starts to seem a bit redundant, when it could be just finishSession(uuid)

AirCasting/Utils/Concurrency.swift
xameYeoh: AFAIK as this is not a preferred practice (but quite used in legacy projects :D), you don't typically have libraries offering these things, but if there is a Swift one that does, I will change it (checked it quickly but didn't find any). In the Android project we had to add a similar thing when we were in the process of refactoring everything to use suspending functions, and later removed this when we were done

PR #1062, opened 2025-06-18
fix/abm-sd-sync-stuck-geolocation

xameYeoh:
Contains fix for [ABM SDSync error when not having geolocation permission](https://trello.com/c/I3yGK9Id/949-ab-mini-mobile-session-sd-sync-stuck-in-finalizing). 

When first asked a geolocation permission while recording a session, user can choose "Allow once". Then, when performing SD Sync, the app didn't ask for the geolocation permission but it needed it in the flow, which produced error and couldn't finish SD syncing the session.

The fix adds "Enable geolocation " screen to the SD sync flow if the app doesn't have it, so it works in the same way as when recording a new session or clearing SD card from the settings.

The changes also contain a minor refactor of TurnOnLocationViewModel and updated README for the distribution procedures

### DIFF
--- a/AirCasting/CreateSessionViews/TurnOnLocation/TurnOnLocationViewModel.swift
+++ b/AirCasting/CreateSessionViews/TurnOnLocation/TurnOnLocationViewModel.swift
@@ -4,62 +4,64 @@
 import Foundation
 import Resolver
 
+enum TurnOnLocationUiProcess {
+    case sdClear
+    case sdSync
+    case createSession(sessionContext: CreateSessionContext)
+}
+
 // [RESOLVER] Init this VM inside View
 class TurnOnLocationViewModel: ObservableObject {
     @Published var isPowerABLinkActive = false
     @Published var isTurnBluetoothOnLinkActive = false
     @Published var isMobileLinkActive = false
     @Published var restartABLink = false
+    @Published var unplugABLink = false
     @Published var alert: AlertInfo?
-    var isSDClearProcess: Bool
-
+    
     @Injected private var locationAuthorization: LocationAuthorization
     @Injected private var bluetoothHandler: BluetoothPermisionsChecker
-    private let sessionContext: CreateSessionContext
+    
+    private let process: TurnOnLocationUiProcess
+    
+    init(process: TurnOnLocationUiProcess) {
+        self.process = process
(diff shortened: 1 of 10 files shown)


AirCasting/CreateSessionViews/TurnOnLocation/TurnOnLocationViewModel.swift
```diff
@@ -4,62 +4,64 @@
...
+                unplugABLink.toggle()
+            case .createSession(let sessionContext):
+                if sessionContext.isMobileSession {
+                    isMobileLinkActive = true
+                } else if bluetoothHandler.isBluetoothDenied() {
+                    isTurnBluetoothOnLinkActive = true
+                } else {
+                    isPowerABLinkActive.toggle()
+                }
+            }
```
xameYeoh, line 60: This looks weird, but seemingly that's how the old logic worked, so I didn't want to change it, fearing something might break

mpanuszewska: Ah, yes, the famous last words.

Readme.md
```diff
@@ -6,88 +6,60 @@
...
-APPSTORE_KEY_ID = XYZ
-APPSTORE_ISSUER_ID = XYZ
-```
-And change the `app` identifier inside Fastfile.
-
-You'll also need an `AuthKey.p8` (an appstoreconnect API key) file placed inside Fastlane directory
+## Distributing builds
+### Certificates and Profiles setup
+Once your Apple ID is added to the project:
+1. Create, download and install (by clicking the downloaded file) into your keychain two types of [certificates](https://developer.apple.com/account/resources/certificates/list):
```
mpanuszewska, line 40: Are you sure this is going to work without the private key?

AirCasting/CreateSessionViews/TurnOnLocation/TurnOnLocationViewModel.swift
mpanuszewska: I retraced this logic just to double check and everything actually looks good, though I have no idea where unplugABLink.toggle() came form :D

mpanuszewska (approved):

PR #1066, opened 2025-07-16
fix/session-name-not-updated-after-change-on-another-device

xameYeoh:
Contains fix for [the ticket](https://trello.com/c/2DyphcCO/950-session-name-doesnt-get-updated)

### DIFF
--- a/AirCasting/CoreData/NSManagedObjectContext+Utils.swift
+++ b/AirCasting/CoreData/NSManagedObjectContext+Utils.swift
@@ -41,13 +41,14 @@ extension NSManagedObjectContext {
         throw MissingSessionEntityError(uuid: uuid)
     }
     
-    func optionalExistingSession(uuid: SessionUUID) throws -> SessionEntity? {
+    func optionalExistingSession(uuid: SessionUUID) -> SessionEntity? {
         do {
             return try existingSession(uuid: uuid)
         } catch is MissingSessionEntityError {
             return nil
         } catch {
-            throw error
+            Log.error("Error when saving changes in \(uuid) session: \(error.localizedDescription)")
+            return nil
         }
     }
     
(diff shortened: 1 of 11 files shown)


AirCasting/CoreData/NSManagedObjectContext+Utils.swift
```diff
@@ -41,13 +41,14 @@ extension NSManagedObjectContext {
...
         do {
             return try existingSession(uuid: uuid)
         } catch is MissingSessionEntityError {
             return nil
         } catch {
-            throw error
+            Log.error("Error when saving changes in \(uuid) session: \(error.localizedDescription)")
+            return nil
         }
     }
```
xameYeoh, line 53: I've checked the call chain of the method and in the end the only error handling was this line of logging, so I moved it here and made the method not throw, to handle the nil cases conveniently from the client code

AirCasting/Utils/Concurrency.swift
```diff
@@ -28,3 +29,18 @@ func waitFor<T>(_ operation: @escaping () async throws -> T) throws -> T {
...
+                do {
+                    let result = try await asyncOperation()
+                    promise(.success(result))
+                } catch {
+                    promise(.failure(error))
+                }
+            }
+        }
+    }
+}
```
xameYeoh, line 46: Using this to bridge Future client code and refactored async code to get rid of the callbacks

AirCasting/Utils/Sequences.swift
```diff
@@ -0,0 +1,21 @@
...
+            if condition(element) {
+                matching.append(element)
+            } else {
+                nonMatching.append(element)
+            }
+        }
+
+        return (matching, nonMatching)
+    }
+}
```
xameYeoh, line 21: Searched for some library that has this specifically but didn't find any

AirCasting/CoreData/NSManagedObjectContext+Utils.swift
mpanuszewska: Behold, useless error throwing. Good call 👍

AirCasting/Utils/BatchError.swift
```diff
@@ -0,0 +1,18 @@
+struct BatchError: Error {
```
mpanuszewska, line 1: Cool, there is 99% chance it will not be used ever again ^^'

mpanuszewska (approved):
I didn't go through the business logic fully, trusting you on that, but from technical side everything looks good. Good to see you getting super comfortable with swift 😏 

Really good job 👏

PR #1069, opened 2025-09-30
fix: Keep location from search from changing

mpanuszewska:
When the user was choosing a location from the list the indicateMapLocationChange was being triggered and was messing with the data, putting the user in the middle of nowhere usually. Now it is triggered only when user scrolls.


mpanuszewska:
merged because we need to release :)

